PR #5, opened 2017-08-18
travis build stages [WIP]

beniwohli:
This splits our matrix into different build stages. First we run linters (only isort so far, but we'll probably add flake8 soonish), then the whole test matrix, then the manylinux1 wheel builds.

Unfortunately, this makes the `travis.yml` much larger due to the fact that we need to spell out the whole matrix manually. According to the travis IRC channel, and independently confirmed by @watson, the implicit build matrix isn't compatible with stages.

### DIFF
--- a/.travis.old.yml
+++ b/.travis.old.yml
@@ -1,0 +1,87 @@
+sudo: false
+language: python
+python:
+- 2.7
+- 3.3
+- 3.4
+- 3.5
+- 3.6
+- nightly
+- pypy
+
+env:
+  matrix:
+    - WEBFRAMEWORK=django-1.8
+    - WEBFRAMEWORK=django-1.9
+    - WEBFRAMEWORK=django-1.10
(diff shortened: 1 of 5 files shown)


ruflin:
@beniwohli Looking at the file I'm thinking we could use a shell / python script to generate the .travis.yml file?

beniwohli:
Haha, we discussed that at the lunch table, how dirty it feels to spell it all out, and that I was sooo close to just write a script. But I think it's ok to have it all defined explicitly, just have to overcome the sense of dirtiness.

watson:
I think it looks good 👍

I think it's overkill to script this 😄 Sure it hurts our soul as engineers to explode the matrix, but when all is said and done it's actually not that hard to maintain. In the future you just copy-paste a section when a new version of Python is released etc. Making a script is not going to pay off IMHO.

PR #6, opened 2017-08-18
Add basic docs structure to agent

ruflin:
This is the basic doc structure needed for the elastic docs build. For convenience I added a script that can be executed with `make docs` to build the docs locally to see if the doc build passes and how the docs will look like.

As soon as we have Jenkins added as CI we should also add a build step for `make docs` to make sure we do not break the doc build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+BUILD_DIR?=build
+
 isort:
 	isort -rc -vb .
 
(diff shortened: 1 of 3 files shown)


beniwohli:
Good stuff. We could also add it to the "Build" stage on travis once #5 is merged

ruflin:
Definitively. I looked at the current travis file to add it and didn't want to have some complex exclude statements so I didn't see it fit. But with the new one that looks promising.

PR #7, opened 2017-08-18
Added flake8 testing, and fixed all current issues reported by flake8

beniwohli:
Note that currently only `elasticapm` is linted, not the `tests` folder. Fixing that one will come later, as it is quite a bit of more work.


elasticapm/__init__.py
```diff
@@ -17,6 +17,6 @@
 except Exception as e:
     VERSION = 'unknown'
 
-from elasticapm.base import *
-from elasticapm.conf import *
-from elasticapm.traces import *
+from elasticapm.base import *  # noqa E403
```
ruflin: ?

ruflin: Probably some flake8 magic?

beniwohli: we import that stuff there to trigger Python to import all the code. Flake8 doesn't like unused imports, so we tell it to please allow it, just this one time

PR #8, opened 2017-08-21
refactor initialization of client using AppConfig

beniwohli:
Since we dropped support for older Django versions, we can make use of
the new `AppConfig.ready` hook to initialize our handlers.

We also add automatic instrumentation of Celery tasks in the `ready`
handler.


elasticapm/contrib/django/handlers.py
```diff
@@ -12,26 +12,60 @@
...
         # Fetch the request from a threadlocal variable, if available
         request = getattr(LogMiddleware.thread, 'request', None)
         request = getattr(record, 'request', request)
 
-        return super(LoggingHandler, self)._emit(record, request=request)
+        return super(LoggingHandler, self)._emit(record, request=request, **kwargs)
+
+
+def exception_handler(request=None, **kwargs):
+    def actually_do_stuff(request=None, **kwargs):
```
jalvz: 😏

jalvz:
there is quite a number of implementations details I can't follow anymore (if I ever could) like the `get_client()` logic, but pretty awesome to get celery support out of the box!

tests look solid too, good stuff 💯

tests/contrib/django/testapp/tasks.py
```diff
@@ -0,0 +1,11 @@
+from ..testapp.celery import app
```
ruflin: These `..` look suspicious.

beniwohli: That's a relative import. But I can change to `tests.contrib.django.testapp.celery` if you prefer

ruflin: All good if it works, it only thought I should probably let this double check.

PR #13, opened 2017-08-25
initial port from Opbeat docs

beniwohli:


docs/getting-started-with-django.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+
+[source,python]
+----
+INSTALLED_APPS = (
+   # ...
+   'elasticapm.contrib.django',
+)
+----
+
+1.  Choose an app name, and set the secret token if needed.
```
ruflin: In the future we should link there to what is allowed for app names and what not, but first we need a doc for it.

docs/getting-started-with-django.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+
+NOTE: For Django 1.8, you can add the same middleware to `MIDDLEWARE_CLASSES`.
+
+Make sure that it is the first middleware in the list.
+
+[[instrumenting-custom-python-code]]
+==== Instrumenting custom Python code
+
+To gain further insights into the performance of your code, please see
+<<./custom-instrumentation.asciidoc#instrumenting-custom-code, instrumenting custom code>>.
```
ruflin: This should work, but normally you should be able to just use the reference. @dedemorton ?

docs/getting-started-with-django.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+    logger.error(
+        'There was some crazy error',
+        exc_info=True,
+        extra={
+            'datetime': str(datetime.now()),
+        }
+    )
+----
+
+the "Extra" tab on detail page for the error group will now have the extra field `datetime`.
```
ruflin: Not sure if I would reference here the UI as the UI could change independent.

docs/getting-started-with-django.asciidoc
```diff
@@ -0,0 +1,299 @@
...
+[source,bash]
+----
+python manage.py elasticapm check
+----
+
+To send a test exception using the current settings, run
+
+[source,bash]
+----
+python manage.py elasticapm test
```
ruflin: Nice.

PR #18, opened 2017-08-31
Add Jenkins and Docker support for testing.

simitt:


Dockerfile
```diff
@@ -0,0 +1,9 @@
+ARG PYTHON_VER
+FROM python:${PYTHON_VER}
+ARG PYTHON_VER
```
ruflin: Is `ARG PYTHON_VER` on purpose twice?

DockerfileDocs
```diff
@@ -0,0 +1,9 @@
+FROM debian:latest
```
ruflin: I suggest to use a specific version here. It's unlikely that debian is going to break the things below in a next major release, but it makes sure it doesn't in case it does.

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+             "PIP_CACHE=.cache/pip"
+             ]) {
+        stage('Checkout'){
+            node('linux'){
+                sh 'echo "do some SCM checkout..."'
+                checkout scm
+            }
+        }
+       
+        stage('Lint'){
```
ruflin: I assume if one stage fails, next stage is not started?

Jenkinsfile
```diff
@@ -0,0 +1,69 @@
...
+                return
+            }
+            python_versions.each{ py_ver -> 
+                frameworks.each{ framework -> 
+                    def job = "${py_ver}_${framework}".toString()
+                    if(exclude_jobs.contains(job)){
+                        return
+                    }
+                    test_jobs[job] = {
+                        node('linux'){
```
ruflin: I assume at a later stage here also `node('windows')` or similar could be added?

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
+version: '2.1'
+
+services:
+  postgres:
+    user: postgres
+    image: postgres
```
ruflin: I suggest to specify a version here.

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+    ports: 
+      - "5432:5432"
+    environment:
+      POSTGRES_USER: 'postgres'
+      POSTGRES_DB: 'elasticapm_test'
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready"]
+      interval: 30s
```
ruflin: Should we make this much more frequent like `1s`? I'm assuming this would make the tests starting quicker as the container is recognised as healthy earlier.

docker-compose.yml
```diff
@@ -0,0 +1,53 @@
...
+      timeout: 30s
+      retries: 3
+
+  mongodb:
+    image: mongo
```
ruflin: add a specific version? Same applies for all other images.

History is that in metricbeat we didn't have this initially but over time new and old builds started to fail because there was a new `mongodb` version.

setup.cfg
```diff
@@ -15,17 +15,17 @@ isort_ignore=
...
 known_first_party=elasticapm,tests
-known_third_party=pytest,flask,aiohttp,urllib3_mock,webob,memcache,pymongo,boto3,logbook,twisted,celery,zope,urllib3,redis,jinja2,requests,certifi
+known_third_party=pytest,flask,aiohttp,urllib3_mock,webob,memcache,pymongo,boto3,logbook,twisted,celery,zope,urllib3,redis,jinja2,requests,certifi,mock
 default_section=FIRSTPARTY
 sections=FUTURE,STDLIB,DJANGO,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
 
 
 [flake8]
-exclude=elasticapm/utils/wrapt
+exclude=elasticapm/utils/wrapt,elasticapm/utils/six.py,build
```
ruflin: @beniwohli Is `six` still needed here?

tests/instrumentation/psycopg2_tests.py
```diff
@@ -14,7 +14,7 @@
 except ImportError:
     has_psycopg2 = False
 
-travis_and_psycopg2 = 'TRAVIS' not in os.environ or not has_psycopg2
+travis_and_psycopg2 = False # 'TRAVIS' not in os.environ or not has_psycopg2
```
ruflin: Should the comment be changed?

ruflin:
@simitt One question which does not have to be covered in this PR but I'm curious: What is the command to run a specific tests suite locally on my machine?

Dockerfile
simitt: yes, because of [how ARG and FROM interact](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact).

> An ARG declared before a FROM is outside of a build stage, so it can’t be used in any instruction after a FROM. To use the default value of an ARG declared before the first FROM use an ARG instruction without a value inside of a build stage:

Jenkinsfile
simitt: yes afaik, I'll remove the `currentBuild` handling.

Jenkinsfile
simitt: That is the plan yes, but as you said on a later stage, since we also have to figure out how to handle it with dependencies/docker. 
Right now the goal is to have the same testsuite as we have on travis, and also enable usage of docker for local testing.

docker-compose.yml
simitt: not sure about that, this way we use the latest one, @beniwohli would you like to have a fixed version for the services?

tests/instrumentation/psycopg2_tests.py
simitt: I leave this up to @beniwohli if he generally wants to adapt this test handling.

simitt:
@ruflin If you have a look at the scripts in [`tests/scripts/docker/`](https://github.com/simitt/apm-agent-python/tree/ed7e0f06601ff5e31cb45cd66c84dc3a2497b63b/tests/scripts/docker) you find the shell scripts triggered by jenkins, you can just trigger those with the proper args.

Dockerfile
ruflin: Nice. Something learned today :-)

Jenkinsfile
```diff
@@ -0,0 +1,63 @@
+#!groovy
```
ruflin: @andrewkroh Have a look at this file. I could see this also becoming useful for us for the `beats` build matrix and stages. Like this we would have more directly the repo itself.

Dockerfile
```diff
@@ -0,0 +1,9 @@
...
+ENV PYTHON_VER ${PYTHON_VER}
```
beniwohli: the Python docker files set an environment variable `PYTHON_VERSION`. Could we use that instead?

DockerfileDocs
```diff
@@ -0,0 +1,9 @@
+FROM debian:8
+
+RUN apt-get update && \
+    apt-get install -y xsltproc libxml2-utils build-essential git curl python2.7
```
beniwohli: to make the image a little bit slimmer, we could clean up after ourselves here, by adding `&& rm -rf /var/lib/apt/lists/*`

DockerfileLint
```diff
@@ -0,0 +1,11 @@
+FROM python:3.6
+
+WORKDIR /app
+
+ENV PIP_CACHE ""
+ENV SCRIPT ""
+COPY ./Makefile /app
```
beniwohli: aren't all these files already available through the run-time mount?

docker-compose.yml
beniwohli: I think it's fine to have the latest version. Generally speaking, our interaction with Postgres is so limited, that I don't think any version change would impact our tests

setup.cfg
beniwohli: nope, can be removed (but this branch probably needs a rebase to get the six-removal)

tests/instrumentation/psycopg2_tests.py
beniwohli: we could change it to

    has_postgres_configured = 'POSTGRES_DB' in os.environ

and then change the skips below to

    @pytest.mark.skipif(not has_postgres_configured, "PostgresSQL not configured")

beniwohli:
Great stuff. One thing that I'm wondering is if we could get rid of the custom Dockerfiles. All they do is set a couple environment variables and customize what's getting run. I think we could just use the appropriate `python:X.Y.Z` image, and run it with the appropriate `-e` flags.

DockerfileLint
simitt: for linting I only mount the `.cache\pip`.

beniwohli: Any specific reason for that? We still need access to the source code for linting, and mounting the folder seems easier than building a new image every time

tests/instrumentation/psycopg2_tests.py
simitt: will change

simitt:
@watson fyi

PR #19, opened 2017-08-31
Improve Docker and Jenkins handling for tests.

simitt:
* Mount volumes for all docker jobs
* Remove unnecessary DockerImage.


docker-compose.yml
```diff
@@ -31,9 +31,9 @@ services:
       - "6397:6397"
 
   run_tests:
-    build: 
-      context: .
-      dockerfile: Dockerfile
+    image: $PYTHON_IMAGE
```
ruflin: Nice, that will speed up the builds I assume.

tests/scripts/docker/remove.sh
```diff
@@ -1,12 +0,0 @@
-#!/usr/bin/env bash
```
ruflin: How is the cleanup happening now?

Jenkinsfile
```diff
@@ -1,8 +1,8 @@
 #!groovy
 
-def python_versions = ['2.7','3.3','3.4','3.5','3.6','pypy','latest']
+def python_versions = ['python:2.7','python:3.3','python:3.4','python:3.5','python:3.6','pypy:2', 'pypy:3','python:latest']
```
beniwohli: I think we can omit `python:latest`, it's essentially the same as `python:3.6`. Unfortunately, there isn't a docker image for nightly Python master builds, but it's not suuper important to have those.

tests/scripts/docker/remove.sh
simitt: Since we do not build new images anymore but just use plain `python` images I don't think it is necessary to remove them again afterwards.

Jenkinsfile
simitt: will do

tests/scripts/docker/remove.sh
ruflin: Got it, makes sense. 

For the cleanup I was more thinking of volumes and network interfaces, to make sure they are removed every time at the end of a build, also if the build crashes.

On the beats side we had several issues in the past where tests were unpredictable because volumes were reused or unused volumes filled up the disk over time. Same for network interfaces, as there is a limit of something like 25 network interfaces max.

I think something like `docker prune ...` or `docker-compose down ...` should do the trick.

PR #22, opened 2017-09-01
Add docs for running tests locally with docker.

simitt:
This should be a start for testsuite documentation.


docs/run-tests-locally.asciidoc
```diff
@@ -0,0 +1,40 @@
+[[run-tests-locally]]
+== Run Tests Locally
+
+To run tests locally you can make use of the docker images also used when running the whole testsuite with Jenkins. 
+Running the full testsuite first does some linting and then runs the actual tests with different versions of python and different webframeworks. For a full overview of the test matrix and supported versions have a look at the [Jenkins configuration](Jenkinsfile). 
+
+[[linter]]
+=== Linter
+We run three kinds of linting `isort`, `flake8`, and `docs`. You can trigger each single linter locally by running:
```
beniwohli: `docs` isn't really a linter IMO. It just tests if the docs can be generated without errors. Not sure how to write it differently, though

docs/run-tests-locally.asciidoc
```diff
@@ -0,0 +1,40 @@
+[[run-tests-locally]]
+== Run Tests Locally
+
+To run tests locally you can make use of the docker images also used when running the whole testsuite with Jenkins.
```
beniwohli: I'd write it as "test suite"

docs/run-tests-locally.asciidoc
```diff
@@ -0,0 +1,40 @@
+[[run-tests-locally]]
+== Run Tests Locally
+
+To run tests locally you can make use of the docker images also used when running the whole testsuite with Jenkins. 
+Running the full testsuite first does some linting and then runs the actual tests with different versions of python and different webframeworks. For a full overview of the test matrix and supported versions have a look at the [Jenkins configuration](Jenkinsfile).
```
beniwohli: "web frameworks"

Also, does that link work?

docs/run-tests-locally.asciidoc
```diff
@@ -0,0 +1,40 @@
...
+
+* docs 
+[source,bash]
+----
+$ ./tests/scripts/docker/docs.sh 
+----
+
+[[running-tests]]
+=== Running Tests
+We run the testsuite on different combinations of python versions and webframeworks. For triggering the testsuite for a specific combination locally you can run:
```
beniwohli: capitalize Python :)

docs/run-tests-locally.asciidoc
```diff
@@ -0,0 +1,43 @@
+[[run-tests-locally]]
+== Run Tests Locally
+
+To run tests locally you can make use of the docker images also used when running the whole test suite with Jenkins. 
+Running the full test suite first does some linting and then runs the actual tests with different versions of Python and different web frameworks. For a full overview of the test matrix and supported versions have a look at 
+https://github.com/elastic/apm-agent-python/blob/master/Jenkinsfile[Jenkins Configuration].
```
simitt: This is linking to the github repo, not sure if that is what we want, but since the `Jenkinsfile` is not included in our docs I did not see any other option here, maybe @ruflin knows more?

ruflin: We could include it but we would have to add it as `sources` in the `conf.yaml` file: https://github.com/elastic/docs/blob/master/conf.yaml#L541 TBH I think for now it is completely fine to link to this file in the Github repo.

PR #26, opened 2017-09-04
series of Django 2.0 fixes

beniwohli:
Main change is a different approach of `MIDDLEWARE_CLASSES` vs. `MIDDLEWARE`. Instead of duplicating the tests, we decide which configuration option to use at runtime.

### DIFF
--- a/tests/contrib/django/django_tests.py
+++ b/tests/contrib/django/django_tests.py
@@ -12,7 +12,6 @@ from django.contrib.sites.models import Site
 from django.core.handlers.wsgi import WSGIRequest
 from django.core.management import call_command
 from django.core.signals import got_request_exception
-from django.core.urlresolvers import reverse
 from django.db import DatabaseError
 from django.http import QueryDict
 from django.http.cookie import SimpleCookie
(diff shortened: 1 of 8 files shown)


tests/contrib/django/django_tests.py
```diff
@@ -242,6 +247,25 @@ def test_user_info_with_non_django_auth(self):
         event = self.elasticapm_client.events.pop(0)['errors'][0]
         assert event['context']['user'] == {}
 
+    @pytest.mark.skipif(django.VERSION < (1, 10),
+                        reason='MIDDLEWARE new in Django 1.10')
+    def test_user_info_with_non_django_auth_django_2(self):
+        with self.settings(INSTALLED_APPS=[
+            app for app in settings.INSTALLED_APPS
+            if app != 'django.contrib.auth'
+        ]) and self.settings(MIDDLEWARE_CLASSES=None, MIDDLEWARE=[
```
simitt: except of the line `self.settings(MIDDLEWARE_CLASSES=[..]` this seems to be the exact same test for `django1` and `django2`. Since it seems that both of the tests will be in the code base for quite some time, what do you think about creating a common method for the test logic and just having the specific parts explicitly defined in the specific methods?

tests/contrib/django/django_tests.py
```diff
@@ -404,28 +446,8 @@ def test_record_none_exc_info(self):
...
-            self.assertEquals(request['url']['search'], '')
-            self.assertEquals(request['body'], None)
-
-    @pytest.mark.skipif(django.VERSION < (1, 10),
-                        reason='new-style middlewares')
-    def test_404_new_style_middleware(self):
-        with self.settings(MIDDLEWARE_CLASSES=None, MIDDLEWARE=[
-                'elasticapm.contrib.django.middleware.Catch404Middleware']):
+        with self.settings(**middleware_setting(django.VERSION,
+                                                ['elasticapm.contrib.django.middleware.Catch404Middleware'])):
```
simitt: Could you briefly motivate why you removed the specific checks?

tests/contrib/django/django_tests.py
beniwohli: That's what I did for most tests with the `middleware_setting` helper function -- it generates the correct settings dict on the fly for the current Django version. I couldn't do that here, because we _remove_ a specific middleware, and leave the rest as is. I could have written another helper function, but because we only use this pattern once or twice, it didn't feel like the right thing to do.

tests/contrib/django/django_tests.py
beniwohli: they are all still there. I merged two duplicate test cases (one for `MIDDLEWARE_CLASSES`, one for `MIDDLEWARE`) into one, using the `middleware_setting` helper function. If you expand the code block right under this diff, you'll see the exact same lines that have been removed again.

PR #27, opened 2017-09-05
added tagging functionality

beniwohli:
To tag a transaction, the user only has to import `elasticapm`, then call
`elasticapm.add_tag()`:

    import elasticapm
    elasticapm.add_tag('foo', 'bar')


beniwohli:
I think @roncohen already opined that #28 is the way to go. I'll close this one.

PR #28, opened 2017-09-05
added tagging functionality (alternative API)

beniwohli:
To tag a transaction, the user only has to import `elasticapm`, then call
`elasticapm.tag(tag1=value1, tag2=value2)`:

    import elasticapm
    elasticapm.tag(foo='bar')

This is an alternative API to PR #27. It's IMO a nicer API, but due to the key being a keyword argument, it has stricter rules than what we enforce in the APM server (e.g. it can't have any whitespace).

### DIFF
--- a/tests/instrumentation/transactions_store_tests.py
+++ b/tests/instrumentation/transactions_store_tests.py
@@ -1,7 +1,9 @@
+import logging
 import time
 
 from mock import Mock
 
+import elasticapm
 from elasticapm.traces import TransactionsStore, get_transaction, trace
 from tests.utils.compat import TestCase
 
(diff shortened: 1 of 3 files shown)


simitt:
nice PR!

beniwohli:
that would basically be a mix of the two, right? I'm not sure if the benefit is worth the added confusion of having two ways to set tags (`tag(key='val')` and `add_tag(key='key', val='val')`)

simitt:
Sorry for being unclear, I did not mean to provide two ways, but instaed of `elasticapm.tag(foo='bar')` providing `elasticapm.add_tag(key='foo', val='bar')`.
But as I said, I think having a bit stricter rules here should be fine, just in case you want to completely align this might be a possibility.

roncohen:
I like this one compared to https://github.com/elastic/apm-agent-python/pull/27.

It's not hard to imagine a scenario where you want to set a bag of tags all at once, and you can with this API:

```
import elasticapm
css_props = {...}
elasticapm.tag(**{prop: css_props[prop] for prop in important_css_props})
```

tests/instrumentation/transactions_store_tests.py
```diff
@@ -128,3 +130,29 @@ def test_get_transaction_clear():
...
+    assert record.levelno == logging.WARNING
+    assert 'foo' in record.args
+
+
+def test_tag_with_non_string_value():
+    requests_store = TransactionsStore(lambda: [], 99999)
+    t = requests_store.begin_transaction("test")
+    elasticapm.tag(foo=1)
+    requests_store.end_transaction(200, 'test')
+    assert t._tags == {'foo': '1'}
```
simitt: you could add a test that checks the string pattern.

beniwohli: I had such a test in #27. But because this API enforces stricter limits to the tag name than our regex (by virtue of being set as a function keyword), there is no way to use a tag name that would fail our regex.

simitt: I remember, was just wondering, since you still validate against the regex.

beniwohli: Yup, I thought about removing the regex check, but decided to leave it there a) for documentation purposes, and b) if we ever change the API

PR #29, opened 2017-09-06
removed Query event and camelCased client API methods

beniwohli:
this removes the long-deprecated camelCased API methods, as well as the `Query` event type, which AFAIK we never really supported.


ruflin:
Our linter does not seem to be happy :-(

beniwohli:
yeah, I'm currently in a refactoring hell-hole, I'll fix this when I dug my way out of it

ruflin:
Do you mean you fix it as part of this PR or want to have it in with linter going happy?

beniwohli:
We can just leave it open until I get around to fix the linting issue

PR #30, opened 2017-09-06
Stop docker containers and fix redis port.

simitt:
* Stop docker containers after running the tests.
* Fix redis port to use default 6379.

Since we will also have other agent tests run on the same Jenkins slaves it is necessary to also stop and clean the database volumes. 

Additionally I changed the `pip cache` handling a bit, since I don't think we get additional value from passing in the pip cache directory as parameters.


Jenkinsfile
```diff
@@ -17,33 +17,44 @@ node{
...
+                        try{ sh 'docker stop $(docker ps -q -a)' }catch(e){}
+                        try{ sh 'docker rm -v $(docker ps -a -q)' }catch(e){}
+                        sh("./tests/scripts/docker/${it}.sh")
                     }
                 }
             }
             parallel linter_jobs
         }
+
+        stage('Docs'){
```
ruflin: I like that the doc build is not a separate stage.

simitt: is the `not` a typo and you meant `now`?

ruflin: yep, typo. sorry.

ruflin:
Yay, Jenkins build was run on the build 🎉 

It seems like some python tests are failing (not sure if that is expected) and then something goes wrong during the cleanup, I assume when it tries to remove the worksspace. Let me know if I can help.

simitt:
Yes the genreal trigger is set up (thx to the infra team), but there is an issue with wiping out the workspace, that was not appearent before. The problem is that tests are run with user `root` instead of `jenkins` in docker, since there is no user `jenkins`, but the current dir is mounted. I am on it and will let you know once it is fixed.

simitt:
@beniwohli , @ruflin Jenkins is green, so if you're fine with the PR we could merge it. I rebase and push again.

ruflin:
LGTM. @beniwohli can you please merge if ok?

beniwohli:
I'm not a huge fan of having a basically empty `Dockerfile` in the root of the repository that has no value outside of testing. 

I understand the reason we did this is because individual test runs created files in the mounted directory, which then the cleanup script then is unable to delete? There should be other ways to fix this:

 * avoid the test suite creating any files. Not sure if that's doable, and it would also be brittle.
 * make sure that the `jenkins` user has write access to files created by the test suite. Maybe running the container with the same user could help, e.g. 
     
       docker-compose run ... --user `id -u`

simitt:
I tried a lot changing file permissions, owners, creating a user inside docker,.. but nothing worked properly, on jenkins+locally. Just passing `--user id -u` does not work properly either.

If you or @ruflin can come up with a working solution for changing permissions instead of only copying the working directory, I am more than happy to apply it, but I put already way too much time into trying to changing permissions.

ruflin:
We have a similar issue an beats and all our approaches so far failed. One for example is https://github.com/elastic/beats/pull/4860

As for now I think it's much more important to have CI running then have it perfect, lets go forward with this one and merge it in.

PR #33, opened 2017-09-11
refactored client configuration to be uniform across all supported frameworks

beniwohli:
This uses a Config class to store client configuration, and to configure
the client in a uniform way.

### DIFF
--- a/elasticapm/conf/__init__.py
+++ b/elasticapm/conf/__init__.py
@@ -44,14 +44,14 @@ class _ConfigValue(object):
 
 class _ListConfigValue(_ConfigValue):
     def __init__(self, dict_key, list_separator=',', **kwargs):
-        self.list_serparator = list_separator
+        self.list_separator = list_separator
         super(_ListConfigValue, self).__init__(dict_key, **kwargs)
 
     def __set__(self, instance, value):
         if isinstance(value, compat.string_types):
-            value = value.split(self.list_serparator)
-        else:
-            value = list(value) if value is not None else value
+            value = value.split(self.list_separator)
+        elif value is not None:
+            value = list(value)
         if value:
             value = [self.type(item) for item in value]
         instance._values[self] = value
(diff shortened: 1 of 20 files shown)


elasticapm/conf/__init__.py
```diff
@@ -10,8 +10,121 @@
...
+
+    def __set__(self, instance, value):
+        if value is not None:
+            value = self.type(value)
+        instance._values[self] = value
+
+
+class _ListConfigValue(_ConfigValue):
+    def __init__(self, dict_key, list_separator=',', **kwargs):
+        self.list_serparator = list_separator
```
simitt: typo `serparator`

elasticapm/conf/__init__.py
```diff
@@ -10,8 +10,121 @@
...
+        super(_ListConfigValue, self).__init__(dict_key, **kwargs)
+
+    def __set__(self, instance, value):
+        if isinstance(value, compat.string_types):
+            value = value.split(self.list_serparator)
+        else:
+            value = list(value) if value is not None else value
```
simitt: why do you need to set `value` to `value` if it is `None`? How about: 
```
elif value not None: 
  value = list(value)
```

tests/contrib/django/django_tests.py
```diff
@@ -706,7 +709,10 @@ def test_request_metrics_301_prepend_www(self):
 
         # enable middleware wrapping
         client = get_client()
-        client.instrument_django_middleware = True
+        client.config.instrument_django_middleware = True
+
+        from elasticapm.contrib.django.middleware import TracingMiddleware
+        TracingMiddleware._elasticapm_instrumented = False
```
simitt: why is this set to `false`?

elasticapm/conf/__init__.py
```diff
@@ -10,8 +10,121 @@
...
+        'elasticapm.processors.sanitize_http_request_body',
+    ])
+    traces_send_frequency = _ConfigValue('TRACES_SEND_FREQ', type=int, default=60)
+    async_mode = _BoolConfigValue('ASYNC_MODE', default=True)
+    instrument_django_middleware = _BoolConfigValue('INSTRUMENT_DJANGO_MIDDLEWARE', default=True)
+    transactions_ignore_patterns = _ListConfigValue('TRANSACTIONS_IGNORE_PATTERNS', default=[])
+    app_version = _ConfigValue('APP_VERSION')
+    disable_send = _BoolConfigValue('DISABLE_SEND', default=False)
+
+    def __init__(self, config_dict=None, env_dict=None, default_dict=None):
```
simitt: Why don't you set `{}` and `os.environ` as default values instead of setting to `None` and then checking for `None`?

beniwohli: That's Python Gotcha Number One ;p 

http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

elasticapm/conf/__init__.py
beniwohli: 🤦‍♂️ thanks!

elasticapm/conf/__init__.py
beniwohli: that is indeed nicer, thanks!

tests/contrib/django/django_tests.py
beniwohli: we're resetting it in case another test already instrumented the middleware. Python modules and classes are basically global state, so we need to ensure that the global state is correct before running the test

elasticapm/conf/__init__.py
simitt: that's interesting, thank you for pointing it out (that's also why I mentioned that I reviewed as good as my python skills allowed ..)

beniwohli: No worries :) It's not the number 1 gotcha because it would be stupid to make the mistake, it's number 1 gotcha because everybody falls into the trap sooner or later.

simitt:
LGTM

PR #36, opened 2017-09-12
added in-app frame detection as a processor

beniwohli:
In Opbeat, this detection was done server-side. The APM Server doesn't have this functionality, so we implement it agent-side.


elasticapm/contrib/django/client.py
```diff
@@ -66,9 +66,9 @@ def get_client(client=None):
 
 
 def get_client_default_config():
-    config = getattr(django_settings, 'ELASTIC_APM', {})
     return dict(
-        include_paths=set(config.get('INCLUDE_PATHS', [])) | _get_installed_apps_paths(),
+        include_paths=_get_installed_apps_paths(),
+        exclude_paths={'django'}
```
jalvz: L71 is not documented, right? default `exclude_paths` shouldn't be "depending on framework"?

beniwohli: Good catch! I mention in the `include_paths` section that for Django, `django.contrib` apps are excluded (which is caused by this exclude), but I should also mention it in the `exclude_paths` setting

docs/configuration.asciidoc
```diff
@@ -242,6 +242,36 @@ WARNING: We recommend to always include the default set of validators if you cus
...
+
+|============
+| Environment                 | Django/Flask    | Default
+| `ELASTIC_APM_EXCLUDE_PATHS` | `EXCLUDE_PATHS` | `set(['elasticapm'])`
+| multiple values separated by commas, without spaces |||
+|============
+
+A set of module paths that should be considered when excluding a frame from being detected as an in-app frame.
+
+NOTE: for a given module path, all sub-modules will also match. E.g. `foo.bar` also matches for `foo.bar.baz`.
```
jalvz: You might want to note that `EXCLUDE_PATHS` has preference over `INCLUDE_PATHS` (ie: if a path is in both, then it is excluded)

jalvz: You might also want to note that both exclude/include override defaults and not append to them. With lists there is always room for confusion i guess

tests/processors/tests.py
```diff
@@ -201,3 +201,27 @@ def test_remove_stacktrace_locals():
...
+                {'module': 'foo'},
+                {'module': 'foo.bar'},
+                {'module': 'foo.bar.baz'},
+                {'module': 'foobar'},
+                {'module': 'foo.bar.bazzinga'},
+            ]
+        }
+    }
+
+    client = Client(include_paths=['foo'], exclude_paths=['foo.bar.baz'])
```
jalvz: re. documenting `exclude_paths` preference:
i mean that if you do the other way around `Client(include_paths=['foo.bar.baz'], exclude_paths=['foo'])),
then `foo.bar.baz` will be excluded (i think?), which might be not so intuitive.

PR #38, opened 2017-09-12
added validation for app_name and transport_class

beniwohli:
Also moved client tests to use pytest tests/fixtures

### DIFF
--- a/elasticapm/conf/__init__.py
+++ b/elasticapm/conf/__init__.py
@@ -19,27 +19,49 @@ from elasticapm.utils import compat
 __all__ = ('setup_logging', 'Config')
 
 
+class ConfigurationError(ValueError):
+    def __init__(self, msg, field_name):
+        self.field_name = field_name
+        super(ValueError, self).__init__(msg)
+
+
 class _ConfigValue(object):
-    def __init__(self, dict_key, env_key=None, type=compat.text_type, validators=None, default=None, doc=None):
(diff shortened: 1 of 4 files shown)


elasticapm/conf/__init__.py
```diff
@@ -19,27 +19,49 @@
...
+            instance._values[self.dict_key] = value
+
+    def _validate(self, instance, value):
+        if value is None and self.required:
+            raise ConfigurationError(
+                'Configuration error: value for {} is required.'.format(self.dict_key),
+                self.dict_key,
+            )
+        if self.validators and value is not None:
+            if not all(validator(value) for validator in self.validators):
```
simitt: hard to read, but elegant solution :)

tests/client/client_tests.py
```diff
@@ -509,3 +482,21 @@ def test_disable_send(mock_send_remote):
...
+    assert client.config.disable_send
+
+
+def test_empty_transport_disables_send():
+    client = Client({'APP_NAME': 'foo', 'TRANSPORT_CLASS': None})
+
+    assert len(client.config.errors) == 1
+    assert 'TRANSPORT_CLASS' in client.config.errors
+
+    assert client.config.disable_send
```
simitt: you could also add a test checking that `servers`is set.

beniwohli:
@simitt totally agree about splitting up the PR. I got carried away refactoring the tests. I'll try to avoid that going forward.

PR #39, opened 2017-09-12
delete old urllib2 based transport class

beniwohli:
This removes some dead, unused code. The new, urllib3 based transport has been the default on Opbeat for more than a year and we never experienced urllib3 specific issues


beniwohli:
... turns out we use `HTTPTransport` as base class for both the urllib3 transport and the asyncio transport. I'll leave it in place for now.

PR #41, opened 2017-09-12
added flask docs, and refactored initialization of the flask agent a bit

beniwohli:
also added `[float]` all over the place and restructured docs a bit

### DIFF
--- a/docs/flask.asciidoc
+++ b/docs/flask.asciidoc
@@ -1,0 +1,232 @@
+[[flask-support]]
+== Flask support
+
+Getting Elastic APM set up for your Flask project is easy,
+and there are various ways you can tweak it to fit to your needs.
+
+[float]
+[[flask-installation]]
+=== Installation
+
+Install the Elastic APM agent using pip:
+
+[source,bash]
+----
+$ pip install elastic-apm[flask]
+----
+
+or add `elastic-apm[flask]` to your project's `requirements.txt` file.
+
+NOTE: If you use Flask with uwsgi, make sure to
+http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads[enable
+threads].
+
+[float]
+[[flask-setup]]
+=== Setup
+
+To set up the agent, you need to initialize it with appropriate settings.
+
+The settings are configured either via environment variables,
+the application's settings, or as initialization arguments.
+
+You can find a list of all available settings in the <<configuration, Configuration>> page.
+
+To initialize the agent for your application using environment variables:
+
+[source,python]
+----
+from elasticapm.contrib.flask import ElasticAPM
+
+app = Flask(__name__)
+
+apm = ElasticAPM(app)
+----
+
+To configure the agent using `ELASTIC_APM` in your application's settings:
+
+[source,python]
+----
+from elasticapm.contrib.flask import ElasticAPM
+
+app.config['ELASTIC_APM'] ={
+    'APP_NAME': '<APP-NAME>',
+    'SECRET_TOKEN': '<SECRET-TOKEN>',
+}
+apm = ElasticAPM(app)
+----
+
+The final option is to initialize the agent with the settings as arguments:
+
+[source,python]
+----
+from elasticapm.contrib.flask import ElasticAPM
+
+apm = ElasticAPM(app, app_name='<APP-ID>', secret_token='<SECRET-TOKEN>')
+----
+
+[float]
+[[flask-debug-mode]]
+==== Debug Mode
+
+Please note that errors and transactions will only be sent to the apm server if your app is *not* in
+http://flask.pocoo.org/docs/0.12/quickstart/#debug-mode[debug mode].
+
+To force the agent to send data while the app is in debug mode,
+set the value of `DEBUG` in the `ELASTIC_APM` dictionary to `True`:
+
+[source,python]
+----
+app.config['ELASTIC_APM'] = {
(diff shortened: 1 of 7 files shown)


docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+----
+apm.capture_message('hello, world!')
+----
+
+[float]
+[[flask-logging]]
+=== Logging
+
+Passing `logging=True` to the `ElasticAPM` constructor will make the agent
+automatically log errors that are logging using Python's built-in `logging` module.
```
simitt: I would delete the `that are logging` part to have 

"Passing `logging=True` to the `ElasticAPM` constructor will make the agent automatically log errors, using Python's built-in `logging` module."

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+[source,python]
+----
+from elasticapm.contrib.flask import ElasticAPM
+
+apm = ElasticAPM(app, logging=True)
+----
+
+For fine-grained control, you can initialize a logging handler and add it,
+just as you would any other handler.
```
simitt: "and add it, just as you would _with_ any other handler."

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+NOTE: `exc_info=True` adds the exception info to the data that gets sent to the APM server.
+Without it, only the message is sent.
+
+[float]
+[[flask-extra-data]]
+==== Extra data
+
+If you want to send more data to ElasticAPM than what you get with the agent by default,
+logging can be done like so:
```
simitt: "In addition to what the agents log by default, you can send _extra_ information":

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+from elasticapm.contrib.flask import ElasticAPM
+apm = ElasticAPM(app)
+----
+
+To configure the agent using `ELASTIC_APM` in your application's settings:
+
+[source,python]
+----
+from elasticapm.contrib.flask import ElasticAPM
+
```
jalvz: missing `app = Flask(__name__)` maybe?

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+apm = ElasticAPM(app, app_name='<APP-ID>', secret_token='<SECRET-TOKEN>')
+----
+
+[float]
+[[flask-debug-mode]]
+==== Debug Mode
+
+Please note that errors and transactions will only be sent to the apm server
+if your app is *not* in
+http://flask.pocoo.org/docs/0.12/quickstart/#debug-mode[debug mode].
```
jalvz: 💯 are you sure transactions _and_ errors? i recall transactions is fine

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+
+To force the agent to send data during development,
```
jalvz: what does this mean?

docs/flask.asciidoc
jalvz: you mean "that are logged* using"...?

docs/flask.asciidoc
```diff
@@ -0,0 +1,232 @@
...
+----
+app.config['ELASTIC_APM'] = {
+    ...
+    'TRANSACTIONS_IGNORE_PATTERNS': ['^OPTIONS ', '/api/']
+    ...
+}
+----
+
+This would ignore any requests using the `OPTIONS` method
+and any `GET` requests containing `GET /api/`.
```
jalvz: why `GET`? 
`POST / api/` wouldn't be ignored?

docs/sanitizing-data.asciidoc
```diff
@@ -1,12 +1,12 @@
...
-Sometimes it is necessary to sanitize the data sent to Elastic APM, e.g. remove
-sensitive data.
+Sometimes it is necessary to sanitize the data sent to Elastic APM,
+e.g. remove sensitive data.
 
-To do this with the Elastic APM module, you create a processor. A processor is a function
-that takes a `client` instance as well as an event (an error or a transaction), and returns
-the modified event.
+To do this with the Elastic APM module, you create a processor.
+A processor is a function that takes a `client` instance as well as an event (an error or a transaction),
```
jalvz: lingo in this side of the pipeline is still `exception`, not `error` (`capture_exception` etc)

docs/flask.asciidoc
beniwohli: The sentence definitely feels awkward right now, but your version isn't quite correct I think. Maybe

> Passing `logging=True` to the `ElasticAPM` constructor will make the agent automatically log all log messages from Python's built-in `logging` module.

docs/flask.asciidoc
beniwohli: You're right of course!

docs/flask.asciidoc
beniwohli: Is

> To force the agent to send data in debug mode

better?

jalvz: yes, totally. because it is in connection with previous sentence, right? so makes sense to use the same wording.

docs/flask.asciidoc
beniwohli: That is actually quite confusing. In Django, both errors and transactions aren't recorded during debug mode. In flask, only errors aren't recorded.

I'll unify the two to both have Django's current behaviour

PR #59, opened 2017-09-26
turned `servers` list setting into single `server` string setting

beniwohli:
Also removed logic that would send events to multiple servers

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -58,20 +58,17 @@ In less regexy terms:
 Your app name must only contain characters from the ASCII alphabet, numbers, dashes, underscores and spaces.
 
 [float]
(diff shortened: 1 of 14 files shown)


docs/configuration.asciidoc
```diff
@@ -105,16 +102,16 @@ Note however that this can have adverse effects on performance.
...
+Both the agents and the APM server have to be configured with the same secret token.
 One example to generate a secure secret token is:
 
 [source,bash]
 ----
 python -c "import uuid; print(str(uuid.uuid4()))"
 ----
 
-WARNING: secret tokens only provide any security if your APM servers use TLS.
+WARNING: secret tokens only provide any security if your APM server use TLS.
```
simitt: `server uses`

PR #61, opened 2017-09-26
finally killed the root trace

beniwohli:
This change was agreed upon a while back, but fell between the cracks


beniwohli:
@jalvz rebased this on top of master after merging the CHANGELOG branch and added an entry.

PR #62, opened 2017-09-26
Prepare jenkins.yml files for Infra tests.

simitt:
@elasticdog fyi


simitt:
@beniwohli fyi: after merging this, the Jenkins tests will break until @elasticdog has set up the new Jenkins handling.

PR #65, opened 2017-09-28
added response data for Flask transactions

beniwohli:


beniwohli:
@ruflin good point, added.

PR #68, opened 2017-10-02
added timer to trigger send if no incoming transactions trigger it

beniwohli:
This should alleviate a point of confusion when users expected that we
send data every x seconds. Instead, we only sent data every x seconds
if there were incoming transactions. If no transactions were happening,
no data was sent (including existing transactions on the queue).

This PR aligns the behaviour with user expectations. The trade-offs are:

 * slightly more resource consumption due to the timer thread
 * collection can either happen on the main thread, or on the timer
   thread, depending on whether the timer or an incoming transaction
   triggers collection.


elasticapm/base.py
```diff
@@ -456,6 +462,8 @@ def set_transaction_extra_data(self, data, _key=None):
 
     def close(self):
         self._traces_collect()
+        if self._send_timer:
```
simitt: Is it possible that a new transaction is started during the time that `close` is processing? If so, it looks like it could happen that the `end_transaction` could trigger a new `_start_send_timer`. This is probably an edge case that shouldn't be too bad though, because I assume when `close()` is called the whole application is going to be restarted..

beniwohli: exactly. Outside of test scenarios, `.close()` is only called by application shutdown logic. At this point, the application shouldn't accept any new transactions.

PR #69, opened 2017-10-03
added Django-2.0, and added some builds to jenkins exclusion

beniwohli:


.travis.yml
```diff
@@ -88,7 +88,7 @@ jobs:
       python: 3.4
     - stage: tests
       script: bash $RUN_SCRIPT
-      env: WEBFRAMEWORK=django-master
+      env: WEBFRAMEWORK=django-2.0
```
simitt: Did you remove testing `django-master` with `python-3.4` on purpose?  (just asking because for 3.5 and 3.6 you added it in addition..)

beniwohli: yes, Django 2.1 will only support Python 3.5+, and master already fails on 3.4

PR #70, opened 2017-10-03
fix two test failures that happened sometimes if tests ran in random order

beniwohli:


tests/contrib/django/django_tests.py
```diff
@@ -462,6 +462,7 @@ def test_404_middleware(self):
             self.assertEquals(request['body'], None)
 
     def test_404_middleware_with_debug(self):
+        self.elasticapm_client.config.debug = False
```
simitt: Does it make a difference for other tests if this is set to `true` or `false`?  
If so I suggest to make use of a `setUp()` function to set it to a default value, and then explicitly override it in the test cases you want another value to be set.

If not, please disregard this comment.

beniwohli: The default is `False`, and for most tests it doesn't make a difference. But yes, `setUp()` would probably have been a better place. I'll fix it in an upcoming PR.

PR #73, opened 2017-10-06
ensure TracingMiddleware and LoggingHandler get the client from the app registry

beniwohli:
While this is mostly helpful for testing, it also reduces our reliance on the `client` / `get_client()` globals (albeit by replacing it with a dependency on the global Django app registry).


beniwohli:
Turns out this can't really live without #71

PR #74, opened 2017-10-11
added support for coverage and set up codecov

beniwohli:


Makefile
```diff
@@ -7,13 +8,12 @@ flake8:
 	flake8 elasticapm
 
 test:
-	if [ "$$TRAVIS_PYTHON_VERSION" != "3.5" ]; then \
-	py.test --ignore=tests/asyncio; \
-	else py.test; fi
+	if [[ "$$TRAVIS_PYTHON_VERSION" =~ ^(3.5|3.6|nightly|pypy3)$$ ]] ; then \
```
ruflin: Makefile magic ...

codecov.yml
```diff
@@ -0,0 +1,46 @@
+# Enable coverage report message for diff on commit
+# Docs can be found here: https://docs.codecov.io/v4.3.0/docs/commit-status
+coverage:
+  status:
+    project:
+      tests:
+        target: 100%
```
ruflin: Sounds like a pretty high target :-)

codecov.yml
```diff
@@ -0,0 +1,46 @@
...
+        # Overall coverage should never drop more then 0.5%
```
ruflin: comment is here but threshold is on line 20. also there the comment is only afterwards

codecov.yml
```diff
@@ -0,0 +1,46 @@
+# Enable coverage report message for diff on commit
```
ruflin: Now I remember that it's not a hidden file. It's one of the things I didn't like ;-)

codecov.yml
beniwohli: We'll see how it goes, but generally, I think all test code should actually run

ruflin: Ah got it, I misread that config option. 👍

PR #75, opened 2017-10-12
both zerorpc and gevent now support Python 3, so no need to special case them

beniwohli:


beniwohli:
Turns out we can't merge this just yet. With pypy, there are failures due to an incompatibility with python-msgpack (https://github.com/0rpc/zerorpc-python/issues/151). While the fix has been merged in python-msgpack quite a while ago, it hasn't made it into a release yet, and is not likely to happen soon (https://github.com/msgpack/msgpack-python/issues/244).

There's also an issue with gevent and Python 3.7 (nightly), but that should hopefully be resolved sometime closer to the 3.7 release date.

beniwohli:
It doesn't look likely that the fixes in msgpack will be released anytime soon, closing this for now

PR #80, opened 2017-10-23
Force pulling latest docker image for python version.

simitt:


simitt:
@beniwohli I removed the versioning for installing `flake8` and `isort` so that also jenkins will download the latest version in the future.

beniwohli:
@simitt can you rebase on master?

simitt:
@beniwohli done 
The tests now fail with this update as you already pointed out - I leave the fixes up to you.

beniwohli:
@simitt they shouldn't fail, master has the fixes from #81 applied. Guess we'll find out :D

simitt:
ah great, I missed that

PR #84, opened 2017-11-02
use processors for transactions, and test that it actually happens

beniwohli:


tests/processors/tests.py
```diff
@@ -222,3 +224,34 @@ def test_mark_in_app_frames():
     assert not frames[3]['in_app']
     assert frames[4]['in_app']
     assert not frames[5]['in_app']
+
+
+def dummy_processor(client, data):
+    return data
+
+
+@mock.patch('tests.processors.tests.dummy_processor', wraps=dummy_processor)
```
jalvz: this is totally up to you, but in this case is soooo easy to test without mocks, just have the dummy processor to actually do something....

beniwohli: @jalvz good point, fixed in 3573a6d

PR #85, opened 2017-11-02
normalize HTTP status code in transaction.result

beniwohli:
This normalizes the status code into status classes 2xx, 3xx, 4xx and 5xx.
The unmodified status code is still available in context.response.status_code

### DIFF
--- a/tests/contrib/flask/flask_tests.py
+++ b/tests/contrib/flask/flask_tests.py
@@ -110,6 +110,7 @@ def test_instrumentation(flask_apm_client):
     assert len(transactions) == 1
     transaction = transactions[0]
     assert transaction['type'] == 'request'
+    assert transaction['result'] == 'HTTP 2xx'
     assert 'request' in transaction['context']
     assert transaction['context']['request']['url']['raw'] == 'http://localhost/users/'
     assert transaction['context']['request']['method'] == 'POST'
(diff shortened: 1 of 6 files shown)


tests/contrib/flask/flask_tests.py
```diff
@@ -144,11 +145,23 @@ def test_instrumentation_404(flask_apm_client):
...
 
+def test_non_standard_http_status(flask_apm_client):
+    with mock.patch("elasticapm.traces.TransactionsStore.should_collect") as should_collect:
+        should_collect.return_value = False
+        resp = flask_apm_client.app.test_client().get('/non-standard-status/')
+    assert resp.status == "0 fail", resp.response
+    assert resp.status_code == 0, resp.response
+
+    transactions = flask_apm_client.client.instrumentation_store.get_all()
+    assert transactions[0]['result'] == '0 fail'  # "0" is prepended by Werkzeug BaseResponse
```
simitt: Does this mean that every `transaction.result` not coming from a `flask` or `django` app has a `0 ` prefix no matter what? Could you get rid of this somehow?

beniwohli: Not exactly. It means that if a flask user returns a non-integer HTTP status code (which is all kinds of wrong in the first place), flask will prepend this `0` to it. Not sure why, but I'm not hugely concerned about it, since it only happens when the user code doing something completely wrong

simitt: oh yes, it is a flask test.. then I am with you to not worry about this too much.

PR #89, opened 2017-11-03
unexplode test matrix now that travis supports ordering of stages

beniwohli:
This basically reverts the exploding we did in #5


beniwohli:
FYI @watson, in case that stage ordering was also the only reason why you exploded your matrix

PR #90, opened 2017-11-06
added `uninstrument` function to reverse instrumentation

beniwohli:
This allows instrumentation to be reversible, e.g. to disable instrumentation on the fly during tests, or even in production.

### DIFF
--- a/elasticapm/instrumentation/control.py
+++ b/elasticapm/instrumentation/control.py
@@ -1,6 +1,23 @@
+import threading
+
 from elasticapm.instrumentation import register
+
+_lock = threading.Lock()
 
 
 def instrument():
-    for obj in register.get_instrumentation_objects():
-        obj.instrument()
+    """
+    Instruments all registered methods/functions with a wrapper
+    """
+    with _lock:
+        for obj in register.get_instrumentation_objects():
+            obj.instrument()
+
+
+def uninstrument():
+    """
+    If present, removes instrumentation and replaces it with the original method/function
+    """
+    with _lock:
+        for obj in register.get_instrumentation_objects():
+            obj.uninstrument()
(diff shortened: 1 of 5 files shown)


elasticapm/instrumentation/control.py
```diff
@@ -1,6 +1,18 @@
...
 from elasticapm.instrumentation import register
 
 
+_lock = threading.Lock()
+
+
 def instrument():
-    for obj in register.get_instrumentation_objects():
-        obj.instrument()
+    with _lock:
```
roncohen: Is it important that it's thread safe? Seems like you're in a bad situation if you have different threads calling instrument/uninstrument simultaneously

elasticapm/instrumentation/packages/base.py
```diff
@@ -208,6 +208,16 @@ def instrument(self):
                          self.name, ex)
         self.instrumented = True
 
+    def uninstrument(self):
+        if not self.instrumented or not self.originals:
+            return
+        for module, method in self.get_instrument_list():
+            if (module, method) in self.originals:
```
roncohen: How about looping over `self.originals` instead?

roncohen:
Nice work @beniwohli. Super useful to have this in.

elasticapm/instrumentation/control.py
beniwohli: The main concern is threaded workers that could all start instrumenting in parallel before the `instrumented` flag is set. You'd end up with multiple layers of wrapping. Granted, we've never heard of such a case, but that's possibly because most people don't uses threaded workers.

Running tests in parallel was another concern (although those are usually run in parallel using processes, not threads, so probably not an issue).

elasticapm/instrumentation/packages/base.py
beniwohli: my goal was to mirror `instrument()` as closely as possible. Looping over `self.originals` is slightly more efficient in some cases (when we only were able to wrap part of the instrument list), but I think mirroring the logic of `instrument()` has more value when it comes to readability.

PR #94, opened 2017-11-08
move docker stuff used for Jenkins testing into tests folder [WIP]

beniwohli:
This necessitates to mount the project folder as a volume instead of `ADD`ing it at build time. The benefit of this is that we don't need to build new docker images on every run, the drawback is that
we need to jump through some hoops to ensure that files created during the tests use the same user as Jenkins.


tests/scripts/docker/docs.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 set -e
 
+cd tests
+
 docker build -t "lint-docs" -f ./DockerfileDocs .
-docker run -i "lint-docs"
+docker run -e LOCAL_USER_ID=`id -u $USER` -v "$(dirname $(pwd))":/app -i "lint-docs"
```
dliappis: Since this is running bash, which what you have here is perfectly correct, you could also simplify it by just using `-e LOCAL_USER_ID=$UID`.

One style comment, the backticks have been deprecated for some time in favor of `$()` in [POSIX shells](http://pubs.opengroup.org/onlinepubs/9699919799/xrat/V4_xcu_chap02.html#tag_23_02_06_03). It's recommended to use `$( )` instead.

tests/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+USER_ID=${LOCAL_USER_ID:-1001}
+
+echo "Starting with UID : $USER_ID"
+useradd --shell /bin/bash -u $USER_ID -o -c "" -m user
+export HOME=/home/user
+
+exec /usr/local/bin/gosu user "$@"
```
dliappis: `gosu` is great!

If you ever want to simplify things without having to download/check gpg (and potentially run into licensing issues -> only an issue with public docker images) the good old `chroot` can do the job as well:

```
exec chroot --userspec=$USER_ID --skip-chdir "$@"
```

tests/entrypoint.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+USER_ID=${LOCAL_USER_ID:-1001}
+
+echo "Starting with UID : $USER_ID"
+useradd --shell /bin/bash -u $USER_ID -o -c "" -m user
```
dliappis: It's always a good idea to make the user part of gid 0 inside Docker.
This is anyway what all users have (even when you run a container with `--user <random_id>`) due to container user being a member of the root group.

Details [here](https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-specific-guidelines)

tests/DockerfileDocs
```diff
@@ -0,0 +1,18 @@
+FROM debian:8
```
dliappis: Why not `debian:9`? jessie is really old and stretch has been out for some time. You will benefit from more up-to-date python versions at least.

tests/DockerfileDocs
```diff
@@ -0,0 +1,18 @@
...
+		rm -rf /var/lib/apt/lists/*
+
+RUN gpg --keyserver ha.pool.sks-keyservers.net --recv-keys B42F6819007F00F88E364FD4036A9C25BF357DD4
+RUN curl -o /usr/local/bin/gosu -SL "https://github.com/tianon/gosu/releases/download/1.10/gosu-$(dpkg --print-architecture)" \
+    && curl -o /usr/local/bin/gosu.asc -SL "https://github.com/tianon/gosu/releases/download/1.10/gosu-$(dpkg --print-architecture).asc" \
+    && gpg --verify /usr/local/bin/gosu.asc \
+    && rm /usr/local/bin/gosu.asc \
+    && chmod +x /usr/local/bin/gosu
+
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
```
dliappis: Is the entrypoint used? There's only `CMD` invoked later on.

beniwohli: Great catch. It should use the entrypoint, as the user switching happens there

tests/entrypoint.sh
beniwohli: Looks like the `chroot` in our base image doesn't support `--skip-chdir`. The base images that we use are all built on top of Jessie: https://hub.docker.com/_/python/

    root@10dbaa600623:/# chroot --skip-chdir
    chroot: unrecognized option '--skip-chdir'
    Try 'chroot --help' for more information.
    root@10dbaa600623:/# chroot --version
    chroot (GNU coreutils) 8.23

Any idea for a good workaround? I would love to get all the `gosu` install boilerplate out of the Dockerfile :D

beniwohli:
jenkins, test this

tests/entrypoint.sh
dliappis: @beniwohli Is there any particular reason why you should base your images on the `jessie` release which was released in 2015 as opposed to `stretch` i.e. debian:9?

beniwohli: Those are the official Python docker images, so I don't have control over it. The `python:3.7` image is built on Stretch, but I guess they don't want to risk unnecessary breakage by updating the images for older versions to Stretch.

dliappis: @beniwohli Ah I was too fast. So it's the python image that is based on debian:8. Have you considered using `python:3.6-stretch` or `python:2.7-stretch`?

```
$ docker run --rm -ti python:3.6-stretch chroot --help
Usage: chroot [OPTION] NEWROOT [COMMAND [ARG]...]
  or:  chroot OPTION
Run COMMAND with root directory set to NEWROOT.

  --groups=G_LIST        specify supplementary groups as g1,g2,..,gN
  --userspec=USER:GROUP  specify user and group (ID or name) to use
  --skip-chdir           do not change working directory to '/'
      --help     display this help and exit
      --version  output version information and exit

If no command is given, run '${SHELL} -i' (default: '/bin/sh -i').

GNU coreutils online help: <http://www.gnu.org/software/coreutils/>
Report chroot translation bugs to <http://translationproject.org/team/>
Full documentation at: <http://www.gnu.org/software/coreutils/chroot>
or available locally via: info '(coreutils) chroot invocation'
```

beniwohli: Ugh, turns out there are no Stretch-based images for `pypy:2` and `pypy:3` :(

PR #103, opened 2017-11-17
stack collection optimizations [WIP]

beniwohli:
This PR tries to mitigate what turned out to be a major bottleneck: collecting stack frames for every single trace of a transaction. 

The most significant win came from not using `inspect.stack(0)`, and instead use `inspect.currentframe()` and then walk up the stack. This is due to the fact that deep inside the call stack of `inspect.stack()`, an `os.exists()` is called for every frame's filename. This causes a lot of IO, which can be avoided.

Other gains mostly come from using iterators instead of lists wherever possible, and avoid work unless it is absolutely necessary.

In a (very) synthetic benchmark, these changes reduces the overhead by about factor 7 (from ~28ms per transaction to ~4ms). The benchmarked transaction has 16 traces, each of which has about ~80 frames (due to using py.test / pytest-benchmark, which bloats the call stack quite a bit).

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -12,10 +12,10 @@ Large portions are
 from __future__ import absolute_import
 
 import datetime
-import itertools
 import logging
 import os
 import platform
+import re
 import socket
 import sys
 import threading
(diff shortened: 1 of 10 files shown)


beniwohli:
Many of the changes that weren't too intrusive got merged in other PRs. Closing this one.

PR #105, opened 2017-11-27
renamed `server` configuration variable to `server_url`

beniwohli:
This aligns the Python agent to agents in other languages

This PR also simplifies the logic for sending data, and
uses a test server instead of mocks for many tests

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -19,7 +19,6 @@ import socket
 import sys
 import threading
 import time
-import warnings
 import zlib
 
 import elasticapm
@@ -106,6 +105,13 @@ class Client(object):
         self.logger = logging.getLogger('%s.%s' % (cls.__module__, cls.__name__))
         self.error_logger = logging.getLogger('elasticapm.errors')
         self.state = ClientState()
+
+        self.instrumentation_store = None
+        self.processors = []
+        self.filter_exception_types_dict = {}
+        self._send_timer = None
+        self._transports = {}
+
         self.config = Config(config, default_dict=defaults)
         if self.config.errors:
             for msg in self.config.errors.values():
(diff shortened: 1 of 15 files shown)


elasticapm/base.py
```diff
@@ -496,10 +481,11 @@ def handle_transport_fail(self, exception=None, **kwargs):
     def _collect_transactions(self):
         self._stop_send_timer()
         transactions = []
-        for transaction in self.instrumentation_store.get_all():
-            for processor in self.processors:
-                transaction = processor(self, transaction)
-            transactions.append(transaction)
+        if self.instrumentation_store:
```
jalvz: this `if` is related to one of the subtle bugs you mentioned? isn't possible to init `instrumentation_store` to something not `None`?, like eg. in a way that if you can `get_all()` on it you get an empty list...

elasticapm/base.py
```diff
@@ -465,8 +449,9 @@ def close(self):
         self._collect_transactions()
         if self._send_timer:
             self._stop_send_timer()
-        for url, transport in self._transports.items():
+        for url, transport in list(self._transports.items()):
             transport.close()
+            self._transports.pop(url)
```
jalvz: hmm good find! (i think...) 😁

elasticapm/base.py
beniwohli: The problem here was if we had a configuration error, we would bail out early in `__init__`, before the `instrumentation_store` attribute was even set. Then when we try to collect transactions, the code explodes. Due to mocking, that code path was never hit in tests.

I don't think it's worth the effort to initialize the `instrumentation_store` to a `DummyStore` or somesuch, only to change it to the real store a few lines down (if the config validates), tbh.

jalvz: i see, makes sense. thanks!

PR #97, opened 2017-11-10
added setting to disable SSL certificate validation

beniwohli:

PR #108, opened 2017-12-01
added setting to disable SSL certificate verification

beniwohli:
replaces #97


beniwohli:
@simitt changed `verify_certificate` to `verify_server_cert`. `verify_ssl` is given by the library we use, so can't align that one.

PR #109, opened 2017-12-04
updated django versions in test requirements

beniwohli:


tests/requirements/requirements-django-1.11.txt
```diff
@@ -1,2 +1,2 @@
-Django>=1.11b1,<1.12
+Django>=1.11,<2.0
```
simitt: Does it make sense to have a `1.11` file that installs Django `<2.0` ?

beniwohli: There never will be a Django 1.12, so saying `<2.0` is equivalent to `1.12`, but makes more sense with regards to the Django version history

PR #112, opened 2017-12-04
api docs and API simplifications [WIP]

beniwohli:
This changes the API of exception tracking. It gets rid of the somewhat opaque `data` argument and instead uses explicit keyword arguments for `context`, `custom` etc.

It also removes Django-template-specific code in the exception tracking. It only worked when `TEMPLATE_DEBUG` was set to true (which shouldn't be the case in production code), and was never exposed in the UI.


docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+
+ * `config`: A dictionary, with key/value configuration. For the possible configuration keys, see <<configuration, Configuration>>.
+ * `**defaults`: default values for configuration. These can be omitted in most cases, and take the least precedence.
+
+[[error-api]]
+=== Error API
+
+[break]
+[[client-api-capture-exception]]
+==== `Client.capture_exception()`
```
simitt: This `Client.capture_exception()` looks like a class method, but then you actually call `client.capture_exception()`

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
+[[api]]
+== Public API
```
simitt: Why do you split it into `Client`, `Error` and `Transaction` API when you actually need a `client` instance for all of them?

How about calling it a `Client API` and then only use `Error` and `Transaction` later on?

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+client.capture_message('Billing process succeeded.')
+----
+
+[break]
+===== Arguments
+
+ * `message`: The message as a string.
+ * `param_message`: Alternatively, a parametrized message as a dictionary.
+    The dictionary contains two values: `message`, and `params`.
+    This allows the APM server to group messages together that have share the same
```
simitt: This sentence is a bit malformed: `This allows the APM server to group messages together that have share the same parametrized message.`

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+[source,python]
+----
+client.capture_message('Billing process succeeded.')
+----
+
+[break]
+===== Arguments
+
+ * `message`: The message as a string.
+ * `param_message`: Alternatively, a parametrized message as a dictionary.
```
simitt: What if one doesn't provide `message` nor `param_message`? Maybe a sentence like `Either message or param_message need to be provided.`.. Or is it ok to not provide any of them?

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+ * `context`: A dictionary with contextual information. This dictionary must follow the 
+    https://www.elastic.co/guide/en/apm/server/current/error-api.html#error-context-schema[Context] schema definition.
+ * `custom`: A dictionary of custom data you want to attach to the event.
+
+Returns the id of the message as a string.
+
+[[transaction-api]]
+=== Transaction API
+
+[break]
```
simitt: Although the usage is straightforward, it would be nice to also have an example in for the `transaction` cmds, as you also have examples for `client` and `error` related usage.

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+===== Arguments
+
+ * `name`: A string describing the name of the transaction, e.g. `process_order`.
+   This is typically the name of the view/controller that handles the request, or the route name.
+ * `result`: A string describing the result of the transaction.
+   This is typicaly the HTTP status code, or e.g. `'success'` for a background task.
+
+[break]
+[[api-elasticapm-instrument]]
+==== `elasticapm.instrument()`
```
simitt: This doesn't seem to be related to `transaction` specific cmds, so maybe you want to move it to its own section like `Auto-Instrumentation`.

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+==== `elasticapm.set_transaction_data()`
+
+Attach custom contextual data to current transaction.
+Supported frameworks will automatically attach information about the HTTP request and the logged in user.
+You can attach further data using this function.
+
+===== Arguments
+
+ * `data`: A dictionary with the data to be attached. This should be a flat key/value `dict` object.
+ * `_key`: The key to use for this `dict` object. Defaults to `custom`.
```
simitt: Does the underscore in `_key` indicate it is private? The documented possibility of modifying the variable looks somehow weird.

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
++
+NOTE: This should only be overridden by framework integrations.
```
simitt: It would be great to add `optional` or `mandatory` to every parameter you describe. Sometimes you mention it is optional, sometimes you mention a default, sometimes none of both.

docs/api.asciidoc
beniwohli: This is one of the things I struggled with when structuring the docs. I though by putting the "Client API" first, the reader will know further down how to get a client.

docs/api.asciidoc
beniwohli: I used a capital C in all the titles that of methods on `Client`, I thought it would be clear that it means "method on the Client class", didn't really think of the issue that it could be mistaken as a class method.

Class methods (or static methods) are used relatively seldom in Python, so I don't think most readers would be confused.

docs/api.asciidoc
beniwohli: Excellent point, one of the two is needed

docs/api.asciidoc
simitt: One definitly should find his way around by the explanations you provide. I just think it is somehow confusing to describe it as three seperate APIs. I would only distinguish between the `elasticapm` and the `elasticapm.Client` API. And the split up `elasticapm.Client` into 
- getting an instance
- error capturing
- transaction capturing

docs/api.asciidoc
```diff
@@ -0,0 +1,159 @@
...
+
+Instruments libraries automatically.
+This includes a wide range of standard library and 3rd party modules.
+A list of instrumented modules can be found in `elasticapm.intrumentation.register`.
+This function should be called as early as possibly in the startup of your application.
+For supported frameworks, this is called automatically.
+
+[break]
+[[api-set-transaction-name]]
+==== `elasticapm.set_transaction_name()`
```
simitt: Why is `begin_transaction` and `end_transaction` a method of `client` byt `set_transaction_name` and `set_transaction_data` are not?

beniwohli: in most places you'd use `set_transaction_name` & Co, you don't have a `client` instance or the current transaction. With this, we can put the difficulty of getting the current transaction into our code, instead of leaving it to the user to acquire the current transaction somehow.

At the moment, the current transaction is stored on a thread local. But that only works for singlethreaded code, so we probably need to change that in the future. The way it is now, that change would be nicely encapsulated.

docs/api.asciidoc
simitt: ok, maybe it is just me then, coming from a ruby background.

docs/api.asciidoc
simitt: How can you not have a current transaction when you want to override the name of the current transaction? 
I ofc don't have the insights into the agent, but from a pure user perspective I'd somwhow expect that `set_transaction_name` and `begin/stop_transaction` is within the same context. 
But this might be somehow out of scope for this PR anyways, just realized that when looking at the docs.

docs/api.asciidoc
```diff
@@ -0,0 +1,191 @@
...
+   If not provided, it will be captured automatically, if `capture_message()` was called in an `except` block.
+ * `date`: A `datetime.datetime` object representing the occurrence time of the error.
+   If left empty, it defaults to `datetime.datetime.utcnow()`.
+ * `context`: A dictionary with contextual information. This dictionary must follow the 
+    https://www.elastic.co/guide/en/apm/server/current/error-api.html#error-context-schema[Context] schema definition.
+ * `custom`: A dictionary of custom data you want to attach to the event.
+
+Returns the id of the message as a string.
+
+NOTE: either the `message` or the `param_message` argument is required.
```
simitt: Could you start with a capital letter here?

docs/api.asciidoc
```diff
@@ -0,0 +1,191 @@
+[[api]]
+== Public API
+
+The Elastic APM Python agent has several public APIs.
+Most of these APIs are not needed when using one of our supported frameworks (Django and Flask),
```
simitt: How about changing this to something like `Most of the public API functionality is not needed when.. `

docs/api.asciidoc
simitt: For `begin_transaction` and `end_transaction` it is not very clear to me if the params are required or not.

docs/api.asciidoc
```diff
@@ -0,0 +1,191 @@
...
+
+[float]
+[[api-elasticapm-instrument]]
+==== `elasticapm.instrument()`
+
+Instruments libraries automatically.
+This includes a wide range of standard library and 3rd party modules.
+A list of instrumented modules can be found in `elasticapm.intrumentation.register`.
+This function should be called as early as possibly in the startup of your application.
+For supported frameworks, this is called automatically. Example:
```
simitt: A link to the docs listing the supported framework would be a nice to have here.

PR #116, opened 2017-12-14
Implemented random sampling

beniwohli:
Given a sample rate, this will randomly choose transactions
to be sampled. If a transaction isn't sampled, no spans will
be collected, as well as no context information

### DIFF
--- a/elasticapm/instrumentation/packages/base.py
+++ b/elasticapm/instrumentation/packages/base.py
@@ -8,7 +8,7 @@ logger = logging.getLogger(__name__)
 
 
 # We have our own `BoundFunctionWrapper` and `FunctionWrapper` here because
-# we want them to be able to now about `module` and `method` and supply it in
+# we want them to be able to know about `module` and `method` and supply it in
 # the call to the wrapper.
 
 class OriginalNamesBoundFunctionWrapper(wrapt.BoundFunctionWrapper):
@@ -219,7 +219,8 @@ class AbstractInstrumentedModule(object):
         self.originals = {}
 
     def call_if_sampling(self, module, method, wrapped, instance, args, kwargs):
-        if not get_transaction():
+        transaction = get_transaction()
+        if not transaction or not transaction.is_sampled:
             return wrapped(*args, **kwargs)
         else:
             return self.call(module, method, wrapped, instance, args, kwargs)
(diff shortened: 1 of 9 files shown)


elasticapm/traces.py
```diff
@@ -230,14 +244,14 @@ def decorated(*args, **kwds):
     def __enter__(self):
         transaction = get_transaction()
 
-        if transaction:
+        if transaction and transaction.is_sampled:
             transaction.begin_span(self.name, self.type, self.extra,
```
roncohen: Do you need the check for `in_sampled` it both here and inside `begin_span`?

elasticapm/instrumentation/packages/base.py
```diff
@@ -219,7 +219,8 @@ def uninstrument(self):
         self.originals = {}
 
     def call_if_sampling(self, module, method, wrapped, instance, args, kwargs):
-        if not get_transaction():
+        transaction = get_transaction()
+        if not transaction or not transaction.is_sampled:
             return wrapped(*args, **kwargs)
```
roncohen: you have the check in `begin_span`. Do you need it here too?

elasticapm/traces.py
```diff
@@ -102,7 +112,8 @@ def to_dict(self):
             'context': self._context,
             'spans': [
```
roncohen: I think it's worth considering removing the `spans` key entirely to clearly indicate this is an unsampled transaction.

elasticapm/instrumentation/packages/base.py
beniwohli: not 100% sure tbh. I'll step through it with the debugger

beniwohli: After taking a closer read of the code, I think it's worth doing the check in both places.

`call_if_sampling` is called for stuff we instrument automatically. Checking here avoids even calling the `capture_span` decorator and doing other stuff we do in the wrappers (e.g. parsing the SQL query for an `sql` span).

the check in `capture_span.__enter__` is useful for cases where code is instrumented manually using the `capture_span` decorator.

The third check, in `Transaction.begin_span` seems to be superfluous, however, as it is only called from `capture_span.__enter__`. Good catch!

PR #117, opened 2017-12-16
added configuration options for collecting local variables and source code

beniwohli:
this allows to enable/disable collecting of local variables and source code independent from each other, for exceptions/logs and transactions

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -233,6 +233,40 @@ This setting can be overridden on an individual basis by setting the `extra`-key
 logger.info('something happened', extra={'stack': False})
 ----
 
+[float]
+[[config-collect-local-variables]]
+==== `collect_local_variables`
+
+|============
+| Environment                           | Django/Flask              | Default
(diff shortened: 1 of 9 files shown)


docs/configuration.asciidoc
```diff
@@ -233,6 +233,40 @@ This setting can be overridden on an individual basis by setting the `extra`-key
...
+can have adverse effects for the performance of your service.
+
+[float]
+[[config-collect-source]]
+==== `collect_source`
+
+|============
+| Environment                  | Django/Flask     | Default
+| `ELASTIC_APM_COLLECT_SOURCE` | `COLLECT_SOURCE` | `all`
+|============
```
simitt: Why did you choose `all` as default? The note states that this can have a sever overhead, so I personally would expect either `none` or `errors` as default here.

beniwohli: Because we want to have the "full experience" out of the box.

I'm not sure how severe the overhead is tbh, now that we have an LRU cache in front of it. Needs some testing I guess

PR #130, opened 2018-01-09
fix shortening of local vars

beniwohli:
This adds configuration option for shortening local variables, as well as a fix for shortening lists

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -131,7 +131,12 @@ class Client(object):
             lambda: self._get_stack_info_for_trace(
                 stacks.iter_stack_frames(),
                 with_source_context=self.config.collect_source in ('all', 'transactions'),
-                with_locals=self.config.collect_local_variables in ('all', 'transactions')
+                with_locals=self.config.collect_local_variables in ('all', 'transactions'),
+                locals_processor_func=lambda local_var: varmap(lambda k, v: shorten(
+                    v,
+                    list_length=self.config.local_var_list_max_length,
+                    string_length=self.config.local_var_max_length,
+                ), local_var)
             ),
             collect_frequency=self.config.transaction_send_frequency,
             sample_rate=self.config.transaction_sample_rate,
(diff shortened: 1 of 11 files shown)


beniwohli:
One issue with this is that it doesn't handle dicts with lots of entries. I'm not sure how we could handle this correctly, due to insertion order not being preserved previous to Python 3.6. So we can't add a `(X more elements)` at the end of the dict.

roncohen:
we used to handle it by sorting them and then only taking the first X number of items

elasticapm/base.py
```diff
@@ -413,16 +418,21 @@ def _build_msg_for_logging(self, event_type, date=None, context=None, custom=Non
         log = event_data.get('log', {})
         if stack and 'stacktrace' not in log:
             if stack is True:
-                frames = stacks.iter_stack_frames()
+                frames = stacks.iter_stack_frames(skip=3)
```
roncohen: why do we need this now?

beniwohli: It's a bug I noticed when looking at the frames in the test. It included some frames from the agent, so the first frame wasn't where you actually had the `logger` call.

I suspect there are more issues with too many frames in other places we collect them, but I'll leave that for another PR.

elasticapm/utils/__init__.py
```diff
@@ -27,23 +27,23 @@
...
     if isinstance(var, dict):
         ret = dict((k, varmap(func, v, context, k)) for k, v in compat.iteritems(var))
     elif isinstance(var, (list, tuple)):
-        ret = [varmap(func, f, context, name) for f in var]
+        ret = func(name, [varmap(func, f, context, name) for f in var])
     else:
         ret = func(name, var)
-    del context[objid]
+    context.remove(objid)
     return ret
```
roncohen: instead of the change on line 43 and instead of the `else` clause here, could this just be:

`return func(name, var)`

?

beniwohli: that would run `func` on `dict` as well. Which isn't entirely wrong, but `shorten` doesn't have support for dicts yet. I'll try to add support for `dict` in `shorten` in a different PR and will change it to your suggested logic one that's in place

PR #132, opened 2018-01-10
use pytest-xdist to distribute tests on multiple CPUs

beniwohli:


beniwohli:
closing this. We run into concurrency issues when testing against databases. Also, the benefit is relatively small, as most of the time is spent in setting up the test environment, not in actually running the tests

PR #133, opened 2018-01-15
refactor context setters, and document them properly

beniwohli:
This also adds functionality to copy transaction context data to errors that happen during the transaction.

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -171,10 +171,10 @@ elasticapm.set_transaction_name('myapp.billing_process')
  * `name`: (*required*) A string describing name of the transaction
 
 [float]
-[[api-set-transaction-data]]
-==== `elasticapm.set_transaction_data()`
+[[api-set-custom-context]]
+==== `elasticapm.set_custom_context()`
 
-Attach custom contextual data to current transaction.
+Attach custom contextual data to the current transaction and errors.
 Supported frameworks will automatically attach information about the HTTP request and the logged in user.
 You can attach further data using this function. Example:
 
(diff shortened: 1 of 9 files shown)


docs/api.asciidoc
```diff
@@ -185,8 +185,49 @@ import elasticapm
 elasticapm.set_transaction_data({'billing_amount': product.price * item_count})
```
simitt: Is it really still `set_transaction_data` and not `set_context`?

tests/client/client_tests.py
```diff
@@ -650,3 +650,18 @@ def test_transaction_max_span_nested(should_collect, elasticapm_client):
...
+    elasticapm.set_custom_context({'a': 'b'})
+    elasticapm.set_user_context(username='foo', email='foo@example.com', user_id=42)
+    elasticapm_client.capture_message('x', custom={'foo': 'bar'})
+    transaction = elasticapm_client.end_transaction('test', 'OK')
+    message = elasticapm_client.events[0]['errors'][0]
+    assert message['context']['custom'] == {'a': 'b', 'foo': 'bar'}
+    assert message['context']['user'] == {'username': 'foo', 'email': 'foo@example.com', 'id': 42}
+    assert message['context']['tags'] == {'foo': 'baz'}
+    assert 'a' in transaction.context['custom']
+    assert 'foo' not in transaction.context['custom']
```
simitt: Maybe I am missing it but I don't see a test that checks that you can call the update functions multiple times. as described in the api docs.

docs/api.asciidoc
beniwohli: great catch!

tests/client/client_tests.py
beniwohli: Excellent point, I added some tests :)

PR #137, opened 2018-01-19
use shell globs on file name for include/exclude patterns

beniwohli:
The previously used module path was very brittle. Especially in the
case of Django, where the default `include_path` was the set of
`INSTALLED_APPS`, this would have a lot of false positives, namely
all 3rd party Django apps.

Using the `lib/python` exclude patterns should work for all cases that
use virtualenvs or the system python.

### DIFF
--- a/elasticapm/contrib/django/client.py
+++ b/elasticapm/contrib/django/client.py
@@ -49,18 +49,11 @@ def get_client(client=None):
 
     if _client[0] != client:
         client_class = import_string(client)
-        instance = client_class(**get_client_default_config())
+        instance = client_class()
         if not tmp_client:
             _client = (client, instance)
         return instance
     return _client[1]
-
-
-def get_client_default_config():
-    return dict(
-        include_paths=_get_installed_apps_paths(),
-        exclude_paths={'django', 'elasticapm'}
-    )
 
 
 class DjangoClient(Client):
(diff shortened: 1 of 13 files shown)


elasticapm/contrib/django/client.py
```diff
@@ -49,20 +49,13 @@ def get_client(client=None):
...
+        instance = client_class()
         if not tmp_client:
             _client = (client, instance)
         return instance
     return _client[1]
 
 
-def get_client_default_config():
-    return dict(
-        include_paths=_get_installed_apps_paths(),
```
graphaelli: looks like `_get_installed_apps_paths` was only called from here, could remove that function too

graphaelli:
I'm unclear on what will happen on windows but noticed your TODO, not sure how you normally handle that.

elasticapm/contrib/django/client.py
beniwohli: excellent point!

beniwohli:
Regarding Windows, I did a quick test with Python 3.6

    >>> import fnmatch
    >>> print(fnmatch.__file__)
    C:\Users\IEUser\project\elastic\py36\lib\fnmatch.py
    >>> import django
    >>> print(django.__file__)
    C:\Users\IEUser\project\elastic\py36\lib\site-packages\django\__init__.py

So `['*\lib\*']` would work on Windows, but it's quite unspecific, so the chance for false positives is relatively high.

(`py36` in the above example is just the name of my virtualenv, so can't use that in the pattern).

I'll do some more tests with Python 2 and pypy and then add Windows-specific patterns.

beniwohli:
jenkins test this

beniwohli:
jenkins test this

PR #147, opened 2018-01-29
fixed several edge cases when loading lines from files, and added tests

beniwohli:
also, by iterating over the file and only encode the parts that we
need, loading lines from files is now about twice as fast


beniwohli:
jenkins test this

beniwohli:
jenkins test this

PR #148, opened 2018-01-29
fix `extract_signature` to work with sql.Composable objects

beniwohli:
see http://initd.org/psycopg/docs/sql.html

### DIFF
--- a/elasticapm/instrumentation/packages/dbapi2.py
+++ b/elasticapm/instrumentation/packages/dbapi2.py
@@ -121,6 +121,11 @@ def scan(tokens):
 
 
 def extract_signature(sql):
+    """
+    Extracts a minimal signature from a given SQL query
+    :param sql: the SQL statement
+    :return: a string representing the signature
+    """
     sql = sql.strip()
     first_space = sql.find(' ')
     if first_space < 0:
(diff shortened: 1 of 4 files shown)


beniwohli:
jenkins test this

beniwohli:
jenkins test this

elasticapm/instrumentation/packages/dbapi2.py
```diff
@@ -170,10 +175,17 @@ def executemany(self, sql, param_list):
         return self._trace_sql(self.__wrapped__.executemany, sql,
                                param_list)
 
+    def _bake_sql(self, sql):
+        """
+        Method to turn the "sql" argument into a string
+        """
+        return sql
```
simitt: The method description might be a bit confusing, as it actually does nothing in this class.
Would it do any harm if you'd actually move the `as_string` check here?

elasticapm/instrumentation/packages/dbapi2.py
```diff
@@ -121,6 +121,11 @@ def scan(tokens):
 
 
 def extract_signature(sql):
```
simitt: Not part of this PR, but out of interest, why is this not a method of `CursorProxy` ? Afaics this is only used within sql cursor proxy implementations.

elasticapm/instrumentation/packages/dbapi2.py
beniwohli: the `as_string` check is specific to psycopg2 (the Python PostgreSQL driver), so I think putting it in the psycopg2-specific subclass makes sense. However I agree that the docstring needs work.

elasticapm/instrumentation/packages/dbapi2.py
beniwohli: I thought about this as well, but into the other direction: remove `CursorProxy.extract_signature` and call `dbapi2.extract_signature()` directly where we need it.

I guess the idea was that if we instrument a SQL backend that needs a different parser, we could plug it in there. @roncohen might remember.

simitt: Thank you for giving some context. Might be worth giving it a thought at some point, but as mentioned unrelated to this PR.

PR #149, opened 2018-01-29
added option to skip instrumenting specific modules via env variable

beniwohli:
this adds back a feature that got removed by mistake, and adds a
test

🎶If you like it, you better put a test on it🎶


beniwohli:
jenkins test this

PR #159, opened 2018-02-13
truncate strings that are stored as keywords in ES (#159)

### LINKED ISSUE #156: Problem validating JSON document against schema - length must be <= 1024, but got 1523
Continuously seeing these errors in the log -

2018-02-13T05:06:49.104Z	ERROR	[request]	beater/handlers.go:384	error handling request	{"request_id": "79f54c15-f99e-4d1d-ab14-07bf622983c8", "error": "Problem validating JSON document against schema: I[#] S[#] doesn't validate with \"transaction#\"\n  I[#/transactions/29/context/request/url/full] S[#/properties/transactions/items/properties/context/properties/request/properties/url/properties/full/maxLength] length must be <= 1024, but got 1523"}

apm-server version is 6.2.1
elastic-apm agent version 2.0.0 with Flask.

Is there any config property that I need to change to let the apm server accept urls of a longer length?

beniwohli:
this should avoid schema errors for values that are too long

closes #156

closes #159

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -29,7 +29,7 @@ from elasticapm.transport.base import TransportException
 from elasticapm.utils import compat, is_master_process
 from elasticapm.utils import json_encoder as json
 from elasticapm.utils import stacks, varmap
-from elasticapm.utils.encoding import shorten, transform
+from elasticapm.utils.encoding import keyword_field, shorten, transform
 from elasticapm.utils.module_import import import_string
 
 __all__ = ('Client',)
@@ -111,6 +111,7 @@ class Client(object):
         self.filter_exception_types_dict = {}
         self._send_timer = None
         self._transports = {}
+        self._service_info = None
 
         self.config = Config(config, default_dict=defaults)
         if self.config.errors:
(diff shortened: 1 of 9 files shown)


beniwohli:
jenkins test this

elasticapm/utils/encoding.py
```diff
@@ -165,3 +168,9 @@ def shorten(var, list_length=50, string_length=200):
         # TODO: when we finish the above, we should also implement this for dicts
         var = list(var)[:list_length] + ['...', '(%d more elements)' % (len(var) - list_length,)]
     return var
+
+
+def keyword_field(string):
+    if not isinstance(string, compat.string_types) or len(string) <= KEYWORD_MAX_LENGTH:
+        return string
+    return string[:KEYWORD_MAX_LENGTH - 1] + u'…'
```
jahtalab: @beniwohli , Adding '…' would add 3 characters to the length which might exceed 1024 limit.

A separate issue: In my opinion '...' is not necessary, I would prefer adding a property to say the name was truncated and let the ui decide about the "ellipsis", but even that I think at this stage is not necessary.

elasticapm/base.py
```diff
@@ -327,45 +328,48 @@ def _send_remote(self, url, data, headers=None):
...
             }
+        self._service_info = result
         return result
 
     def get_process_info(self):
         return {
             'pid': os.getpid(),
             'argv': sys.argv,
-            'title': None,
+            'title': None,  # Note: if we implement this, the value needs to be wrapped with keyword_field
```
jahtalab: I like your note to your future self 😄 , I do that as well!

elasticapm/utils/encoding.py
beniwohli: The character I use there is the unicode ellipsis, so it's only one character :)

I'm not sure adding an additional field for every keyword field to indicate if it was truncated or not is feasible. Having to truncate a field should be an absolute edge case.

jahtalab: I see, sorry my python is a bit rusty 😄 !


It doesn't have to be a field for every keyword field. But my point was that I would prefer not adding anything to the fields that might have a meaning in different contexts and keeping what the data was originally

jahtalab: I'm going to approve the changes, then you can decide about the 'ellipsis'!

PR #164, opened 2018-02-20
replace subclassing of wrapt wrappers by using a partial

beniwohli:
We need to pass along the method and module names to the function
wrapper. This has so far been done by subclassing wrapt.FunctionWrapper.
While this works, it adds a lot of duplicate code that we need to sync
up every time we update wrapt. Using a partial should mean much less
maintenance work.


beniwohli:
pypy tests fail because somehow this change introduces two additional frames into the captures span stack traces, but only on PyPy. No idea why...

PR #169, opened 2018-02-23
initial elasticsearch-py instrumentation support 

beniwohli:
This adds instrumentation for an initial set of methods of the
elasticsearch client library, as well as matrix tests for version
2, 5 and 6.

Due to the still relatively high download numbers for elasticsearch-py 2.x, we also added instrumentation for this version. Download counts between Jan 1-Apr 5 2018:

| Major Version | Download Count|
|---|---|
|6|871349|
|5|751404|
|2|237812|
|1|152074|
|0|22718|

For many methods, we'd like to show index/doctype/id if available. Due to the ordering of these arguments changing between different versions of elasticsearch-py, we need a way to look up the ordering of the positional arguments depending on the major version. Unfortunately, this can't be done via introspection due to the `query_params` decorator "hiding" this information (see https://github.com/elastic/elasticsearch-py/issues/736).

### DIFF
--- a/elasticapm/instrumentation/packages/elasticsearch.py
+++ b/elasticapm/instrumentation/packages/elasticsearch.py
@@ -1,0 +1,199 @@
+from __future__ import absolute_import
+
+import json
+import logging
+
+import elasticapm
+from elasticapm.instrumentation.packages.base import AbstractInstrumentedModule
+from elasticapm.utils import compat
+
+logger = logging.getLogger(__name__)
+
+
+class ElasticsearchIndicesIntrumentation(AbstractInstrumentedModule):
+    name = 'elasticsearch.index'
+
+    instrument_list = [
+        ("elasticsearch.client.indices", "IndicesClient.create"),
+        ("elasticsearch.client.indices", "IndicesClient.get"),
+        ("elasticsearch.client.indices", "IndicesClient.open"),
+        ("elasticsearch.client.indices", "IndicesClient.close"),
+        ("elasticsearch.client.indices", "IndicesClient.delete"),
+        ("elasticsearch.client.indices", "IndicesClient.exists"),
+        ("elasticsearch.client.indices", "IndicesClient.analyze"),
+        ("elasticsearch.client.indices", "IndicesClient.refresh"),
+        ("elasticsearch.client.indices", "IndicesClient.flush"),
+    ]
+
+    def call(self, module, method, wrapped, instance, args, kwargs):
+        cls_name, method_name = method.split('.', 1)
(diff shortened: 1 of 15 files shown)


tests/.jenkins_framework.yml
```diff
@@ -8,3 +8,6 @@ FRAMEWORK:
   - flask-0.10
   - flask-0.11
   - flask-0.12
+  - elasticsearch-2
+  - elasticsearch-5
+  - elasticsearch-6
```
simitt: WDYT about either adding another Jenkins Job for testing driver versions, or renaming`framework` to something more generic like `external_library` so you can use this for generally testing libraries in different versions against different python versions?

beniwohli: yeah, definitely considering that :)

beniwohli:
jenkins please test this

beniwohli:
jenkins test this

elasticapm/instrumentation/packages/elasticsearch.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self):
+        super(ElasticsearchInstrumentation, self).__init__()
+        try:
+            from elasticsearch import VERSION
+            self.version = VERSION[0]
+        except ImportError:
+            self.version = None
+
+    def instrument(self):
+        if self.version and not 2 <= self.version < 7:
```
graphaelli: should instrumentation also be skipped if version is none, but with a different message?

tests/instrumentation/elasticsearch_tests.py
```diff
@@ -0,0 +1,314 @@
...
+    """Elasticsearch client fixture."""
+    client = Elasticsearch(hosts=os.environ['ES_URL'])
+    yield client
+    client.indices.delete(index='*')
+
+
+@pytest.mark.integrationtest
+def test_ping(elasticapm_client, elasticsearch):
+    elasticapm_client.begin_transaction('test')
+    search_query = {"query": {"term": {"user": "kimchy"}}}
```
graphaelli: unused

elasticapm/instrumentation/packages/elasticsearch.py
beniwohli: `version` is set to `None` only in the case of an `ImportError`. We handle that case in the super method with an appropriate log message, so I think it's ok to fall through.

https://github.com/elastic/apm-agent-python/blob/e74bd1a55b275e4d147699308c22785595c4636e/elasticapm/instrumentation/packages/base.py#L73-L77

beniwohli:
closing in favor of #191

PR #174, opened 2018-02-28
fix bug with OPTIONS requests and body capturing

### LINKED ISSUE #172: OPTIONS Request in Flask uses content-type of None
OPTIONS requests (CORS Preflight) are checked for content-type on the request object in the flask contrib module (`elasticapm/contrib/flask/utils.py`):

```
    if request.method not in ('GET', 'HEAD'):
        body = None
        if request.content_type == 'application/x-www-form-urlencoded':
            body = compat.multidict_to_dict(request.form)
        elif request.content_type.startswith('multipart/form-data'):
            body = compat.multidict_to_dict(request.form)
```

However an OPTIONS requests does not have a content_type by default which causes the `startswith()` method to be called on a NoneType resulting in a crash as this exception is unhandled.

beniwohli:
fixes #172


beniwohli:
fixed in e0b8c5f77569c2b2

PR #181, opened 2018-03-21
preliminary benchmarks for tracing [WIP]

beniwohli:
This is very WIPy. Questions to solve:

 * how can we track this over time?
 * are these benchmarks realistic? At the moment, they are completely dominated by frame collection, is that also the case for real environments?


tests/instrumentation/perf_tests/tests.py
```diff
@@ -0,0 +1,47 @@
...
+    'exclude_paths': ('tests.instrumentation.django_tests', 'elasticapm')
+}], indirect=True)
+def test_stacktrace_performance(elasticapm_client, benchmark):
+    with mock.patch('elasticapm.traces.TransactionsStore.should_collect') as should_collect:
+        def do_it():
+            elasticapm_client.begin_transaction('perf')
+            go()
+            elasticapm_client.end_transaction('perf', 'OK')
+            should_collect.return_value = False
+        benchmark.pedantic(do_it, rounds=10, iterations=100)
```
danielmitterdorfer: Add warmup rounds?

tests/instrumentation/perf_tests/tests.py
```diff
@@ -0,0 +1,47 @@
...
+    'exclude_paths': ('tests.instrumentation.django_tests', 'elasticapm')
+}], indirect=True)
+def test_stacktrace_performance_django(django_elasticapm_client, benchmark):
+    with mock.patch('elasticapm.traces.TransactionsStore.should_collect') as should_collect:
+        def do_it():
+            django_elasticapm_client.begin_transaction('perf')
+            go()
+            django_elasticapm_client.end_transaction('perf', 'OK')
+            should_collect.return_value = False
+        benchmark.pedantic(do_it, rounds=10, iterations=100)
```
danielmitterdorfer: Add warmup rounds?

tests/instrumentation/perf_tests/tests.py
```diff
@@ -0,0 +1,47 @@
...
+
+@pytest.mark.parametrize('elasticapm_client', [{
+    'include_paths': ('tests.instrumentation.perf_tests', 'instrumentation.perf_tests', 'a.b', 'c.d.e', 'f.g.h.x'),
+    'exclude_paths': ('tests.instrumentation.django_tests', 'elasticapm')
+}], indirect=True)
+def test_stacktrace_performance(elasticapm_client, benchmark):
+    with mock.patch('elasticapm.traces.TransactionsStore.should_collect') as should_collect:
+        def do_it():
+            elasticapm_client.begin_transaction('perf')
+            go()
```
danielmitterdorfer: Maybe you should have a small (parameterised) backoff time period here to get a more realistic impression of overhead?

danielmitterdorfer:
> how can we track this over time?

Maybe based on this: http://pytest-benchmark.readthedocs.io/en/stable/comparing.html? You could probably do something similar to what we do for our microbenchmarks for Elasticsearch: We transform the output JSON document and put it into an Elasticsearch instance and visualize the results with Kibana. An alternative would be to upload the graph from pytest to an S3 bucket and serve it from there.

beniwohli:
I added a utility script than runs all the tests in the "benchmark" folder on a range of commits. Is uses a git worktree for that, checking out each commit in the work tree, then copying over the "benchmark" folder from the base directory. Finally, the resulting JSON file from pytest-benchmark is slightly transformed and put into ES.

The nice thing about this is that as we add new benchmarks, we can run them for old commits, giving us some historical perspective. However, it does require some additional work to ensure that new benchmarks work with old code.

The resulting graph looks something like this:



@danielmitterdorfer do you think it would make sense to pipe our benchmark data into the same ES instance the elasticsearch team uses for benchmarking? Or should we run our own thing?

tests/instrumentation/perf_tests/tests.py
beniwohli: I removed the `pedantic` call and added some warmup to the `pytest` command line call

danielmitterdorfer:
That looks pretty neat!

> @danielmitterdorfer do you think it would make sense to pipe our benchmark data into the same ES instance the elasticsearch team uses for benchmarking? Or should we run our own thing?

I think it makes sense to decouple it - among other things for maintenance reasons (it's easier to schedule cluster updates if we don't need to coordinate across multiple teams). Besides, there should be already a separate instance for the APM Java agent microbenchmarks (cc: @felixbarny). I think it would make sense to put it there?  If you expose the results publicly, you could add it to our company-wide benchmarks landing page at benchmarks.elastic.co.

tests/instrumentation/benchmarks/base_perf_tests.py
```diff
@@ -0,0 +1,37 @@
...
+
+capture_span = elasticapm.capture_span if hasattr(elasticapm, 'capture_span') else elasticapm.trace
+
+
+@pytest.mark.parametrize('elasticapm_client', [{
+    'include_paths': ('tests.instrumentation.perf_tests', 'instrumentation.perf_tests', 'a.b', 'c.d.e', 'f.g.h.x'),
+    'exclude_paths': ('tests.instrumentation.django_tests', 'elasticapm')
+}], indirect=True)
+def test_stacktrace_performance(elasticapm_client, benchmark):
+    with mock.patch('elasticapm.traces.TransactionsStore.should_collect') as should_collect:
```
jalvz: question you would never expect from me: why do you mock this?

tests/instrumentation/benchmarks/base_perf_tests.py
```diff
@@ -0,0 +1,37 @@
...
+    for i in range(5):
+        with capture_span('in-go'):
+            go_2()
+            go_someplace_else()
+
+
+@capture_span()
+def go_2():
+    with capture_span('in-go2'):
+        pass
```
jalvz: would be more realistic a time sleep here?

tests/instrumentation/benchmarks/base_perf_tests.py
beniwohli: we're disabling the collection process here, as it might run in the background and suddenly collect transactions, which might influence the actual benchmark. There's other ways to do that, but this one's the easiest.

beniwohli:
Closing this, as the benchmarks no live in https://github.com/elastic/apm-agent-python-benchmarks

PR #183, opened 2018-03-22
Remove Inc from Elastic name in README

watson:
The Elastic company isn't called `Elastic, Inc.` so I took the liberty to clean this up - hope it's ok 😅 If we wanted to call it by the legal name it should be `Elasticsearch BV`, but I don't see the benefit of that in a line like this

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -33,4 +33,4 @@ License
 BSD-3-Clause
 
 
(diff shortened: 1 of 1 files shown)


beniwohli:
Thanks @watson :)

PR #184, opened 2018-03-23
added documentation for which modules/methods we automatically instrument

beniwohli:


alvarolobato:
Should we add the Python versions we support? Also, do we support all the versions of the modules?

beniwohli:
@alvarolobato the question about which versions of which modules we support was one of the reasons that this PR stalled. I just didn't have an answer for that back then. Now that we have matrix tests for all these modules, it's easier to answer. 

Also, yes, supported Python versions and framework versions should also be part of the document

beniwohli:
Finally managed to "finish" (as finished as it can be) documenting this

docs/supported-technologies.asciidoc
```diff
@@ -0,0 +1,324 @@
...
+
+Collected trace data:
+
+    * CQL query
+
+[float]
+[[automatic-instrumentation-http]]
+=== External HTTP requests
+
+[float[
```
bmorelli25:
```suggestion
[float]
```

docs/supported-technologies.asciidoc
```diff
@@ -0,0 +1,324 @@
...
+Additionally, the instrumentation wraps the following methods of the `Elasticsearch` client class:
+
+ * `elasticsearch.client.Elasticsearch.delete_by_query`
+ * `elasticsearch.client.Elasticsearch.search`
+ * `elasticsearch.client.Elasticsearch.count`
+ * `elasticsearch.client.Elasticsearch.update`
+
+Collected trace data:
+
+ * if available, the query string
```
bmorelli25:
```suggestion
 * the query string (if available)
```

docs/supported-technologies.asciidoc
```diff
@@ -0,0 +1,324 @@
...
+ * if available, the `query` element from the request body
```
bmorelli25:
```suggestion
 * the `query` element from the request body (if available)
```

beniwohli:
> 10 months later tada beer

Good things come to those who wait ;p 

> A couple things below... but most importantly:
> 
>     * You'll need to include this file in `index.asciidoc` for it to build `include::supported-technologies.asciidoc[]`

Riight :man_facepalming: 

> Also, a table might be a better way to display all of this information. I'll explore prettifying this later though.

I remember looking into that and having difficulties with the relatively narrow main content column on elastic.co. But I'll trust your talents to find a better solution :)

bmorelli25:
> I remember looking into that and having difficulties with the relatively narrow main content column on elastic.co. But I'll trust your talents to find a better solution :)

Yeah, just played around with it for a bit - the super long method names for elasticsearch and mongo make a table look terrible. Scratch that idea for now.

PR #191, opened 2018-04-06
initial elasticsearch-py instrumentation support, alternative implementation

beniwohli:
This adds instrumentation for an initial set of methods of the
elasticsearch client library, as well as matrix tests for version
2, 5 and 6.

This is an alternative implementation to #169. It moves most of the
work onto a wrapper of `Connection.perform_request`. This has several
benefits:

 * we instrument everything by default, and can do more specific instrumentations
   for other some API methods (e.g. capture the query in the body for `search`)
 * we know which specific ES instance we connect to
 * if connections get retried, they appear as separate spans (more smartness, like
   storing retry information on the span will come later)

There is also one draw back: if we manage to instrument the Elasticsearch object,
but not the pooled Connection object, we could cause errors, because the former
instrumentation adds stuff into the `params` dict that the latter instrumentation
removes again. If the removal doesn't happen, things get messy. I haven't found
a workaround for this yet.

### DIFF
--- a/elasticapm/instrumentation/packages/elasticsearch.py
+++ b/elasticapm/instrumentation/packages/elasticsearch.py
@@ -1,0 +1,144 @@
+from __future__ import absolute_import
+
+import json
+import logging
+
+import elasticapm
+from elasticapm.instrumentation.packages.base import AbstractInstrumentedModule
+
+logger = logging.getLogger(__name__)
+
+
+API_METHOD_KEY_NAME = '__elastic_apm_api_method_name'
+BODY_REF_NAME = '__elastic_apm_body_ref'
+
+
+class ElasticsearchConnectionInstrumentation(AbstractInstrumentedModule):
+    name = 'elasticsearch_connection'
+
+    instrument_list = [
+        ("elasticsearch.connection.http_urllib3", "Urllib3HttpConnection.perform_request"),
+        ("elasticsearch.connection.http_requests", "RequestsHttpConnection.perform_request"),
+    ]
+
+    query_methods = (
+        'Elasticsearch.search',
+        'Elasticsearch.count',
+        'Elasticsearch.delete_by_query'
+    )
+
+    def call(self, module, method, wrapped, instance, args, kwargs):
+        args_len = len(args)
+        http_method = args[0] if args_len else kwargs.get('method')
+        http_path = args[1] if args_len > 1 else kwargs.get('url')
+        params = args[2] if args_len > 2 else kwargs.get('params')
+        body = params.pop(BODY_REF_NAME, None)
+
+        api_method = params.pop(API_METHOD_KEY_NAME, None)
+
+        signature = 'ES %s %s' % (http_method, http_path)
+        context = {'db': {'type': 'elasticsearch'}}
+        if api_method in self.query_methods:
+            query = []
+            # using both q AND body is allowed in some API endpoints / ES versions,
+            # but not in others. We simply capture both if they are there so the
+            # user can see it.
+            if 'q' in params:
+                # 'q' is already encoded to a byte string at this point
+                query.append('q=' + params['q'].decode('utf-8'))
+            if isinstance(body, dict) and 'query' in body:
+                query.append(json.dumps(body['query']))
+            context['db']['statement'] = '\n\n'.join(query)
+        if api_method == 'Elasticsearch.update':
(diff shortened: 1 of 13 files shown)


beniwohli:
One possible solution for the "`Elasticsearch` object is instrumented, but not the `Connection` object" would be

    hasattr(instance.transport.get_connection(), '__wrapped__')

But since the transport has a pool of connections, we'd need to be reasonably sure that if one connection is instrumented, all are, and I'm not sure we can make that assumption.

beniwohli:
Just as an update to the above stated problem: since every `Elasticsearch` instance creates its own `Connection` object, and they aren't shared among `Elasticsearch` instances, this shouldn't be a problem. Either both of them are instrumented, or neither.

elasticapm/instrumentation/packages/elasticsearch.py
```diff
@@ -0,0 +1,145 @@
...
+        signature = 'ES %s %s' % (http_method, http_path)
+        context = {'db': {'type': 'elasticsearch'}}
+        if api_method in self.query_methods:
+            query = []
+            # using both q AND body is allowed in some API endpoints / ES versions,
+            # but not in others. We simply capture both if they are there so the
+            # user can see it.
+            if 'q' in params:
+                # 'q' is already encoded to a byte string at this point
+                query.append('q=' + params['q'].decode('utf-8'))
```
graphaelli: probably very edge case but do query params have to be utf-8 encoded?

elasticapm/instrumentation/packages/elasticsearch.py
```diff
@@ -0,0 +1,145 @@
...
+            if isinstance(body, dict) and 'query' in body:
+                query.append(json.dumps(body['query']))
+            context['db']['statement'] = '\n\n'.join(query)
+        if api_method == 'Elasticsearch.update':
+            if isinstance(body, dict) and 'script' in body:
+                context['db']['statement'] = json.dumps(body)
```
graphaelli: Why doesn't this grab just the `script` here?  Seems care is taken to avoid capturing `doc` in the partial updated, but as is this would capture scripted upsert docs?

elasticapm/instrumentation/packages/elasticsearch.py
```diff
@@ -0,0 +1,145 @@
...
+            # using both q AND body is allowed in some API endpoints / ES versions,
+            # but not in others. We simply capture both if they are there so the
+            # user can see it.
+            if 'q' in params:
+                # 'q' is already encoded to a byte string at this point
+                query.append('q=' + params['q'].decode('utf-8'))
+            if isinstance(body, dict) and 'query' in body:
+                query.append(json.dumps(body['query']))
+            context['db']['statement'] = '\n\n'.join(query)
+        if api_method == 'Elasticsearch.update':
```
graphaelli: `elif` ?

elasticapm/instrumentation/packages/elasticsearch.py
beniwohli: Yes, they are encoded to utf8 here:

https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L34-L38

elasticapm/instrumentation/packages/elasticsearch.py
beniwohli: yes, good catch!

elasticapm/instrumentation/packages/elasticsearch.py
graphaelli: Only if it's not already encoded though - https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/client/utils.py#L29-L31

```
>>> elasticsearch.client.utils._escape(u"\U0001f3d6")
b'\xf0\x9f\x8f\x96'

>>> elasticsearch.client.utils._escape(u"\U0001f3d6".encode('utf-16'))
b'\xff\xfe<\xd8\xd6\xdf'


>>> elasticsearch.client.utils._escape(u"\U0001f3d6".encode('utf-16')).decode('utf-8')
UnicodeDecodeError
```

Not sure if this is worth a lot of effort to handle.

beniwohli: ugh, right. I guess we at least can catch the `UnicodeDecodeError` if it happens

PR #200, opened 2018-04-18
added support for transaction marks, and added option to mark errors [WIP]

beniwohli:


tests/client/test_transaction_marks.py
```diff
@@ -0,0 +1,70 @@
...
+    assert "No transaction currently active" in caplog.records[0].message
+
+
+@pytest.mark.parametrize('elasticapm_client', [
+    {'transaction_mark_errors': 'all'},
+    {'transaction_mark_errors': 'handled'},
+    {'transaction_mark_errors': 'unhandled'},
+    {'transaction_mark_errors': 'off'},
+], indirect=True)
+def test_mark_errors(elasticapm_client):
```
felixbarny: I think in this case you might be better off creating separate test methods instead of using if statements inside the method. You could extract a method for the begin/end transaction to get rid of the repetition.

elasticapm/traces.py
```diff
@@ -316,3 +320,20 @@ def set_user_context(username=None, email=None, user_id=None):
     if user_id is not None:
         data['id'] = encoding.keyword_field(user_id)
     set_context(data, 'user')
+
+
+def mark(group_name, name, override=True):
```
felixbarny: why not make `mark` an instance method of `transaction`?

elasticapm/base.py
```diff
@@ -178,6 +178,12 @@ def capture(self, event_type, date=None, context=None, custom=None, stack=None,
         if data:
             url = self.config.server_url + constants.ERROR_API_PATH
             self.send(url, **data)
+            c = self.config.transaction_mark_errors
```
felixbarny: Maybe it would be easier to just overlay the errors in the UI using the errors index, instead of "duplicating" the error events in marks? WDYT?

elasticapm/traces.py
beniwohli: Generally, the agent user isn't interacting directly with the transaction anywhere, and instead uses these module-level functions (which can also be accessed from the "root" module, so e.g. `elasticapm.mark()`). There's two reasons for this API design:

 * it makes for a nicer API in my opinion. Compare

       elasticapm.mark('my-mark')

   with

       transaction = get_transaction()
       transaction.mark('my-mark')
 * `get_transaction()` is not a public API (as in not documented, there's not really anything we can do to stop people from using it), because at the moment it is a very thin wrapper around a thread local. But once we start supporting async event loops, thread locals will not work anymore, so we will need other ways to get the current transaction. I currently don't know how exactly that will look, so this API gives me some future wiggle room.

elasticapm/base.py
beniwohli: Probably (although it would need another ES query to get that data). However, this is just an experiment, hence the `[WIP]` marker, and mostly was intended as a source of example data for the frontend team. I have no intention of merging it anytime soon :)

beniwohli:
Closing this for now until we come up with a proper use case

PR #211, opened 2018-05-03
updated pytest and dependencies to newest versions

beniwohli:
with a bit of luck, this will resolve occasional test failures
due to a threading lock issue in pytest-catchlog
(which has been integrated into pytest proper)


beniwohli:
Turns out pytest 3.3 isn't working on Python 3.3...

PR #215, opened 2018-05-08
test: remove deleted APM Server JSON schema files

watson:
The removed files are:
- `meta.json`
- `schema.json`

I also changed tabs to spaces as you used a combination of both in the same file. So now it's aligned. I hope that was ok?


beniwohli:
test failure is unrelated, unfortunately some times our http server fails on appveyor :(

PR #220, opened 2018-06-04
docs: use shared attributes

axw:
Use the shared attributes provided
by elastic/docs for linking to other
product docs, forums, etc.

### DIFF
--- a/docs/index.asciidoc
+++ b/docs/index.asciidoc
@@ -1,5 +1,8 @@
 = APM Python Agent Reference
 
+:branch: current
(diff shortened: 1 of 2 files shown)


docs/index.asciidoc
```diff
@@ -1,5 +1,8 @@
 = APM Python Agent Reference
 
+:branch: current
+include::{asciidoc-dir}/../../shared/attributes.asciidoc[]
```
beniwohli: Excuse my ignorance, but where does `asciidoc-dir` come from? Is it like a global variable?

PR #223, opened 2018-06-07
added CONTRIBUTING.md

beniwohli:
Thanks @watson, @Qard  and https://github.com/elastic/apm-agent-nodejs for large parts of the
document.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,94 @@
+# Contributing to the Python APM Agent
+
+The APM Agent is open source and we love to receive contributions from our community — you!
+
+There are many ways to contribute,
+from writing tutorials or blog posts,
+improving the documentation,
+submitting bug reports and feature requests or writing code.
+
+You can get in touch with us through [Discuss](https://discuss.elastic.co/c/apm),
+feedback and ideas are always welcome.
+
+Please note that this repository is covered by the [Elastic Community Code of Conduct](https://www.elastic.co/community/codeofconduct).
+
+## Code contributions
+
+If you have a bugfix or new feature that you would like to contribute,
+please find or open an issue about it first.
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,94 @@
...
+    We might not be able to review your code as fast as we would like to,
+    but we'll do our best to dedicate it the attention it deserves.
+    Your effort is much appreciated!
+
+### Testing
+
+To run the full test suite,
+you need to install several databases (Elasticsearch, PostgreSQL, MySQL, Cassandra, Redis).
+This can be quite a hassle, so we recommend to use our dockerized test setup.
+See [Running tests](https://www.elastic.co/guide/en/apm/agent/python/current/run-tests-locally.html) for detailed instructions.
```
watson: Consider if you want to use `master` instead of `current` as people mostly will make PR's against the master branch

beniwohli: good point!

Qard: You can change the default target branch for PRs. For example, [npm](https://github.com/npm/npm) uses `latest` rather than `master` and actually deleted the `master` branch to avoid the potentially problematic terminology.

beniwohli: @Qard I don't think a PR in this tiny little repo is the right venue for this discussion. If you think the usage of `master` in the context of git branches is problematic, I suggest to bring it up in a company/community wide forum.

Qard: Oh, I didn't intend to put attention on that part. I was just trying to point out you can *change* the default target branch for PRs easily, regardless of whatever reason one might have to do that. I just provided that as an example. It was intended as a counter to Thomas stating that most people will PR against master. 😅

Issue #227, opened 2018-06-11
IndexError: pop from empty list


gusibi:
elasticapm/traces.py

```python 
def end_span(self, skip_frames):
        span = self.span_stack.pop()

```
span_stack may be empty

IndexError: pop from empty list

beniwohli:
Thanks @gusibi! Usually, when this happens, it is caused by another bug. Do you have the full stack trace of this error?

gusibi:
![2018-06-11 18 51 13](https://user-images.githubusercontent.com/1282277/41227864-eb2cb9ce-6da8-11e8-8a76-6f7c93b5b917.png)
![2018-06-11 18 51 29](https://user-images.githubusercontent.com/1282277/41227866-eb682298-6da8-11e8-922c-824fdc179feb.png)
![2018-06-11 18 51 41](https://user-images.githubusercontent.com/1282277/41227867-eb9e9fc6-6da8-11e8-9e0f-932438844483.png)

gusibi:
redis server is pika

beniwohli:
Thanks for the update. It's hard to say what exactly can cause this. The only place where we call `begin_span` is in `capture_span.__enter__`, and the only place where we call `end_span` is in `capture_span.__exit__`. According to the context manager protocol, if we `__enter__`, we will also always `__exit__`.

So the only explanation that I have is that the `transaction` object in `__enter__` is different than the one in `__exit__`.

@gusibi do you per chance use an asynchronous framework (asyncio, twisted, ...)? We currently don't support asynchronous execution (see #16), and I could imagine that it could cause an error like this.

gusibi:
@beniwohli Thanks
I use the flask, not used asynchronous framework.

This error all in the process of using redis.
And not every time this error occurs.

PR #228, opened 2018-06-11
#227 fix index error

gusibi:


beniwohli:
@gusibi I went with a slightly different approach to fix this, see #229. By catching the error in the `capture_span` context manager, we can log some details about which span caused the error. I included you as co-author in that commit.

PR #242, opened 2018-06-26
drop support for Python 3.3 by removing it from the test matrix

beniwohli:


beniwohli:
test failures are unrelated, due to MS SQL Server docker getting stuck

PR #266, opened 2018-07-31
Fix wrong user_id parameter for set_user_context function in docs

lyrixderaven:


beniwohli:
Thanks for the PR @lyrixderaven, great catch! And sorry for the delay, I was on vacation 😊

lyrixderaven:
No worries, vacation's important :-) It was just a small issue in the docs, after all.

PR #282, opened 2018-09-06
refactored validators for config to allow the validator to change the value

### LINKED ISSUE #280: Implement duration config suffixes
For duration configs (e.g. ELASTIC_APM_API_REQUEST_TIME), we will support time units like 1ms, 1s, 1m. For pre-v2 config, we should allow suffixes, but also continue to allow no suffix and assume the previous unit (e.g. "ms" for min span duration). For v2+ config, we should require suffixes.

beniwohli:
this allows implementation of e.g. unit parsers in the validator. A
UnitValidator that makes use of this is also included

closes #280

### DIFF
--- a/tests/config/tests.py
+++ b/tests/config/tests.py
@@ -4,7 +4,17 @@ import logging
 
 import mock
 
-from elasticapm.conf import Config, _BoolConfigValue, _ConfigBase, _ConfigValue, _ListConfigValue, setup_logging
+from elasticapm.conf import (
+    Config,
+    RegexValidator,
+    _BoolConfigValue,
+    _ConfigBase,
+    _ConfigValue,
+    _ListConfigValue,
+    duration_validator,
+    setup_logging,
+    size_validator,
+)
 
 
 def test_basic_not_configured():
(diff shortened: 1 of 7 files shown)


tests/config/tests.py
```diff
@@ -131,3 +141,42 @@ class MyConfig(_ConfigBase):
...
+
+    c = MyConfig({"BYTE": "10b", "KBYTE": "5kib", "MBYTE": "17mib", "WRONG_PATTERN": "5 kb"})
+    assert c.byte == 10
+    assert c.kbyte == 5 * 1024
+    assert c.mbyte == 17 * 1024 * 1024
+    assert c.wrong_pattern is None
+    assert "WRONG_PATTERN" in c.errors
+
+
+def test_duration_validation():
```
graphaelli: Need a test for multiple validators?

beniwohli: good idea, added

beniwohli:
Thanks for the review @graphaelli, great input as always!

I didn't really realize until now that renaming `span_frames_min_duration_ms` is backwards incompatible. That kinda sucks :/ We'll have to do a major version bump anyway for v2 and distributed tracing, but I hoped I could merge this before. Guess that's not going to happen then. I'll merge it into the v2-api branch.

beniwohli:
Merged into #281, closing this one

PR #252, opened 2018-07-09
implement support for asyncio [double-plus-WIP]

beniwohli:
DO NOT TRY THIS AT HOME JUST YET

Note: in the v2 API implementation PR, we [removed](https://github.com/elastic/apm-agent-python/pull/281/commits/c2b8d7f1a886) the `AsyncIOHTTPTransport`. It might make sense to re-add it with this PR.

PR #291, opened 2018-09-20
refactored span stack and added contextvars as alternative for threadlocals

beniwohli:
this is a partial "import" of the asyncio work done in #252, sans the
asyncio specific extensions


beniwohli:
Merged into 4.x branch

PR #299, opened 2018-10-15
remove any "store" remnants from TransactionStore and rename it to Tr…

beniwohli:
…acer


elasticapm/traces.py
```diff
@@ -112,53 +108,54 @@ def to_dict(self):
...
         :param context: context dictionary
-        :param leaf: is this transaction a leaf transaction?
+        :param leaf: is this span a leaf span?
         """
-        self.idx = idx
-        self.transaction_id = transaction_id
-        self.trace_id = trace_id
+        self.start_time = _time_func()
+        self.id = "%016x" % random.getrandbits(64)
+        self.transaction = transaction
```
graphaelli: any perf worries about referencing the transaction rather than pulling the bits needed out from it here in init?

elasticapm/traces.py
```diff
@@ -112,53 +108,54 @@ def to_dict(self):
...
         self.name = name
         self.type = span_type
         self.context = context
         self.leaf = leaf
-        self.start_time = start_time
+        # timestamp is bit of a mix of monotonic and non-monotonic time sources.
+        # we take the (non-monotonic) transaction timestamp, and add the (monotonic) difference of span
+        # start time and transaction start time. In this respect, the span timestamp is guaranteed to grow
```
graphaelli: love this comment

elasticapm/traces.py
beniwohli: other than the transaction not being able to be GC'd before the last span is GC'd (which is anyway the normal course of execution), not really

In fact, it should use a bit less memory, less attributes per span. But that's probably not of any measurable relevance.

PR #309, opened 2018-10-29
fix traceparent header propagation for older versions of requests

beniwohli:
until fairly recently, requests used to vendor urllib3. To make
header propagation work, we need to instrument than vendored urllib3 as
well.


beniwohli:
jenkins run full matrix tests

beniwohli:
jenkins run the full test matrix

PR #310, opened 2018-10-29
added note about lack of support for environment in APM UI

### LINKED ISSUE #284: Add warning about use of environment config option
The `environment` config option isn't supported by the APM UI yet. The UI will mix data from multiple environments together if used without also differentiating the service name. This will be confusing for the user.

A warning should be added to the documentation instructing the user to also add the environment as a suffix to the service name.

beniwohli:
fixes #284

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -134,6 +134,11 @@ Note however that this can have adverse effects on performance.
 The name of the environment this service is deployed in,
 e.g. "production" or "staging".
 
+NOTE: `environment` is currently not supported in the Kibana APM UI.
+This means that data from different environments that share the same
+`<<config-service-name, service name>>` will be shown together in the APM UI.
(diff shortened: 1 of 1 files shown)


docs/configuration.asciidoc
```diff
@@ -134,6 +134,11 @@ Note however that this can have adverse effects on performance.
 The name of the environment this service is deployed in,
 e.g. "production" or "staging".
 
+NOTE: `environment` is currently not supported in the Kibana APM UI.
+This means that data from different environments that share the same
+`<<config-service-name, service name>>` will be shown together in the APM UI.
```
watson:
```suggestion
`<<config-service-name,service name>>` will be shown together in the APM UI.
```

watson: Not sure if this actually makes a difference in the rendered HTML, but I think it's a better syle.

PR #312, opened 2018-11-01
Update LoggingHandler to stringify record.msg

### LINKED ISSUE #295: HTTP 400: data validation error: Problem validating JSON document against schema
Sometimes when sending an event to the APM instance a HTTP 400 response is returned:

    HTTP 400: data validation error: Problem validating JSON document against schema: 
    I[#] S[#] doesn't validate with "error#" I[#/errors/0/log/message] 
    S[#/properties/errors/items/properties/log/properties/message/type] 
    expected string, but got array

I've also seen a similar, but subtly different message:

    HTTP 400: data validation error: Problem validating JSON document against schema: 
    I[#] S[#] doesn't validate with "error#" I[#/errors/0/log/param_message] 
    S[#/properties/errors/items/properties/log/properties/param_message/type] 
    expected string or null, but got array

This is then followed by a log message:

    Failed to submit message: '<no message value>'

I'm using the latest elastic-apm client available from PyPI (3.0.1) with Python 3.6.5 and the APM server is using version 6.4.1.

(This seems similar to #135, but not quite the same).

axw:
Update LoggingHandler so that it converts the log record
message to a string. The message may be any arbitrary object,
and currently non-string messages are being sent as objects,
which the APM Server rejects.

Fixes #295

### DIFF
--- a/elasticapm/handlers/logbook.py
+++ b/elasticapm/handlers/logbook.py
@@ -16,6 +16,7 @@ import traceback
 import logbook
 
 from elasticapm.base import Client
+from elasticapm.utils import compat
 from elasticapm.utils.encoding import to_unicode
 
 LOOKBOOK_LEVELS = {
@@ -80,7 +81,7 @@ class LogbookHandler(logbook.Handler):
             exception = None
 
         return self.client.capture_message(
-            param_message={"message": record.msg, "params": record.args},
+            param_message={"message": compat.text_type(record.msg), "params": record.args},
             exception=exception,
             level=LOOKBOOK_LEVELS[record.level],
             logger_name=record.channel,
--- a/elasticapm/handlers/logging.py
+++ b/elasticapm/handlers/logging.py
@@ -130,7 +130,7 @@ class LoggingHandler(logging.Handler):
 
         return self.client.capture(
             "Message",
-            param_message={"message": record.msg, "params": record.args},
+            param_message={"message": compat.text_type(record.msg), "params": record.args},
             stack=stack,
             custom=custom,
             exception=exception,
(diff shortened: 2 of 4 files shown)


axw:
> A quick note on branches. Master has a bunch of backwards incompatible changes, and will be released as 4.0 together with APM Server 6.5. We will continue to cut maintenance releases of the 3.x line for some time. I can create a new 3.x with a backport of this PR some time next week (I'm traveling at the moment).

I've targeted 3.x (targeting master was a mistake), so that @jaap3 can test.

There are failures in apm-ci, but they appear to have something to do with json-schema; nothing I touched.

jaap3:
@axw I've installed your branch, I'll keep it running for a while and report back if I see anything (earliest monday)

jaap3:
@beniwohli this is probably not the best place to discuss this, but, after the release of APM Server 6.5/Client 4.0 what would an upgrade path look like?

e.g. should all projects be upgraded to client version 4.0 immediately? Or can the server 6.5 handle messages from Client < 4.0? Or, inverted, can projects be safely upgraded to client 4.0 while the server is still < 6.5?

This is not a huge issue for me currently, but in the future I will probably have many project reporting to the same APM server. So I'm very interested in a smooth upgrade path where I don't have to worry about updating multiple things simultaneously.

From reading the CHANGELOG it seems that I should at least be pinning the client to <4 for now.

beniwohli:
@jaap3 regarding upgrading: APM Server 6.5 (and all coming 6.x versions) will have support for the v1 API, which the agent versions < 4.0 talk. So you should be able to upgrade your APM Server(s) first, and then upgrade the agents. 

The other way around is unfortunately not possible, as it would have been way too complicated to support both v1 and the new v2 API in the agent.

Generally speaking, we highly suggest to pin the elastic-apm dependency to at least the major version. New major versions usually indicate a backwards incompatible change. We should probably add a note to the install docs about this...

axw:
> @axw force-pushed a rebase on top of updated 3.x branch to your repo, hope that's ok blush will merge as soon as tests pass (travis ci will not pass due to unrelated breakage with django master)

Sure is. Thanks for taking this over.

beniwohli:
Merged into 3.x and cherry-picked into master. Thanks for the work on this @axw, and thanks @jaap3 for the report!

PR #313, opened 2018-11-06
new version of schema download script

beniwohli:
the first version was brittle because it explicitly listed all files
to be downloaded, which broke every time APM Server changed the naming
or layout of the files.

The second version used the Github API to get a directory listing, then
downloaded the files individually. That worked, except for the fact that
unauthenticated API requests are severily rate limited per IP, and more
often than not the rate limit of the Travis machine was already reached,
resulting in a failed build.

This version downloads the tarball, and extracts only the docs/spec
subdirectory. Timings indicate that this is actually the fastest
solution, even though it downloads much more data in total. This is
probably due to the fact that only one HTTP request is made, and heavy
caching of tarballs on Github's side.


tests/scripts/download_json_schema.sh
```diff
@@ -0,0 +1,28 @@
...
+        curl --silent --fail https://codeload.github.com/elastic/apm-server/tar.gz/${2} | tar xzvf - --wildcards --directory=${1} --strip-components=1 "*/docs/spec/*"
+        result=$?
+        if [ $result -eq 0 ]; then break; fi
+        sleep 1
+    done
+
+    mv -f ${1}/docs/spec/* ${1}/
+    rm -rf ${1}/docs
+
+    if [ $result -ne 0 ]; then exit $result; fi
```
graphaelli: I'd move this up to before the `mv` and then `set -e`, just in case.

beniwohli: good point! Somehow I forgot that I requested a review from you and already merged 🙈 I'll fix it in a separate commit

PR #319, opened 2018-11-07
Add instrumentation for SUDS (Soap Client library)

Ismael:
Adds support for SUDS: https://pypi.org/project/suds/
It captures each SOAP call as a span named suds.service.[ServiceName]


basepi:
I'm a bit embarrassed that we have left this for so long.

I'm going to go ahead a close this for now, as it appears it hasn't had a lot of input/traffic. If you're still interested in making this happen @Ismael, can you please open a new PR with this code and add some tests? Sorry for our lack of response here for so long.

PR #322, opened 2018-11-13
added context processors and docs for RUM correlation

### LINKED ISSUE #318: [DT] Implement possibility to pass trace/parent ID to RUM agent
The parent ID, if not set, should be generated through an `transaction.ensure_parent()` call, which generates a parent ID and sets it as the parent of the transaction

In Django, this is probably best solved with a context processor that adds all variables to the context. A slight complication here is that `ensure_parent()` should only be called if the variable is actually rendered. It should be possible to just pass a callable instead of the actual string into the context.

I'm not quite sure yet what the best solution for Flask is

beniwohli:
fixes #318

### DIFF
--- a/docs/django.asciidoc
+++ b/docs/django.asciidoc
@@ -115,6 +115,48 @@ ELASTIC_APM['TRANSACTIONS_IGNORE_PATTERNS'] = ['^OPTIONS ', 'views.api.v2']
 This example ignores any requests using the `OPTIONS` method and any requests containing `views.api.v2`.
 
 [float]
+[[django-integrating-with-the-rum-agent]]
+==== Integrating with the RUM agent
+
+To correlate performance measurement in the browser with measurements in your Django app,
+you can help the RUM (Real User Monitoring) agent by configuring it with the Trace ID and Span ID of the backend request.
+We provide a handy template context processor which adds all the necessary bits into the context of your templates.
+
+To enable this feature, first add the `rum_tracing` context processor to your `TEMPLATES` setting.
+You most likely already have a list of `context_processors`, in which case you can simply append ours to the list.
+
+[source,python]
+----
+TEMPLATES = [
+    {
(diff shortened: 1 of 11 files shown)


elasticapm/contrib/flask/__init__.py
```diff
@@ -131,6 +132,23 @@ def init_app(self, app, **defaults):
...
+        def rum_tracing():
+            """
+            Adds APM related IDs to the context used for correlating the backend transaction with the RUM transaction
+            """
+            transaction = get_transaction()
+            if transaction.trace_parent:
+                return {
+                    "apm": {
+                        "trace_id": transaction.trace_parent.trace_id,
+                        "span_id": lambda: transaction.ensure_parent_id(),
```
graphaelli: could this be `transaction.ensure_parent_id` instead of a lambda?

docs/django.asciidoc
```diff
@@ -114,6 +114,48 @@ ELASTIC_APM['TRANSACTIONS_IGNORE_PATTERNS'] = ['^OPTIONS ', 'views.api.v2']
...
+----
+
+Then, update the call to initialize the RUM agent (which probably happens in your base template) like this:
+
+[source,javascript]
+----
+elasticApm.init({
+    serviceName: "my-frontend-service",
+    pageLoadTraceId: "{{ apm.trace_id }}",
+    pageLoadSpanId: "{{ apm.span_id }}",
```
graphaelli: is `apm.span_id` automatically called or should this have `()` in it?

beniwohli: The Django Template Language calls callables automatically 👍

elasticapm/contrib/flask/__init__.py
beniwohli: Yes! Good catch! I'll also add a comment to avoid that me + 6 months thinks present me was an idiot and forgot `()`

beniwohli:
jenkins test this please

PR #335, opened 2018-11-15
fix threafing bug with send buffer and threaded workers

### LINKED ISSUE #334: [4.0.0] UnicodeDecodeError: 'utf-8' codec can't decode byte 0x96 in position 91: invalid start byte
Django 1.11.16
APM agent 4.0.0
APM server 6.5

Exception is raised at https://github.com/elastic/apm-agent-python/blob/f600230a1643a9442605fce959946790c85c4c61/elasticapm/transport/http.py#L62

For my case `body` is the bytes string below:

    b'data read error: invalid character \'\\x00\' looking for beginning of value [\x00\x00P\x02\x00\x00\.....

beniwohli:
When using threaded workers (e.g. Django runserver in Django 2.0+),
writes to the send buffer need to be guarded by a threading lock.
Otherwise we end up with garbled content in the buffer, which then later
on leads to parsing errors in the APM Server.

fixes #334

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -93,9 +93,11 @@ class Transport(object):
 
     @property
     def queued_data_size(self):
-        f = self.queued_data
-        # return size of the underlying BytesIO object if it is compressed
-        return f.fileobj.tell() if hasattr(f, "fileobj") else f.tell()
+        f = self._queued_data
+        if f:
+            # return size of the underlying BytesIO object if it is compressed
+            return f.fileobj.tell() if hasattr(f, "fileobj") else f.tell()
+        return 0
 
     def flush(self, sync=False, start_flush_timer=True):
         """
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -53,12 +53,13 @@ class Transport(object):
         self._max_flush_time = max_flush_time
         self._max_buffer_size = max_buffer_size
         self._queued_data = None
-        self._flush_lock = threading.Lock()
+        self._queue_lock = threading.Lock()
         self._last_flush = timeit.default_timer()
         self._flush_timer = None
 
     def queue(self, event_type, data, flush=False):
-        self._queue(self.queued_data, {event_type: data})
+        with self._queue_lock:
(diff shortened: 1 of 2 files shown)


elasticapm/transport/base.py
```diff
@@ -53,12 +53,13 @@ def __init__(
...
+        self._queue_lock = threading.Lock()
         self._last_flush = timeit.default_timer()
         self._flush_timer = None
 
     def queue(self, event_type, data, flush=False):
-        self._queue(self.queued_data, {event_type: data})
+        with self._queue_lock:
+            self.queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
         since_last_flush = timeit.default_timer() - self._last_flush
         queue_size = self.queued_data_size
```
roncohen: `self.queued_data_size` calls `self.queued_data` so calls to `self.queued_data_size` must also be protected by the lock, or you lock inside `self.queued_data`

beniwohli: Great catch. I'll remove the call to `self.queued_data` in `queued_data_size` and instead check for existence of `self._queued_data` directly and return `0` if it doesn't exist. Then we don't need a lock as it is read only. Makes sense?

simitt: You could do something like:
```
with self._queue_lock:
            self.queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
            queue_size = self.queued_size

```

beniwohli: @roncohen @simitt like this af2ffd490bc

roncohen: yeah that would probably work. It seems like `queue_size` here could end up being the size of a queue that was just flushed and we would then flush again, but that might not be an issue

beniwohli:
@ciarancourtney oh, you're right. Weirdly enough, I can't replicate the issue in Django 1.11 with runserver

elasticapm/transport/base.py
simitt: agree, this shouldn't really matter

beniwohli:
merged in 4a02ee52a8f8f0ae63f97c9abc00dd1f2a9b527e

PR #344, opened 2018-11-23
added option to limit logging by level in Flask

### LINKED ISSUE #342: [Flask] logging=True can be extremely noisy, needs to be documented
Using `ElasticAPM(logging=True)` sets up our `LoggingHandler` to log *everything* to ElasticAPM. This is almost always not what the user wants, and needs to be documented as possibly very noisy.

### LINKED ISSUE #343: [Flask] Add option to set up logging by log level
Currently, instantiating the flask client with `logging=True` will log everything to Elastic APM. It would be nice to also support `logging=SOME_LOG_LEVEL`, which would only send messages with level `SOME_LOG_LEVEL` and higher to Elastic APM.

beniwohli:
fixes elastic/apm-agent-python#342
fixes elastic/apm-agent-python#343

### DIFF
--- a/docs/flask.asciidoc
+++ b/docs/flask.asciidoc
@@ -143,6 +143,18 @@ from elasticapm.contrib.flask import ElasticAPM
 apm = ElasticAPM(app, logging=True)
 ----
 
+You can also limit logging to a certain level:
+
+[source,python]
+----
+import logging
+
+apm = ElasticAPM(app, logging=logging.ERROR)
+----
+
+NOTE: using `logging=True` can lead to high number of logs being sent to Elastic APM.
+We recommend to always limit logging with Elastic APM to a high level (`WARNING` or `ERROR`).
+
 For fine-grained control, you can initialize a logging handler and add it,
 just as you would with any other handler.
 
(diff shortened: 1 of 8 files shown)


docs/flask.asciidoc
```diff
@@ -143,6 +143,18 @@ from elasticapm.contrib.flask import ElasticAPM
 apm = ElasticAPM(app, logging=True)
```
roncohen: `logging=True` seems like a footgun. I'd rather lead with the explanation of `logging=logging.ERROR` and then mention that you can do `logging=True` to get everything incl. the nice warning that you currently have below the `logging=logging.ERROR` example.

It's also worth considering to completely stop mentioning `logging=True`. Setting `logging=logging.DEBUG` has the same effect and is more obviously a footgun.

beniwohli: That makes thev docs ten times better! 👍

beniwohli:
Looks like adding `logging=xyz` has some side effects, not sure which. I hoped that removing the installed handler in the fixture tear down would be enough, but that doesn't seem to be the case :(

beniwohli:
merged in e05afc43987

PR #351, opened 2018-12-04
move queue size logic into a lock-protected block

### LINKED ISSUE #348: AttributeError: 'NoneType' object has no attribute 'tell'
I'm using 4.0.1 version.
Exceptions are raised from `elasticapm.transport.base.queued_data_size`

Here is last trace.

```
        if f:
            # return size of the underlying BytesIO object if it is compressed
            return f.fileobj.tell() if hasattr(f, "fileobj") else f.tell()
        return 0
```

screen shot 2018-11-28 at 12 23 39 pm
screen shot 2018-11-28 at 12 24 41 pm

beniwohli:
This is a shot in the dark to fix #348 as I wasn't able to reproduce.

The idea is to move the size check into a code block that is protected by the thread lock, and also to make sure to check that `fileobj` exists.

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -59,9 +59,10 @@ class Transport(object):
 
     def queue(self, event_type, data, flush=False):
         with self._queue_lock:
-            self.queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
-        since_last_flush = timeit.default_timer() - self._last_flush
-        queue_size = self.queued_data_size
+            queued_data = self.queued_data
+            queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
+            since_last_flush = timeit.default_timer() - self._last_flush
+            queue_size = 0 if queued_data is None or queued_data.fileobj is None else queued_data.fileobj.tell()
         if flush:
             logger.debug("forced flush")
             self.flush()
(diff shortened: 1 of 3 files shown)


elasticapm/transport/base.py
```diff
@@ -59,9 +59,10 @@ def __init__(
 
     def queue(self, event_type, data, flush=False):
         with self._queue_lock:
-            self.queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
-        since_last_flush = timeit.default_timer() - self._last_flush
-        queue_size = self.queued_data_size
+            queued_data = self.queued_data
+            queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
+            since_last_flush = timeit.default_timer() - self._last_flush
+            queue_size = 0 if queued_data is None or queued_data.fileobj is None else queued_data.fileobj.tell()
```
axw: shouldn't you still be checking if queued_data _has_ a fileobj? i.e. for the case where compression is disabled, and queued_data is a BytesIO?

beniwohli: ugh, right. Having the option to turn off the use of gzip compression was probably a mistake in hindsight. Especially because using `compression_level=0` already turns off any compression :/ this might be as good a time as any to remove it. I don't think it would violate backwards compatibility

axw: Seems reasonable. Maybe just treat None the same as 0 for compress_level: create a GzipFile with compresslevel=0?

beniwohli: done in a479fc7 👍

beniwohli:
closed in 4425fad9e7f386d

Issue #347, opened 2018-11-27
Add kubernetes metadata to system data


beniwohli:
See https://github.com/elastic/apm/issues/21 for details

implemented in elastic/apm-agent-python#352

Issue #346, opened 2018-11-27
Add docker container metadata to system data


beniwohli:
See https://github.com/elastic/apm/issues/22 for details

implemented in elastic/apm-agent-python#352

PR #352, opened 2018-12-05
added parsing of cgroups file

beniwohli:
due to the somewhat underdefined and volatile format of this file,
the "parsing" got pretty spaghetti-code-y :(

This implements elastic/apm-agent-python#346 and elastic/apm-agent-python#347


tests/client/client_tests.py
```diff
@@ -39,15 +39,82 @@ def test_process_info(elasticapm_client):
...
+        system_info = elasticapm_client.get_system_info()
+    assert "kubernetes" in system_info
+    assert system_info["kubernetes"] == {
+        "pod": {"uid": "podid", "name": "pod"},
+        "node": {"name": "node"},
+        "namespace": "namespace",
+    }
+    assert system_info["container"] == {"id": "123"}
+
+
```
graphaelli: Worth a test to ensure system hostname is used as pod name when kubernetes info is not detected but partially set by environ?  Seems like overkill but figured I'd mention it to prove I reviewed this.

```python
@mock.patch.dict(
    "os.environ",
    {
        "KUBERNETES_NAMESPACE": "namespace",
    },
)
def test_docker_kubernetes_system_info_except_hostname_from_environ():
    # initialize agent only after overriding environment
    elasticapm_client = Client()
    # mock docker/kubernetes data here to get consistent behavior if test is run in docker
    with mock.patch("elasticapm.utils.cgroup.get_cgroup_container_metadata") as mock_metadata, mock.patch(
        "socket.gethostname"
    ) as mock_gethostname:
        mock_metadata.return_value = {}
        mock_gethostname.return_value = "foo"
        system_info = elasticapm_client.get_system_info()
    assert "kubernetes" in system_info
    assert system_info["kubernetes"] == {
        "pod": {"name": "foo"},
        "namespace": "namespace",
    }
```

beniwohli: this would have been the ideal chance to use that new fancy review-as-commit thingy :D set you as author of the commit instead. Not that it makes a huge difference after I squash-merge this :D

PR #353, opened 2018-12-06
implement dedotting of context keys and tag names

beniwohli:
(and destaring and dedoublequoting)

These characters are invalid in elasticsearch field names.


beniwohli:
looks like the two failed builds are due to network failures

PR #361, opened 2018-12-14
system metrics (CPU/memory)

beniwohli:


elasticapm/metrics/base_metrics.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(self, label):
+        self.label = label
+        self._val = None
+
+    @property
+    def val(self):
+        return self._val
+
+    @val.setter
+    def val(self, value):
```
kuisathaverat: Should it be thread safe?

beniwohli: I don't think adding a lock adds anything here regarding thread safety. The new state doesn't depend on the previous state. If two threads call it concurrently, one of it wins. Which one wins isn't really important.

Or am I not thinking this right?

kuisathaverat: I asked because the counter uses a lock

beniwohli: yep, there the new state depends on the old state, so two threads incrementing/decrementing the counter concurrently could step on each others' feet without a lock I think

Ah, threading... should have worn my new T-Shirt today, with this on it:

elasticapm/metrics/cpu_linux.py
```diff
@@ -0,0 +1,81 @@
...
+        self.sys_stats_file = sys_stats_file
+        self.process_stats_file = process_stats_file
+        self.memory_stats_file = memory_stats_file
+        with self._read_data_lock:
+            self.previous.update(self.read_process_stats())
+            self.previous.update(self.read_system_stats())
+        super(CPUMetricSet, self).__init__()
+
+    def collect(self):
+        cpu_count = float(multiprocessing.cpu_count())
```
axw: You don't need to divide by `cpu_count`. You're reporting the fraction of non-idle total CPU time, which is going to be in the range 0-1 already.

elasticapm/metrics/cpu_linux.py
```diff
@@ -0,0 +1,81 @@
...
+        new = self.read_process_stats()
+        new.update(self.read_system_stats())
+        with self._read_data_lock:
+            prev = self.previous
+            delta = {k: new[k] - prev[k] for k in new.keys()}
+            cpu_usage_ratio = (delta["cpu_total"] - delta["cpu_usage"]) / delta["cpu_total"]
```
axw: `cpu_usage` is non-idle time, so this expression calculates the percentage of *idle* time. This should work:

```python
cpu_usage_ratio = delta["cpu_usage"] / delta["cpu_total"]
self.gauge("system.cpu.total.norm.pct").val = cpu_usage_ratio
```

elasticapm/metrics/cpu_linux.py
beniwohli: fixed in 3048596

elasticapm/metrics/cpu_linux.py
beniwohli: fixed in 3048596

elasticapm/metrics/sets/cpu.py
```diff
@@ -0,0 +1,7 @@
+import os
+import platform
+
+if platform.system() == "Linux" and "ELASTIC_APM_FORCE_PSUTIL_METRICS" not in os.environ:
```
graphaelli: why not always use `psutil` if available, even on linux?

beniwohli: as it is right now, psutil returns somewhat weird numbers (e.g. process CPU utilization is always 0.0). We might even have to yank psutils out completely :(

beniwohli: with 7ab887d, psutil should now also be operational. I still think we should default to our own `/proc/` parsing on Linux, as psutil is pretty humongous

graphaelli: I can live with that, though I do like that psutil handles older formats of `/proc/stat`, etc.  Probably premature but it would be neat if we could use psutil if available if we're sure that the homegrown routines won't work (like on older kernels)

beniwohli:
@graphaelli re logging: can you use https://pypi.org/project/logging_tree/ to print the logger tree in both environments?

beniwohli:
jenkins test this

graphaelli:
The warnings (of course) goes away in 2.7 with psutil installed.  Without, here is the logging_tree printout:

python2.7:
```
>>> from elasticapm import Client
>>> client = Client()
No handlers could be found for logger "elasticapm.metrics"
>>> from logging_tree import printout
>>> printout()
<--""
   Level WARNING
   |
   o<--"elasticapm"
   |   Level NOTSET so inherits level WARNING
   |   |
   |   o<--[elasticapm.base]
   |   |   |
   |   |   o<--"elasticapm.base.Client"
   |   |       Level NOTSET so inherits level WARNING
   |   |
   |   o<--"elasticapm.errors"
   |   |   Level NOTSET so inherits level WARNING
   |   |
   |   o<--"elasticapm.metrics"
   |   |   Level NOTSET so inherits level WARNING
   |   |
   |   o<--"elasticapm.traces"
   |   |   Level NOTSET so inherits level WARNING
   |   |
   |   o<--"elasticapm.transport"
   |   |   Level NOTSET so inherits level WARNING
   |   |   |
   |   |   o<--"elasticapm.transport.http"
   |   |       Level NOTSET so inherits level WARNING
   |   |
   |   o<--"elasticapm.utils"
   |       Level NOTSET so inherits level WARNING
   |
   o<--"urllib3"
       Level NOTSET so inherits level WARNING
       Handler <logging.NullHandler object at 0x10af1e150>
       |
       o<--"urllib3.connection"
       |   Level NOTSET so inherits level WARNING
       |
       o<--"urllib3.connectionpool"
       |   Level NOTSET so inherits level WARNING
       |
       o<--"urllib3.poolmanager"
       |   Level NOTSET so inherits level WARNING
       |
       o<--"urllib3.response"
       |   Level NOTSET so inherits level WARNING
       |
       o<--[urllib3.util]
           |
           o<--"urllib3.util.retry"
               Level NOTSET so inherits level WARNING
```

PR #366, opened 2018-12-20
 Fix extract_signature when sql is a bytes object

ellisvalentiner:
In psycopg2's `execute_batch` and `execute_values` functions, the SQL statement is a bytes object and not a string (see psycopg2's [extras.py#L1198](https://github.com/psycopg/psycopg2/blob/60935b9b3d1a32da5ce0702f1bdb71eb148c2b34/lib/extras.py#L1198)). This may also be a problem for other database adapters but I'm only aware of this behavior with psycopg2.

`extract_signature` raises a TypeError when the `sql` parameter is a bytes object because `sql.find(" ")` expects a bytes object and `" "` is a string.

My proposed solution is to check whether `sql` is a bytes object and decode if necessary.

I have signed the CLA.

### DIFF
--- a/elasticapm/instrumentation/packages/dbapi2.py
+++ b/elasticapm/instrumentation/packages/dbapi2.py
@@ -123,6 +123,8 @@ def extract_signature(sql):
     :param sql: the SQL statement
     :return: a string representing the signature
     """
+    if isinstance(sql, bytes):
+        sql = sql.decode()
     sql = sql.strip()
     first_space = sql.find(" ")
     if first_space < 0:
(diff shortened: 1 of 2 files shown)


elasticapm/instrumentation/packages/dbapi2.py
```diff
@@ -123,6 +123,8 @@ def extract_signature(sql):
     :param sql: the SQL statement
     :return: a string representing the signature
     """
+    if isinstance(sql, bytes):
+        sql = sql.decode()
```
beniwohli: It might make sense to use `elasticapm.utils.encoding.force_text` here, as it will handle encoding errors somewhat gracefully

beniwohli:
Thanks for the fix @ellisvalentiner, it's now merged to master (after cleaning up a test that now failed), and I will push a release later today

PR #372, opened 2018-12-24
remove custom context of botocore instrumentation

beniwohli:
specifically, the `service` property clashes with the existing `context.service` property in the Elasticsearch storage backend

backport of #367 to 3.x


beniwohli:
merged into 3.x

PR #377, opened 2019-01-03
implement span subtypes and actions

beniwohli:
this implements elastic/apm#8

Closes https://github.com/elastic/apm-agent-python/issues/528


beniwohli:
jenkins test this

elasticapm/traces.py
```diff
@@ -159,6 +170,16 @@ def __init__(self, transaction, name, span_type, context=None, leaf=False):
         self.duration = None
         self.parent = None
         self.frames = None
+        if not span_subtype and "." in span_type:
```
graphaelli: I think we should allow an explicit empty subtype, like:

```suggestion
        if span_subtype is None and "." in span_type:
```

PR #378, opened 2019-01-03
don't set custom date in logging handler anymore

beniwohli:
support for this has been deprecated. The date/time will be set automatically


beniwohli:
jenkins test this

PR #385, opened 2019-01-21
make the internal context API a bit more explicit

beniwohli:


beniwohli:
This will be merged with #388

PR #388, opened 2019-01-22
Opentracing bridge

beniwohli:


elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+        if self._agent.config.instrument:
+            instrument()
+
+    def start_active_span(
+        self,
+        operation_name,
+        child_of=None,
+        references=None,
+        tags=None,
+        start_time=None,
```
jalvz: why bother propagating unused values (`start_time`, `ignore_active_span`)?

elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+        return scope
+
+    def start_span(
+        self, operation_name=None, child_of=None, references=None, tags=None, start_time=None, ignore_active_span=False
+    ):
+        if isinstance(child_of, OTSpanContext):
+            parent_context = child_of
+        elif isinstance(child_of, OTSpan):
+            parent_context = child_of.context
+        elif references and references[0].type == ReferenceType.CHILD_OF:
```
jalvz: where did you find about `references` in the API? i don't know what is this

elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+            parent_context = references[0].referenced_context
+        else:
+            parent_context = None
+        transaction = traces.execution_context.get_transaction()
+        if not transaction:
+            trace_parent = parent_context.trace_parent if parent_context else None
+            transaction = self._agent.begin_transaction("opentracing", trace_parent=trace_parent)
```
jalvz: not sure how meaningful is this `opentracing` transaction type... the default `custom` seems just as good to me, but probably makes sense to align with other agents (maybe you did that already)

elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+            span_context = OTSpanContext(trace_parent=transaction.trace_parent)
+            ot_span = OTSpan(self, span_context, transaction)
+        else:
+            parent_span_id = (
+                parent_context.span.elastic_apm_ref.id
+                if parent_context and parent_context.span and not parent_context.span.is_transaction
+                else None
+            )
+            span = transaction.begin_span(operation_name, None, parent_span_id=parent_span_id)
+            trace_parent = parent_context.trace_parent if parent_context else transaction.trace_parent
```
jalvz: just to wrap my head around this, `parent_context.trace_parent` is not `None` when the span's parent is propagated from another service, right? if so, `parent_context.trace_parent` is not `None` means that `transaction` is `None` (because this is the first span of a service)?

another way of putting it: i actually don't understand this bit :see_no_evil:

elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+            span_context = OTSpanContext(trace_parent=trace_parent.copy_from(span_id=span.id))
+            ot_span = OTSpan(self, span_context, span)
+        if tags:
+            for k, v in compat.iteritems(tags):
+                ot_span.set_tag(k, v)
+        return ot_span
+
+    def extract(self, format, carrier):
```
jalvz: why some errors return an exception (like carrier not supported) and others do not (like `elastic-apm-traceparent` header bad formatted)?

docs/opentracing.asciidoc
```diff
@@ -0,0 +1,133 @@
...
+==== Span References
+Currently, this bridge only supports `child_of` references.
+Other references,
+like `follows_from` are not supported yet.
+
+[float]
+[[opentracing-baggage]]
+==== Baggage
+The `Span.set_baggage_item(key, value)` method is not supported.
+Baggage items are silently dropped.
```
jalvz: how this exactly happens? i was expecting to have `set_baggage_item` method implemented as a noop, but i don't find it

elasticapm/contrib/opentracing/span.py
```diff
@@ -0,0 +1,91 @@
...
+        else:
+            logger.debug("Can't handle non-exception type opentracing logs")
+        return self
+
+    def set_operation_name(self, operation_name):
+        self.elastic_apm_ref.name = operation_name
+        return self
+
+    def set_tag(self, key, value):
+        if self.is_transaction:
```
jalvz: aren't a few known tags missing? `context.request.cookies`, `context.request.headers`, `context.response`...?

elasticapm/traces.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
     def end_transaction(self):
         self.duration = _time_func() - self.start_time
 
-    def begin_span(self, name, span_type, context=None, leaf=False, tags=None):
-        parent_span = get_span()
+    def begin_span(self, name, span_type, context=None, leaf=False, tags=None, parent_span_id=None):
```
jalvz: is this part of the agent API? if so, what do you think about making this private and keep a public `begin_span` without the `parent_span_id` argument?

docs/opentracing.asciidoc
```diff
@@ -0,0 +1,133 @@
...
+[[opentracing-scope-managers]]
+==== Scope Managers
+Currently, only the `ThreadLocalScopeManager` is supported.
+Using other scope managers will lead to unpredictable and possibly app-breaking behaviour.
+
+[float]
+[[opentracing-instrumentation]]
+==== Instrumentation
+
+It is recommended to not use the built-in instrumentations of Elastic APM together with third-party OpenTracing instrumentations like https://pypi.org/project/opentracing_instrumentation/[opentracing_instrumentation].
```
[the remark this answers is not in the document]
beniwohli: (side note: OpenTracing and distributed tracing aren't really related, and I don't think a goal of opentracing is to mix vendors. As far as I understand, the main goal of opentracing is to easily allow the user to _switch_ vendor without (too many) code changes, but it doesn't define a common trace propagation protocol like the W3C traceparent (hopefully-soon-)standard does, and it says nothing about storage or data model.)

Do I understand you correctly that you'd like to add "in the same service" to the above sentence? That seems reasonable.

elasticapm/context/threadlocal.py
```diff
@@ -1,30 +1,35 @@
 import threading
 
-thread_local = threading.local()
-thread_local.transaction = None
+from elasticapm.context.base import BaseContext
+
 elasticapm_span_var = None
```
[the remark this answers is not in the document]
beniwohli: because I forgot to remove it when refactoring this whole bit :) good catch!

elasticapm/contrib/opentracing/tracer.py
```diff
@@ -0,0 +1,96 @@
...
+from elasticapm.utils import compat, disttracing
+
+
+class Tracer(TracerBase):
+    _elasticapm_client_class = elasticapm.Client
+
+    def __init__(self, client_instance=None, config=None, scope_manager=None):
+        self._agent = client_instance or self._elasticapm_client_class(config=config)
+        if scope_manager and not isinstance(scope_manager, ThreadLocalScopeManager):
+            warnings.warn(
```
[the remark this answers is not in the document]
beniwohli: Well, things will fail in hilarious ways anyway if you use asyncio or gevent with a `ThreadLocalScopeManager`. With this, people have at least the choice what kind of hilarity they want. Also, I hope to get at least asyncio covered Real Soon Now :tm: (maybe even this decade!)

elasticapm/contrib/opentracing/tracer.py
beniwohli: Those arguments are defined by the base class in python-opentracing. And I hope to add support for them at some point.

elasticapm/contrib/opentracing/tracer.py
beniwohli: References are defined in the spec here: https://opentracing.io/specification/#references-between-spans

And in python-opentracing here: https://github.com/opentracing/opentracing-python/blob/f1e3cd6e7139536d4e85605f873af3f2f76fcce4/opentracing/tracer.py#L255

docs/opentracing.asciidoc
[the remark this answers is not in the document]
jalvz: ah, thanks for clarifying!

docs/opentracing.asciidoc
beniwohli: We inherit the noop-method from the super class

elasticapm/traces.py
beniwohli: Did you think of something like this? d33852df89124

docs/opentracing.asciidoc
beniwohli: done in 4ca885930dc280 (plus another few tests plus a fix for an issue I found thanks to those tests. "write more tests" is the lesson here I guess)

PR #390, opened 2019-01-23
Use native string for disttracing header

hannseman:
The value of `TRACEPARENT_HEADER_NAME` is set as a byte string. This causes the header dict in `requests` to have mixed string and byte values on Python 3. The expected behaviour is for both keys and values to be native strings, that is the `str` type for both Python 2 and 3.

This behaviour causes our application to crash due to us doing `json.dumps` on `request.headers`:

`TypeError: Object of type bytes is not JSON serializable`
```python
{
"Accept": '*/*',
"Accept-Encoding": 'gzip, deflate',
"Authorization": 'Filtered',
"Connection": 'keep-alive',
"elastic-apm-traceparent": b'00-706eb5e0137e6d165e10d964b00ea43c-70016c1f38b65b27-01',
"User-Agent": 'python-requests/2.19.1'
}
```

### DIFF
--- a/elasticapm/instrumentation/packages/urllib3.py
+++ b/elasticapm/instrumentation/packages/urllib3.py
@@ -56,7 +56,7 @@ class Urllib3Instrumentation(AbstractInstrumentedModule):
                 trace_parent = transaction.trace_parent.copy_from(
                     span_id=parent_id, trace_options=TracingOptions(recorded=True)
                 )
-                headers[constants.TRACEPARENT_HEADER_NAME] = trace_parent.to_ascii()
+                headers[constants.TRACEPARENT_HEADER_NAME] = trace_parent.to_str()
             return wrapped(*args, **kwargs)
 
     def mutate_unsampled_call_args(self, module, method, wrapped, instance, args, kwargs, transaction):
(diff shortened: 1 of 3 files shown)


beniwohli:
Hi @hannseman, thanks for opening this! One of your proposed changes (turninging `TraceParent.to_ascii()` into `TraceParent.to_string()`) is actually already part of an unrelated PR that I intend to merge today or Monday at the latest (#388).

The reason for this header being ASCII encoded in the first place is that, eventually, the header has to be a binary string. Python 3 will [encode it](https://github.com/python/cpython/blob/62c35a8a8ff5854ed470b1c16a7a14f3bb80368c/Lib/http/client.py#L1185-L1194) automatically, but that doesn't happen in [Python 2](https://github.com/python/cpython/blob/06b15424b0dcacb1c551b2a36e739fffa8d0c595/Lib/httplib.py#L1005).

But you are right that if we just use the "native" `str` type, we should be good. In Python 2, it is already a bytestring, and on Python 3, it'll be encoded to one automatically (and since it is only a bunch of hexadecimal string representations, it should be perfectly safe to automatically encode to ASCII).

I'll merge this once #388 has landed. Thanks again for your contribution!

beniwohli:
jenkins test this

hannseman:
@beniwohli sounds good! Looks like I forgot to run black. Should be fixed now.

beniwohli:
jenkins test this

hannseman:
@beniwohli I see that the jenkins builds are failing but from I can gather the errors are unrelated to this PR?

PR #392, opened 2019-01-28
introduces `capture_headers` config option

### LINKED ISSUE #365: Add capture headers config option
The `capture_headers` should default to `true` (which is also the current behavior). If set to `false`, HTTP headers (and potentially other headers) should not be logged on transactions, spans or errors.

Besides being documented on the configuration page, this config option should also be documented on the performance optimization page.

beniwohli:
closes #365

### DIFF
--- a/docs/tuning.asciidoc
+++ b/docs/tuning.asciidoc
@@ -95,4 +95,17 @@ There are four settings you can modify to control this behavior:
 
 As you can see, these settings are divided between app frames, which represent your application code, and library frames, which represent the code of your dependencies. Each of these categories are also split into separate error and span settings. 
 
-Reading source files inside a running application can cause a lot of disk I/O, and sending up source lines for each frame will have a network and storage cost that is quite high. Turning down these limits will help prevent excessive memory usage. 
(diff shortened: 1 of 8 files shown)


docs/tuning.asciidoc
```diff
@@ -95,4 +95,17 @@ There are four settings you can modify to control this behavior:
...
+
+[float]
+[[tuning-body-headers]]
+=== Collecting headers and request body
+
+You can configure the Elastic APM agent to capture headers of both requests and responses (<<config-capture-headers,`capture_headers`>>),
+as well as request bodies (<<config-capture-body,`capture_body`>>).
+By default, capturing request bodies is disabled.
+Enabling it for transactions may introduce noticeable overhead, as well as increased storage use, depending on the nature of your POST requests.
+In most scenarios, we advise against enabling request body capturing for transactions, and only enable it if necessary for errors.
```
felixbarny: If it's only enabled for errors, does the Python agent actually only read it in case there was an error? Or would it ready anyway and discard if there was no error?

felixbarny: Ah, found the answer in `self.client.capture_exception`. The agent would only read the request body in case of an exception.

PR #394, opened 2019-01-29
store the URL of instrumented requests/urllib3 spans in the correct place

### LINKED ISSUE #393: move context.url to context.http.url
While investigating the impact of https://github.com/elastic/apm-server/issues/411, I realized instrumentation of  [httplib3](https://github.com/elastic/apm-agent-python/blob/d215428603c0aaed3de9cc3fdb2fcb3f4583ad85/elasticapm/instrumentation/packages/urllib3.py#L46) and [requests](https://github.com/elastic/apm-agent-python/blob/d215428603c0aaed3de9cc3fdb2fcb3f4583ad85/elasticapm/instrumentation/packages/requests.py#L31) creates `context.url`.  By convention, that should have been going to `context.http.url`.  If the referenced issue is implemented, the convention will be codified and this information will start to be dropped, so let's move it either way.

beniwohli:
fixes #393

### DIFF
--- a/elasticapm/instrumentation/packages/requests.py
+++ b/elasticapm/instrumentation/packages/requests.py
@@ -28,5 +28,5 @@ class RequestsInstrumentation(AbstractInstrumentedModule):
         signature = request.method.upper()
         signature += " " + get_host_from_url(request.url)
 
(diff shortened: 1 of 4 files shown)


beniwohli:
jenkins test this

PR #395, opened 2019-01-29
use xenial base image for testing on Travis

beniwohli:
The SQLite version that comes with Ubuntu Precise is no longer supported
by new versions of Django


beniwohli:
Closed in 3fc6464e

PR #410, opened 2019-02-11
reduce code protected by lock as much as possible

beniwohli:
this avoids deadlocks that are caused by code inside the lock
triggering log messages that could be captured and queued, which in turn
leads to a deadlock.

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -6,8 +6,7 @@ import timeit
 from collections import defaultdict
 
 from elasticapm.contrib.async_worker import AsyncWorker
-from elasticapm.utils import json_encoder
-from elasticapm.utils.compat import BytesIO
+from elasticapm.utils import compat, json_encoder
 
 logger = logging.getLogger("elasticapm.transport")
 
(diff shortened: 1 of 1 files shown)


elasticapm/transport/base.py
```diff
@@ -100,28 +98,28 @@ def flush(self, sync=False, start_flush_timer=True):
...
-                    try:
-                        self.send(data)
-                        self.handle_transport_success()
-                    except Exception as e:
-                        self.handle_transport_fail(e)
-            self._last_flush = timeit.default_timer()
-            if start_flush_timer:
-                self._start_flush_timer()
+        if queued_data and not self.state.should_try():
+            logger.error("dropping flushed data due to transport failure back-off")
```
graphaelli: to be clear, this is the important line to move outside?

beniwohli: that one, and the two `Starting/Cancelling flush timer`

graphaelli:
For visibility - ran this branch with apm-integration-testing and it appears the issue remains.  That shouldn't stop this from being merged but we can't call it case closed.

beniwohli:
Retracting this in favor of a more complete refactor of the transport queuing logic

PR #411, opened 2019-02-12
introduce an event queue and a processing thread into transports

beniwohli:
Using this queue, we can avoid that any other thread than the processing
thread will access the compressed StringIO queue, making coarse blocking
unnecessary. This should avoid dead locks.


elasticapm/transport/base.py
```diff
@@ -54,74 +53,93 @@ def __init__(
         self._max_flush_time = max_flush_time
         self._max_buffer_size = max_buffer_size
         self._queued_data = None
-        self._queue_lock = threading.Lock()
+        self._event_queue = compat.queue.Queue(maxsize=100)
```
beniwohli: wondering if this should be configurable.

@axw is it in the Go agent? Also, how did you determine the (default) size of the queue (or ring buffer in your case IIRC)

tests/transports/test_base.py
```diff
@@ -77,88 +82,85 @@ def test_flush_time(mock_send, caplog):
...
 
 
 @mock.patch("elasticapm.transport.base.Transport.send")
 def test_forced_flush(mock_send, caplog):
+    transport = Transport(metadata={}, max_buffer_size=1000, compress_level=0)
     with caplog.at_level("DEBUG", "elasticapm.transport"):
-        transport = Transport(metadata={}, max_buffer_size=1000, compress_level=0)
         transport.queue("error", "x", flush=True)
-    record = caplog.records[0]
-    assert "forced" in record.message
```
beniwohli: Unfortunately, it seems the caplog fixture has some difficulties in multithreaded environments (or I have difficulties using it correctly), so I had to remove some of these asserts

elasticapm/transport/base.py
axw: I have separate queues for transactions, spans, and errors; mostly for type-safety. Each queue has a capacity of 1000. I don't think I was very scientific about the choice, just eyeballed a high-throughput test app to ensure it didn't drop a significant amount.

Originally I made them configurable, but the number of queues and buffers increased to the point that I thought it would be unhelpful to expose it initially.

The Go agent has this initial queue, then a ring buffer, and then finally the gzip request buffer. I think that's all of them ;)

elasticapm/handlers/logging.py
```diff
@@ -45,7 +45,7 @@ def emit(self, record):
         self.format(record)
 
         # Avoid typical config issues by overriding loggers behavior
-        if record.name.startswith("elasticapm.errors"):
+        if record.name.startswith(("elasticapm.errors",)):
```
axw: what's this for?

elasticapm/transport/base.py
```diff
@@ -135,7 +152,10 @@ def close(self):
         Cleans up resources and closes connection
         :return:
         """
-        self.flush(sync=True, start_flush_timer=False)
+        self._closed = True
+        self.queue("close", None)
+        if self._event_process_thread.is_alive():
+            self._event_process_thread.join()
```
axw: Unless you ensure the queue above didn't drop, this could block forever. If you do add a timeout to `close`, I think you would need to measure how long the `queue` call takes, and subtract that from the timeout that you then pass to `join`.

elasticapm/transport/base.py
```diff
@@ -54,74 +53,92 @@ def __init__(
...
+            timeout = max(0, self._max_flush_time - since_last_flush) if self._max_flush_time else None
+            timed_out = False
+            try:
+                event_type, data, flush = self._event_queue.get(block=True, timeout=timeout)
+            except compat.queue.Empty:
+                event_type, data, flush = None, None, None
+                timed_out = True
+
+            if event_type == "close":
+                self.flush(sync=True)
```
axw: Use the `flush` value sent through the queue, as for other events?

elasticapm/transport/base.py
```diff
@@ -54,74 +53,92 @@ def __init__(
...
         self._counts = defaultdict(int)
+        self._closed = False
 
     def queue(self, event_type, data, flush=False):
-        with self._queue_lock:
-            queued_data = self.queued_data
-            queued_data.write((self._json_serializer({event_type: data}) + "\n").encode("utf-8"))
-            self._counts[event_type] += 1
+        if not self._event_process_thread.is_alive() and not self._closed:
+            self._event_process_thread.start()
```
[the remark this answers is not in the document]
beniwohli: Yes, that might be an issue. I had it in the constructor at first, but that made testing extremely cumbersome.

Maybe a flag in the constructor, `start_processing_thread` that defaults to `True` could help with that.

elasticapm/transport/base.py
```diff
@@ -135,7 +158,18 @@ def close(self):
...
-        self.flush(sync=True, start_flush_timer=False)
+        self._closed = True
+        self.queue("close", None)
+        if self._event_process_thread.is_alive():
+            self._event_process_thread.join()
+
+    def flush(self):
+        """
+        Trigger a flush of the queue.
+        Note: the flush happens asynchronously in a background thread, which means that the queue won't be immediately
```
axw: In the Go agent I make the Flush method block until the server has sent the queued events to the server, successfully or not. I did this primarily to make testing simpler and more robust. The way I did it is roughly this:

- when `flush` is called, check if the transport has a `flushed` event attribute already. If not, set a new one, and then send the flush request to the queue and wait on the event to be signalled
- when the server has finished processing the flush request, signal the event and clear the `flushed` event field

You might want to do this eventually, otherwise you'll probably run into intermittent failures on slow CI machines due to the sleeps.

elasticapm/transport/base.py
```diff
@@ -135,7 +175,22 @@ def close(self):
...
+            return
+        self._closed = True
+        self.queue("close", None)
+        if not self._flushed.wait(timeout=self._max_flush_time):
+            raise ValueError("close timed out")
+
+    def flush(self):
+        """
+        Trigger a flush of the queue.
+        Note: this method will only return once the queue is empty. This means it can block indefinitely if more events
```
axw: I know this is only meant for tests, but FWIW: in the Go agent, flush just flushes the _currently_ enqueued events. When  a flush command is received by the background thread, it gets the current size of the queue and drains it by that many events. Even that's not necessary here because you have a single queue (whereas in Go we have one per event type, and a separate event/condition variable for flushing.)

PR #420, opened 2019-03-12
Fix `set_user_context` API doc

messense:
https://github.com/elastic/apm-agent-python/blob/d0568eac3bb8f55084c3ca2fa733f86cda32ffc5/elasticapm/traces.py#L388-L396


beniwohli:
Excellent find, thanks @messense!

Issue #409, opened 2019-02-10
v4.1.0 Introduces `http.client.RemoteDisconnected` error


jensenbox:
**Describe the bug**: ...
Events getting reported to APM-Server are randomly getting rejected. It appears that every so often the event getting reported is being met with a RemoteDisconnected exception.

**To Reproduce**

1. Setup a django project as per directions
2. Hit the homepage of the project
3. Wait for several events to hit the APM server
4. One will eventually cause the exception

**Expected behavior**: ...
No Exception

**Environment (please complete the following information)**
- OS: Mojave but running under `python:3` docker container
- Python version: 3.7.2
- Framework and version: Django version 2.1.5
- APM Server version: apm-server version 6.6.0 (amd64), libbeat 6.6.0
- Agent version: 4.1.0

**Additional context**

downgrading to 4.0.3 remedies the issue.

Directly related: https://discuss.elastic.co/t/unable-to-reach-apm-server-connection-aborted-remotedisconnected-remote-end-closed-connection-without-response/166500

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```
    ELASTIC_APM = {
        'SERVICE_NAME': 'web-backend',
        'SECRET_TOKEN': '8149e743-2524-2524-2524-94aadd3fcde7',
        'SERVER_URL': 'http://192.168.16.100:8200',
        'DEBUG': True,
    }

    ```
  </details>
- `requirements.txt`:
  <details>
    <summary>Click to expand</summary>

     ```
     django-allauth
    django-guardian
    django-qr-code
    django-icons
    django-phonenumber-field
    django-crispy-forms
    django-autocomplete-light
    django-money
    django-anymail
    django-debug-toolbar
    django-extensions
    django-countries
    djangorestframework
    psycopg2-binary
    stripe
    geoip2
    redis
    celery
    sentry-sdk
    googlemaps
    django
    pillow
    phonenumbers
    gunicorn
    whitenoise
    elastic-apm
     ```
  </details>

```
web_1      | Failed to submit message: "Unable to reach APM Server: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response')) (url: http://192.168.16.100:8200/intake/v2/events)"
web_1      | Traceback (most recent call last):
web_1      |   File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 600, in urlopen
web_1      |     chunked=chunked)
[...]

PR #422, opened 2019-03-13
set SO_KEEPALIVE on underlying TCP socket of http connection to APM Server

beniwohli:
this is a tentative fix for #409


beniwohli:
This didn't work out as hoped, closing

PR #430, opened 2019-03-20
can't always assume our log message is on index 1 :/

beniwohli:


beniwohli:
closing in favor of #431

PR #435, opened 2019-03-21
don't use ChilledQueue if queue.Queue has been monkeypatched

beniwohli:
this should fix issues with eventlet. To this effect, a bare bones
test for eventlet was added.

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -89,6 +89,7 @@ class Transport(object):
         self._max_buffer_size = max_buffer_size
         self._queued_data = None
         self._event_queue = self._init_event_queue(chill_until=queue_chill_count, max_chill_time=queue_chill_time)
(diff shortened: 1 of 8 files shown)


elasticapm/transport/base.py
```diff
@@ -105,18 +105,17 @@ def __init__(
     def queue(self, event_type, data, flush=False):
         try:
             self._flushed.clear()
-            self._event_queue.put((event_type, data, flush), block=False, chill=not (event_type == "close" or flush))
+            if isinstance(self._event_queue, ChilledQueue):
```
graphaelli: is this cheap enough to do each time / could this change during the runtime of the application?  Seems like this could be decided once at init and have a different queue action based onthat.

beniwohli: good point, changed in e18a74a

PR #444, opened 2019-04-01
use worker_init/worker_process_init to start event processor threads in celery

beniwohli:
Celery uses a prefork model for its worker processes. The app is loaded
in a "worker", which then forks one or more "worker processes". Similiarly
to the uwsgi prefork issues we had to work around, this fork happens
_after_ we created the event processor thread, so the worker processes
don't have an event processor thread of their own, and just fill up the
event queue without it ever being processed.

To work around this, we use worker_init and worker_process_init signals
to start the event processor thread _after_ the fork.

This only became an issue with version 4.2.0 of the agent. In earlier
versions of the agent, threads were started relatively late, after
celery already forked the worker processes.

### DIFF
--- a/elasticapm/contrib/celery/__init__.py
+++ b/elasticapm/contrib/celery/__init__.py
@@ -52,6 +52,7 @@ def register_exception_tracking(client):
 
     signals.task_failure.disconnect(process_failure_signal, dispatch_uid=dispatch_uid)
     signals.task_failure.connect(process_failure_signal, weak=False, dispatch_uid=dispatch_uid)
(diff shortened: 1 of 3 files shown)


beniwohli:
> Did you try this with the non-prefork pool types? (solo, eventlet, gevent)

Not yet, good point. These signals are only fired for the pre-fork worker model AFAIK, so at least it shouldn't make things worse for the other worker pool types.

beniwohli:
Update: solo works, eventlet and gevent are missing spans, but this was always the case, as we don't currently support eventlet/gevent.

Issue #450, opened 2019-04-05
RuntimeError: can't start new thread


rvanlaar:
**Describe the bug**: We're getting errors "RuntimeError: can't start new thread".
The only change we've introduced is turning Elastic on. 
Since than we're getting these errors. When running long enough, this
exhausts the thread pool on the server. 

**To Reproduce**

1. Use APM with Django

**Expected behavior**: To not run out of threads.

**Environment (please complete the following information)**
- OS: Ubuntu 16.04.5 LTS
- Python version: 3.5.2
- Framework and version [e.g. Django 2.1]: Django 1.11.20
- APM Server version: elastic cloud
- Agent version: elastic-apm 4.2.1

**Additional context**

RuntimeError: can't start new thread
  File "threading.py", line 1180, in run
    self.function(*self.args, **self.kwargs)
  File "elasticapm/transport/base.py", line 124, in flush
    self._start_flush_timer()
  File "elasticapm/transport/base.py", line 160, in _start_flush_timer
    self._flush_timer.start()
  File "threading.py", line 844, in start
    _start_new_thread(self._bootstrap, ())

We're running with celery and sentry. 

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```
   if env('ELASTIC_APM'):
    ELASTIC_APM = {
        # Set required service name. Allowed characters:
        # a-z, A-Z, 0-9, -, _, and space
        'SERVICE_NAME': 'CLIENT PRODUCTION',
        # Use if APM Server requires a token
        'SECRET_TOKEN': env('ELASTIC_APM_SECRET_TOKEN'),
        # Set custom APM Server URL (default: http://localhost:8200)
        'SERVER_URL': env('ELASTIC_APM_SERVER_URL)
    }

    INSTALLED_APPS += ('elasticapm.contrib.django',)
    # The Elastic APM tracing middleware needs to be first.
    MIDDLEWARE_CLASSES = (
        'elasticapm.contrib.django.middleware.TracingMiddleware',
        ) + MIDDLEWARE_CLASSES
    ```
  </details>

beniwohli:
Hi @rvanlaar 

[...]

PR #452, opened 2019-04-09
introduce `IntervalTimer` and use it instead of `threading.Timer`

beniwohli:
`threading.Timer` only runs the function once, and then stops the thread.
To get an interval timer, we used to start a new thread when the original
thread expired. This is wasteful, and in the worst case can lead to
runoff creation of timer threads (#450).

Instead, the `IntervalTimer` will run in an endless loop, executing the
given function every n seconds.


beniwohli:
closed in 5337c43

PR #453, opened 2019-04-09
Execution context - use thread locals if gevent/eventlet has monkey patched it

ramshaw888:
* Extract out logic to decide which backing to use for `execution_context` into a function `init_execution_context`
* Check if `gevent` or `eventlet` has monkey patched `_threading.local`, if it has then use `elasticapm.context.threadlocal` as the backing.

PR #454, opened 2019-04-09
Execution context - use thread locals if gevent/eventlet has monkey patched it

ramshaw888:
Re-PR of #453 

* Extract out logic to decide which backing to use for execution_context
into a function init_execution_context
* Check if gevent or eventlet has monkey patched _threading.local, if it
has then use elasticapm.context.threadlocal as the backing.

### DIFF
--- a/elasticapm/context/__init__.py
+++ b/elasticapm/context/__init__.py
@@ -27,3 +27,40 @@
 #  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 #  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+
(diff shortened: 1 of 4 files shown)


ramshaw888:
@beniwohli I've added the eventlet check as well. I've squashed my commits with the black/isort/flake8 stuff fixed up (as well as some commit signing stuff).

Jenkins seems to be erroring on the checkout stage for this and other PRs - is that intentional?

While travis reported OK in the status check, looks like theres something failing with django versions on master https://travis-ci.org/elastic/apm-agent-python/jobs/518036683

beniwohli:
Awesome, thanks @ramshaw888.

I had to fix another small linting issue, all code files require a license header. I added that to the two files you created.

The "failure" on Jenkins is a security precaution. It'll fail automatically for PRs that aren't from members of the elastic Github org, and needs to be re-run manually after reviewing the code. This should avoid things like bad actors submitting PRs that run cryptominers on our infrastructure.

The failure on Django master can be ignored for now. After releasing Django 2.2 recently, they bumped the minimal Python version for the next Django version.

ramshaw888:
Checks out! Thanks for the review and merge :)

PR #461, opened 2019-04-23
fix copyright header check with new grep versions

beniwohli:
with grep 3.2, a change was introduced that essentially inverts the
exit code returned by grep when using the --files-without-match flag.
To get around this, we calculate our own exit code based on the stdout
output from grep.


kuisathaverat:
I'm late, but LGTM :)

PR #464, opened 2019-04-23
instrument urllib2 / urllib.request

### LINKED ISSUE #462: Capture Python  request.urlopen runtime metrics
Python agent can capture  requests.get /  requests.post runtime metrics . but  request.urlopen can not capture  runtime metrics
Thanks

beniwohli:
fixes #462

### DIFF
--- a/elasticapm/instrumentation/packages/urllib.py
+++ b/elasticapm/instrumentation/packages/urllib.py
@@ -1,0 +1,102 @@
+#  BSD 3-Clause License
+#
+#  Copyright (c) 2019, Elasticsearch BV
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+#  * Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+#  * Neither the name of the copyright holder nor the names of its
+#    contributors may be used to endorse or promote products derived from
+#    this software without specific prior written permission.
+#
+#  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+#  AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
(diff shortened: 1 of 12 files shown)


elasticapm/instrumentation/packages/urllib.py
```diff
@@ -0,0 +1,102 @@
...
+    return host
+
+
+class UrllibInstrumentation(AbstractInstrumentedModule):
+    name = "urllib"
+
+    if compat.PY2:
+        instrument_list = [("urllib2", "AbstractHTTPHandler.do_open")]
+    else:
+        instrument_list = [("urllib.request", "AbstractHTTPHandler.do_open")]
```
axw: Should we be using [urllib.request.OpenerDirector.add_handler](https://docs.python.org/3/library/urllib.request.html#urllib.request.OpenerDirector.add_handler) for Python 3? That would make it possible to capture time spent in pre/post-processors, I think. Not sure how common that is - maybe not worthwhile.

elasticapm/instrumentation/packages/urllib.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def call(self, module, method, wrapped, instance, args, kwargs):
+        request_object = args[1] if len(args) > 1 else kwargs["req"]
+
+        method = request_object.get_method()
+        host = request_host(request_object)
+
+        url = request_object.get_full_url()
```
axw: Is it possible for this to include a user:password? as in "http://user:pass@foo.bar/baz". It would be good to see a test that ensures we don't leak application secrets.

beniwohli: urllib (as well as urllib3) doesn't support encoding user/password in the URL, so this would only be an issue for code that is buggy in the first place.

`requests` OTOH does support this according to my testing (even though it isn't documented in https://2.python-requests.org//en/master/user/authentication/#basic-authentication), so we need the sanitization code anyway.

beniwohli: done in e8826da

tests/utils/tests.py
```diff
@@ -141,3 +142,8 @@ def x(self, x):
 
 def test_get_name_from_func_lambda():
     assert "tests.utils.tests.<lambda>" == get_name_from_func(lambda x: "x")
+
+
+def test_url_sanitization():
+    sanitized = sanitize_url("http://user:pass@localhost:123/foo?bar=baz#bazzinga")
```
axw: Would you mind throwing in some percent-encoded bytes in the password?

It appears that urlparse doesn't unescape them, but it would be good to have a test to ensure that remains the case over time, in case the implementation of sanitize_url changes.

beniwohli: added in 388d6c2c

PR #471, opened 2019-04-29
adapt opentracing bridge to some minor changes in opentracing-python 2.1.0

beniwohli:
specifically, https://github.com/opentracing/opentracing-python/pull/101

### DIFF
--- a/elasticapm/contrib/opentracing/span.py
+++ b/elasticapm/contrib/opentracing/span.py
@@ -30,12 +30,22 @@
 
 import logging
 
-from opentracing.ext import tags
 from opentracing.span import Span as OTSpanBase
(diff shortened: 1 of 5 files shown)


elasticapm/contrib/opentracing/span.py
```diff
@@ -49,17 +59,20 @@ def __init__(self, tracer, context, elastic_apm_ref):
             context.span = self
 
     def log_kv(self, key_values, timestamp=None):
+        exc_type, exc_val, exc_tb = None, None, None
```
roncohen: do you need this line?

roncohen: ah yes, didn't realise it's an "elif". Nevermind

Issue #468, opened 2019-04-28
Parent DroppedSpan causing AttributeError


nchudleigh:
**Describe the bug**: ...

A parent span being a DroppedSpan is causing an AttributeError to be called within end_span.  I cannot find the reason why.  Randomly started happening at 4am on a Sunday (in other words not connected to rollout).  Happened across many different task contexts.  Believe it was connected to some connection issues with the instrumented code.  In anycase the code should protect against this type of problem

Here is the trace from Sentry:
![image](https://user-images.githubusercontent.com/5580824/56857022-c2915b80-6935-11e9-8153-4451793619b9.png)

**To Reproduce**

I have not been able to reproduce, but changing the check to a `hasattr(parent, "id")` should be sufficient.

**Expected behavior**: ...

Don't throw errors when ending span.

**Environment (please complete the following information)**
- OS: [e.g. Linux] Ubuntu
- Python version: 2.7
- Framework and version [e.g. Django 2.1]: Flask 0.12
- APM Server version: 6.6.1
- Agent version: 4.1.0

**Additional context**

- Agent config options 
  <details>
    <summary>Click to expand</summary>

    ```python
    TRANSACTIONS_IGNORE_PATTERNS = ["^OPTIONS "]
    SERVER_URL = "http://apm:8200"
    ```
  </details>
- `requirements.txt`:
  <details>
    <summary>Click to expand</summary>

    ```
    elastic-apm==4.1.0
    Flask==0.12
    ```
  </details>

nchudleigh:
Heres an example of the fix that I mentioned in the description.  Obviously this is a result of something off further up the chain that someone more experienced should diagnose.  But this is a good quick fix for that gnarly one-liner.
```python
    def to_dict(self):
        if self.parent_span_id:
            parent_id = self.parent_span_id
        elif self.parent and hasattr(self.parent, "id"):
            parent_id = self.parent.id
        else:
            parent_id = self.transaction_id

        result = {
            "id": self.id,
            "transaction_id": self.transaction.id,
            "trace_id": self.transaction.trace_parent.trace_id,
            # use either the explicitly set parent_span_id, or the id of the parent, or finally the transaction id
            "parent_id": parent_id,
            "name": encoding.keyword_field(self.name),
            "type": encoding.keyword_field(self.type),
            "timestamp": int(self.timestamp * 1000000),  # microseconds
            "duration": self.duration * 1000,  # milliseconds
        }
```

beniwohli:
[...]

PR #478, opened 2019-05-08
add some handling and debugging info for erroneous DroppedSpan parents

beniwohli:
refs elastic/apm-agent-python#468


basepi:
@beniwohli I don't think this fix is necessary any longer. #468 should be fixed by #602. I'm going to close this, but feel free to re-open if you think it would still be a useful add!

PR #481, opened 2019-05-20
agent-specific metrics [WIP]

beniwohli:


elasticapm/metrics/sets/gc.py
```diff
@@ -0,0 +1,74 @@
...
+        for k, v in compat.iteritems(delta):
+            self.counter(k).inc(v)
+        return super(GarbageMetrics, self).collect()
+
+    def get_gc_stats(self):
+        stats = {}
+        if hasattr(gc, "get_stats"):  # only available in Python 3.4+
+            generations = gc.get_stats()
+            for i, generation in enumerate(generations):
+                stats["python.gc.gen{}.collections".format(i + 1)] = generation["collections"]
```
axw: Rather than including the generation number in the name, this is where labels are meant to be used. See `jvm.gc.count` for an example, under https://www.elastic.co/guide/en/apm/agent/java/current/metrics.html#metrics-jvm.

elasticapm/metrics/sets/gc.py
```diff
@@ -0,0 +1,74 @@
...
+                stats["python.gc.gen{}.collected".format(i + 1)] = generation["collected"]
+                stats["python.gc.gen{}.uncollectable".format(i + 1)] = generation["uncollectable"]
+        return stats
+
+    def gc_callback(self, phase, info):
+        if phase == self.START:
+            self.gc_start = timeit.default_timer()
+        elif phase == self.STOP and self.gc_start:
+            pause_time = timeit.default_timer() - self.gc_start
+            self.counter("python.gc.pause_time.ms").inc(pause_time * 1000.0)
```
axw: Do you think it would be useful to label this metric with `info['generation']`, the oldest generation being collected? Pause times will likely differ accordingly, right?

elasticapm/metrics/sets/gc.py
beniwohli: @axw is there an example payload somewhere for this? This would mean I'd have to split up the metrics by label into 3 metric sets, right?

axw: Handily there is an example in the Java agent's javadoc:

https://github.com/elastic/apm-agent-java/blob/b8d4ee1e7de5420caab010cda5e50314e1522367/apm-agent-core/src/main/java/co/elastic/apm/agent/metrics/MetricSet.java#L38-L42

> This would mean I'd have to split up the metrics by label into 3 metric sets, right?

Yes

beniwohli: > >  This would mean I'd have to split up the metrics by label into 3 metric sets, right?

> Yes

But isn't that bad news for storage efficiency? IIRC, the more metrics per document, the more efficient it is to store. If we split up the generation numbers, we'll store 4 documents of GC data per collection interval (one per generation, plus an unlabled one).

axw: It's not _ideal_ for storage, but with a 30 second interval, this is in the order of a few megabytes of storage per day per agent.

Using labels gives the option of aggregating the values _and_ getting a breakdown per generation. In Metrics Explorer you would chart `python.gc.collections` and then optionally group by `labels.generation`.

elasticapm/metrics/sets/gc.py
roncohen: As a user, i might want to alert on total GC time spent, for instance, per minute. If my application spends more than 5 seconds in GC over 2 minutes, i want to know about it. Splitting GC time into generations here would make it very complicated to calculate the sum of time spent in GC across generations. We'd essentially create a metrics stream per generation. To get the sum of them, you'd need to do a derivative for each one and then take the sum of them. If you're graphing them, a stacked graph might do, but for alerting, this is complicated.

It might be worth considering reporting the time spent in GC since the last collection instead. This is similar to the discussion we had on aggregated views (reporting deltas since collection versus ever increasing counters). I know the Java agent already shipped with other metrics, but i think it's worth having the discussion. ccing @felixbarny for input

roncohen: also @exekias

felixbarny: The Java agent currently reports total time spent in GC but I'm very happy to revisit that. In fact, I think it makes more sense to store the deltas when dealing with counters as it's easier to query (no need for a series_id/derivative agg. etc). But we'd need to find a way to make that change in a backwards compatible way, for example, by reporting both the cumulative pause time, like it's done today, and additionally report the deltas, for example in `*.gc.time.delta`.

roncohen:
I'm good to go with labels. There's a higher storage cost, but doesn't seems like it would be significant. The labeling approach has advantages wrt. aligning with other agents on metric names, being able to dynamically draw a graph per generation etc.

beniwohli:
closing this as it is out of date

PR #485, opened 2019-06-04
Add missing last back-tick (documentation)

antoniobusrod:


beniwohli:
Thanks @antoniobusrod, good catch!

Issue #500, opened 2019-06-24
System metrics not available on linux host with kernel < 3.14


NielkS:
**Bug description**
I see the following exception twice whenever I start a (django) application and execute a transaction for the first time.

> Exception in thread eapm metrics collect timer:
> Traceback (most recent call last):
>   File "/opt/pack/PYT/envs/pyt-3.6.6.1/lib/python3.6/threading.py", line 916, in _bootstrap_inner
>     self.run()
>   File "/data/nielks/virtualenv/lib/python3.6/site-packages/elasticapm/utils/threading.py", line 64, in run
>     self._function(*self._args, **self._kwargs)
>   File "/data/nielks/virtualenv/lib/python3.6/site-packages/elasticapm/metrics/base_metrics.py", line 86, in collect
>     data = metricset.collect()
>   File "/data/nielks/virtualenv/lib/python3.6/site-packages/elasticapm/metrics/sets/cpu_linux.py", line 73, in collect
>     self.gauge("system.memory.actual.free").val = new["mem_available"]
> KeyError: 'mem_available'

As a result, no information related to CPU & memory usage is collected.

**Environment (please complete the following information)**
- OS: Linux, kernel 3.0.101-108.18-default 
- Python version: 3.6.6
- Framework and version  Django 2.1.4 
- APM Server version: 7.1.0
- Agent version: 4.2.2

**Additional context**
Looking at the implementation of elasticapm/metrics/sets/cpu_linux.py, I see that the "mem_available" gauge is evaluated from the "MemAvailable" line of /proc/meminfo.
That info was added in https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0a, release in 2014 in kernel 3.14.
My host machine has a kernel older than this, hence the exception.

I managed to implement a simple fix, that works fine on my machine:

```
    def collect(self):
        new = self.read_process_stats()
        new.update(self.read_system_stats())
        with self._read_data_lock:
             # ...
            mem_available = new["mem_available"] if "mem_available" in new else new["mem_free"]
            self.gauge("system.memory.actual.free").val = mem_available
            # ...

    def read_system_stats(self):
        stats = {}
        # ...
        with open(self.memory_stats_file, "r") as memfile:
            for line in memfile:
                if line.startswith("MemTotal:"):
                    stats["mem_total"] = int(whitespace_re.split(line)[1]) * 1024
                elif line.startswith("MemAvailable:"):
[...]

PR #510, opened 2019-06-28
#500 - Linux Metric - Memory: evaluate free memory from MemFree when …

NielkS:
…MemAvailable not mentioned in /proc/meminfo

### DIFF
--- a/elasticapm/metrics/sets/cpu_linux.py
+++ b/elasticapm/metrics/sets/cpu_linux.py
@@ -40,6 +40,7 @@ MEM_STATS = "/proc/meminfo"
 PROC_STATS = "/proc/self/stat"
 
 CPU_FIELDS = ("user", "nice", "system", "idle", "iowait", "irq", "softirq", "steal", "guest", "guest_nice")
+MEM_FIELDS = ("MemTotal", "MemAvailable", "MemFree", "Buffers", "Cached")
 
 whitespace_re = re.compile(r"\s+")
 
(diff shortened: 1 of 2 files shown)


beniwohli:
jenkins test this please

NielkS:
HI @beniwohli ,
Not sure why the build fails - maybe related to my force-push ?
Please let me know if I should create another PR with "clean" commit history.

elasticapm/metrics/sets/cpu_linux.py
```diff
@@ -70,7 +70,9 @@ def collect(self):
             delta = {k: new[k] - prev[k] for k in new.keys()}
             cpu_usage_ratio = delta["cpu_usage"] / delta["cpu_total"]
             self.gauge("system.cpu.total.norm.pct").val = cpu_usage_ratio
-            self.gauge("system.memory.actual.free").val = new["mem_available"]
+            # mem_available not present in linux before kernel 3.14 - fallback to mem_free if not present - see #500
```
axw: At least some of the other agents are are using `Free + Buffers + Cached` as the result, when `MemAvailable` is not present. Would you please change this to do the same for consistency?

For example, this is the relevant code for the Go agent: https://github.com/elastic/go-sysinfo/blob/master/providers/linux/memory_linux.go#L71

NielkS: Thanks for the pointer, will do.

PR #518, opened 2019-07-04
feat(jenkins): support check license with go-licenser

v1v:
## Highlights
- Enable the go-licenser stage in the pipeline
- Enable the go-licenser goal in the makefile.

## Tasks
- [ ] Agree whether this is the approach to go
- [ ] It requires https://github.com/elastic/go-licenser/issues/24 to support the python headers.
- [ ] Makefile and Pipeline could potentially use the same scripts somehow.
- [ ] Pre-commit should be adapted to the go-licenser

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,6 +77,13 @@ pipeline {
                 ./tests/scripts/docker/black.sh
                 """, label: "Black code formatting"
                 sh script: './tests/scripts/license_headers_check.sh', label: "Copyright notice"
+                catchError(
+                    buildResult: 'SUCCESS',
+                    stageResult: 'UNSTABLE',
+                    message: 'Some files does not contain license'
+                    ) {
+                  checkLicenses(skip: true, junit: true, ext: '.py', licensor: 'Elasticsearch BV')
+                }
               }
             }
           }
--- a/Makefile
+++ b/Makefile
@@ -21,4 +21,11 @@ docs:
 update-json-schema:
 	bash ./tests/scripts/download_json_schema.sh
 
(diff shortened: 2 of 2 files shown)


beniwohli:
@v1v can you give a short explanation on the benefits of using go-licenser instead of the relatively simple bash script?

One benefit of the bash script is that we can use it very easily in a pre-commit hook: https://github.com/elastic/apm-agent-python/blob/a2442b364fa2c8ddeddc465f7ce3964ff2841204/.pre-commit-config.yaml#L21-L29

If we switch to go-licenser, it would be great if that could also be applied to the pre-commit hook somehow to ensure parity between the two checks.

v1v:
That's a fair question. 

The major benefit is to provide the same toolchain for all the agents if possible, I don't say the `go-licenser` is the one to go, but as it's already a tool provided and used, we could potentially extend it for the rest of agents and being able to have the same common toolchains.

At the moment the current implementation doesn't provide all the requirements for the python agent though. That's one of the reasons I raised this PR as a draft. Do you think I should move this conversation to an issue firstly rather than in this PR?

Are there any concerns in your end?

Thanks

beniwohli:
I agree that harmonizing the license check would be great. But if it hugely complicates go-licenser just to be able to handle the Python agent (AFAIK, most other Elastic projects share the same Apache2 license), it might not be worth it. And as I said, having the check work in a [pre-commit](https://pre-commit.com/) hook is pretty much a requirement for me.

v1v:
I'll close this PR for the time being, and once we go the go-licenser in place we can revisit it.

beniwohli:
zube didn't agree with having this closed. Let's try again

PR #529, opened 2019-07-22
implemented top-level labels for metrics

beniwohli:


beniwohli:
@axw good catch, those `defaultdicts` are a leftover of a previous implementation iteration. A normal `dict` should suffice.

PR #530, opened 2019-07-22
use self-signed certs for tests

beniwohli:
this insulates us from issues with the pytest-localserver certificate


beniwohli:
test failure is unrelated (there seem to be issues with newest pymssql version on Python 3.6... again)

PR #535, opened 2019-07-29
Breakdown metrics

beniwohli:
implements elastic/apm-agent-python#479


elasticapm/traces.py
```diff
@@ -55,7 +59,57 @@
...
+    __slots__ = ("obj", "_nesting_level", "_start", "_duration")
+
+    def __init__(self, obj):
+        self.obj = obj
+        self._nesting_level = AtomicNumber(0)
+        self._start = AtomicNumber(0.0)
+        self._duration = AtomicNumber(0.0)
+
+    def start(self, timestamp):
+        if self._nesting_level.inc() == 1:
```
axw: I wouldn't expect a lot of contention on a single transaction or span. I think a threading.Lock is appropriate here, to ensure nesting_level/duration are consistent.

elasticapm/traces.py
```diff
@@ -64,17 +118,42 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
         self.duration = _time_func() - self.start_time
+        if self._breakdown:
+            for (span_type, span_subtype), timer in compat.iteritems(self._span_timers):
+                labels = {
+                    "span.type": span_type,
+                    "span.subtype": span_subtype,
+                    "transaction.name": self.name,
+                    "transaction.type": self.transaction_type,
+                }
+                self._breakdown.timer("span.self_time", reset_on_collect=True, **labels).val = timer.val
```
axw: I think this should be `.update(timer.val)`? Otherwise if you have two transactions in a row, the second one will wipe out the first's timer values right?

elasticapm/traces.py
```diff
@@ -64,17 +118,42 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
+            labels = {"transaction.name": self.name, "transaction.type": self.transaction_type}
+            self._breakdown.counter("transaction.breakdown.count", reset_on_collect=True, **labels).inc()
+            self._breakdown.timer("transaction.duration", reset_on_collect=True, **labels).val = (self.duration, 1)
```
axw: `.update`

elasticapm/traces.py
```diff
@@ -64,17 +118,42 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
+            self._breakdown.timer(
+                "span.self_time",
+                reset_on_collect=True,
+                **{"span.type": "app", "transaction.name": self.name, "transaction.type": self.transaction_type}
+            ).val = (self.duration - self._child_durations.duration, 1)
```
axw: and here
(it's also important within one transaction, as the implicit transaction "app" self_time could end up with the same labels as an explicit span's)

elasticapm/metrics/base_metrics.py
```diff
@@ -255,6 +294,40 @@ def val(self):
...
+            self._val.value = 0
+            self._count.value = 0
+
+    @property
+    def val(self):
+        with self._lock:
+            return self._val.value, self._count.value
+
+    @val.setter
+    def val(self, value):
```
axw: Maybe drop this method? I'm not sure there's a valid use case for it at the moment.

elasticapm/traces.py
```diff
@@ -198,8 +261,12 @@ def to_dict(self):
             result["context"] = self.context
         return result
 
+    def track_span_duration(self, span_type, span_subtype, self_duration):
+        with self._duration_lock:
+            self._span_timers[(span_type, span_subtype)].update(self_duration)
```
axw: perhaps return early if the transaction is already ended?

elasticapm/traces.py
```diff
@@ -64,17 +118,42 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
+        self.tracer = tracer
 
         self.dropped_spans = 0
         self.context = {}
-        self.tags = {}
 
         self.is_sampled = is_sampled
         self._span_counter = 0
+        self._span_timers = defaultdict(Timer)
+        self._duration_lock = threading.Lock()
```
axw: nit: calling this _span_timers_lock would probably make it easier to understand

elasticapm/traces.py
```diff
@@ -64,17 +118,42 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
+            for (span_type, span_subtype), timer in compat.iteritems(self._span_timers):
+                labels = {
+                    "span.type": span_type,
+                    "span.subtype": span_subtype,
+                    "transaction.name": self.name,
+                    "transaction.type": self.transaction_type,
+                }
+                self._breakdown.timer("span.self_time", reset_on_collect=True, **labels).val = timer.val
+            labels = {"transaction.name": self.name, "transaction.type": self.transaction_type}
+            self._breakdown.counter("transaction.breakdown.count", reset_on_collect=True, **labels).inc()
```
axw: just to confirm: spans for non-sampled transactions are included in breakdowns?

elasticapm/metrics/base_metrics.py
```diff
@@ -255,6 +294,40 @@ def val(self):
...
+    def reset(self):
+        self._val = 0
+
+
+class Timer(object):
+    __slots__ = ("name", "_val", "_count", "_lock", "reset_on_collect")
+
+    def __init__(self, name=None, reset_on_collect=False):
+        self.name = name
+        self._val = AtomicNumber()
```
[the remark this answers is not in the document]
beniwohli: "Better safe than sorry"? :D but yeah, it's probably also quite a bit more efficient to use builtin types.

elasticapm/traces.py
beniwohli: ... and that kills `AtomicNumber` :D

elasticapm/traces.py
beniwohli: Great catch! Will add a test for this case.

elasticapm/traces.py
beniwohli: riiight, they are not. Both `transaction.breakdown.count` and the transaction's `span.self_time` shouldn't be tracked for unsampled transactions in this case, right?

elasticapm/traces.py
beniwohli: we currently don't support asynchronous spans, so this shouldn't happen. But I'll add a comment to indicate that this should be checked once we have asynchronous spans.

elasticapm/metrics/base_metrics.py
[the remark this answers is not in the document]
axw: My position regarding locking and concurrency is to always to exactly what's necessary, nothing more or less. Doing less is self explanatory, but doing unnecessary locking makes things (IMO) more confusing.

e.g. in this case if you have a lock for both _val and _count, then you can infer that they should be updated or acquired together -- they're treated as one atomic value. If you then start locking them individually as well, I'm wondering... are they a single atomic value or not? In what circumstances would you lock and update one and not the other?

elasticapm/traces.py
axw: Correct. `transaction.duration` should still be updated for all, though.

tests/metrics/breakdown_tests.py
```diff
@@ -0,0 +1,198 @@
...
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import time
+
+import mock
+import pytest
+
+import elasticapm
+from elasticapm.utils import compat
+
+
```
axw: can you please also add a test that illustrates that an explicitly created span with type "app" will have its breakdown metrics merged with those of the transaction's? e.g. the third acceptance test in the doc

tests/metrics/breakdown_tests.py
```diff
@@ -0,0 +1,198 @@
...
+            elif elem["span"] == {"type": "template", "subtype": "django"}:
+                assert elem["samples"]["span.self_time.count"]["value"] == 1
+                assert 5000 < elem["samples"]["span.self_time.sum.us"]["value"] < 10000
+                assert elem["transaction"] == {"name": "test", "type": "request"}
+                asserts += 1
+    assert asserts == 4
+
+
+@pytest.mark.parametrize("elasticapm_client", [{"breakdown_metrics": False}], indirect=True)
+def test_disable_breakdowns(elasticapm_client):
```
axw: This test should be updated once we report `transaction.duration` even with the feature disabled.

elasticapm/traces.py
```diff
@@ -64,17 +122,43 @@ def __init__(self, tracer, transaction_type="custom", trace_parent=None, is_samp
...
+            for (span_type, span_subtype), timer in compat.iteritems(self._span_timers):
+                labels = {
+                    "span.type": span_type,
+                    "span.subtype": span_subtype,
+                    "transaction.name": self.name,
+                    "transaction.type": self.transaction_type,
+                }
+                self._breakdown.timer("span.self_time", reset_on_collect=True, **labels).update(*timer.val)
+            labels = {"transaction.name": self.name, "transaction.type": self.transaction_type}
+            self._breakdown.timer("transaction.duration", reset_on_collect=True, **labels).update(self.duration)
```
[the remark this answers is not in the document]
beniwohli: ugh, yeah... that kinda kills how I implemented the config option :/ now wondering if it would make sense to put `transaction.duration` into its own metricset. It's not really related to breakdowns anyway.

tests/metrics/breakdown_tests.py
```diff
@@ -0,0 +1,198 @@
...
+import mock
+import pytest
+
+import elasticapm
+from elasticapm.utils import compat
+
+
+def test_bare_transaction(elasticapm_client):
+    elasticapm_client.begin_transaction("request")
+    time.sleep(0.005)
```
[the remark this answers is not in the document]
beniwohli: Makes sense, but if you're ok with it, I'll do that in a separate PR after the 5.0 release

elasticapm/traces.py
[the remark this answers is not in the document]
axw: yeah, I think a separate metricset is probably the way to go

tests/metrics/breakdown_tests.py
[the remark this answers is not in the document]
axw: No worries.

elasticapm/traces.py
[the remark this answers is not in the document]
beniwohli: done in 8d3ce7050. Do you mean `before_yield`? I moved that into a common base class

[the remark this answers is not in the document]
axw: Sorry, yes, before_yield.

tests/metrics/breakdown_tests.py
beniwohli: done in 6e1bccc876

PR #538, opened 2019-07-30
deprecate tags, add labels, and add support for typed labels

### LINKED ISSUE #517: Rename tag to label
See https://github.com/elastic/apm/issues/42

beniwohli:
fixes #517

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -120,7 +120,7 @@ class BaseSpan(object):
                 value = encoding.keyword_field(compat.text_type(value))
             self.labels[TAG_RE.sub("_", compat.text_type(key))] = value
 
-    @deprecated(label)
+    @deprecated("transaction/span.label()")
     def tag(self, **tags):
         """
         This method is deprecated, please use "label()" instead.
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -35,6 +35,7 @@ import re
 import threading
 import time
 import timeit
+import warnings
 from collections import defaultdict
 
 from elasticapm.conf import constants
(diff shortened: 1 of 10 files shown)


elasticapm/traces.py
```diff
@@ -491,7 +516,7 @@ def end_transaction(self, result=None, transaction_name=None):
 
 
 class capture_span(object):
-    __slots__ = ("name", "type", "subtype", "action", "extra", "skip_frames", "leaf", "tags")
+    __slots__ = ("name", "type", "subtype", "action", "extra", "skip_frames", "leaf", "labels")
```
jalvz: don't you need both `tags` and `labels` here?

tests/instrumentation/transactions_store_tests.py
```diff
@@ -217,46 +218,61 @@ def test_get_transaction_clear():
 def test_tag_transaction():
     requests_store = Tracer(lambda: [], lambda: [], lambda *args: None)
     transaction = requests_store.begin_transaction("test")
-    elasticapm.tag(foo="bar")
-    transaction.tag(baz="bazzinga")
+    elasticapm.label(foo="bar")
```
jalvz: I would go as far as duplicating these tests to make sure that the deprecated `tag` methods still work as expected. And while at it, change the method name :grin:

elasticapm/traces.py
```diff
@@ -511,7 +537,15 @@ def __init__(
         self.extra = extra
         self.skip_frames = skip_frames
         self.leaf = leaf
-        self.tags = tags
+        if tags and not labels:
```
jalvz: if a users forgets `s/tags/labels/` in one call site; will they have both `tags` and `labels` here and therefore get `tags` silently dropped?
is this behaviour consistent with other agents?

beniwohli: No, this would only be an issue if the user uses both `tags` and `labels` in the same `capture_span` call, e.g.

    @capture_span(tags={"foo": "bar"}, labels={"bar": "baz"})
    def my_func():
        pass

That seemed like a silly enough case to not bother handling it.

elasticapm/traces.py
beniwohli: no, the `capture_span` object doesn't have a `tags` attribute :)

tests/instrumentation/transactions_store_tests.py
beniwohli: yeah, good point!

elasticapm/traces.py
jalvz: ah you are right ofc!

tests/instrumentation/transactions_store_tests.py
beniwohli: done in 7a8524d9

PR #544, opened 2019-08-05
limit number of distinct labels per metric to 1000

### LINKED ISSUE #540: Limit the max amount of timer metrics
We missed a bullet point from #479 in the implementation (https://github.com/elastic/apm-agent-python/pull/535):

- [ ] [Limit the max amount of timers](https://docs.google.com/document/d/1-_LuC9zhmva0VvLgtI0KcHuLzNztPHbcM0ZdlcPUl64#heading=h.tzyh7loxf2is)

beniwohli:
fixes #540

### DIFF
--- a/elasticapm/metrics/base_metrics.py
+++ b/elasticapm/metrics/base_metrics.py
@@ -112,9 +112,9 @@ class MetricsRegistry(object):
 class MetricsSet(object):
     def __init__(self, registry):
         self._lock = threading.Lock()
-        self._counters = defaultdict(dict)
-        self._gauges = defaultdict(dict)
-        self._timers = defaultdict(dict)
+        self._counters = {}
+        self._gauges = {}
+        self._timers = {}
         self._registry = registry
         self._label_limit_logged = False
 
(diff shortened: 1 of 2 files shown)


elasticapm/metrics/base_metrics.py
```diff
@@ -164,6 +167,15 @@ def _metric(self, container, metric_class, name, reset_on_collect, labels):
                     pattern.match(name) for pattern in self._registry._ignore_patterns
                 ):
                     metric = noop_metric
+                elif len(container) >= DISTINCT_LABEL_LIMIT:
```
axw: This will allow 1000 counters, 1000 gauges, and 1000 timers. While it means we won't exhaust memory, the logged message is different in meaning. Maybe keep a separate counter?

beniwohli: The log message is confusing anyway tbh (why bring up transaction names in a place that has nothing to do with transaction names, just because one of the possible causes for exhausting the limit could be a large cardinality of transaction names?)

I'll do `if len(self._gauges) + len(self._counters) + len(self._timers) >= ...`

felixbarny: > why bring up transaction names in a place that has nothing to do with transaction names, just because one of the possible causes for exhausting the limit could be a large cardinality of transaction names?

I suspect this to be the most common source of exhausting the limit, especially if custom metrics are not supported. I think it's valuable to include some indication where the problem might be, even if that's only for 80% of the cases. Arguably, the wording could be improved to better reflect that this is only a potential source of the error.

tests/metrics/base_tests.py
```diff
@@ -141,3 +141,20 @@ def test_client_doesnt_start_collector_thread_in_master_process(is_master_proces
...
     assert mock_start_collect_timer.call_count == before + 1
     client.close()
+
+
+@mock.patch("elasticapm.metrics.base_metrics.DISTINCT_LABEL_LIMIT", 2)
+def test_metric_limit(caplog):
+    m = MetricsSet(MetricsRegistry(0, lambda x: None))
+    with caplog.at_level(logging.WARNING, logger="elasticapm.metrics"):
+        for i in range(4):
+            metric = m.timer("x", some_label=i)
```
axw: can you please update the test to show that the other metric types will also be affected?

PR #549, opened 2019-08-05
added support for global labels

### LINKED ISSUE #459: Add support for global labels
See https://github.com/elastic/apm-agent-go/issues/507

beniwohli:
fixes #459

### DIFF
--- a/elasticapm/conf/__init__.py
+++ b/elasticapm/conf/__init__.py
@@ -101,6 +101,22 @@ class _ListConfigValue(_ConfigValue):
         instance._values[self.dict_key] = value
 
 
+class _DictConfigValue(_ConfigValue):
+    def __init__(self, dict_key, item_separator=",", keyval_separator="=", **kwargs):
+        self.item_separator = item_separator
+        self.keyval_separator = keyval_separator
+        super(_DictConfigValue, self).__init__(dict_key, **kwargs)
+
+    def __set__(self, instance, value):
+        if isinstance(value, compat.string_types):
+            items = (item.split(self.keyval_separator) for item in value.split(self.item_separator))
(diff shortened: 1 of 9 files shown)


elasticapm/conf/__init__.py
```diff
@@ -101,6 +101,22 @@ def __set__(self, instance, value):
...
+class _DictConfigValue(_ConfigValue):
+    def __init__(self, dict_key, item_separator=",", keyval_separator="=", **kwargs):
+        self.item_separator = item_separator
+        self.keyval_separator = keyval_separator
+        super(_DictConfigValue, self).__init__(dict_key, **kwargs)
+
+    def __set__(self, instance, value):
+        if isinstance(value, compat.string_types):
+            items = (item.split(self.keyval_separator) for item in value.split(self.item_separator))
+            value = {key: self.type(val) for key, val in items}
```
axw: I think we should strip key and val. Someone's bound to have white space in their environment variables.

elasticapm/conf/__init__.py
```diff
@@ -101,6 +101,22 @@ def __set__(self, instance, value):
...
+        elif not isinstance(value, dict):
+            # TODO: better error handling
+            value = None
+        instance._values[self.dict_key] = value
```
axw: should probably check that the tag keys match `TAG_RE` here? also, truncate value to 1024 chars?

elasticapm/conf/__init__.py
beniwohli: done in 06b64dc9d3ec

elasticapm/conf/__init__.py
beniwohli: Both done in 06b64dc9d3ec

PR #550, opened 2019-08-06
added duration argument to some tracing APIs

### LINKED ISSUE #537: Add `duration` argument to `end_transaction` and `capture_span`
This should help with testing, where we currently rely on `time.sleep(x)` and fuzzy duration comparisons (which tend to fail on CI).

beniwohli:
fixes #537


tests/metrics/breakdown_tests.py
```diff
@@ -84,12 +81,12 @@ def test_single_span(elasticapm_client):
         elif "span.self_time.sum.us" in elem["samples"]:
             if elem["span"] == {"type": "app", "subtype": ""}:
                 assert elem["transaction"] == {"name": "test", "type": "request"}
-                assert 10000 < elem["samples"]["span.self_time.sum.us"]["value"]
+                assert elem["samples"]["span.self_time.sum.us"]["value"] == 9999
```
axw: Where is this 9999 coming from? 

I think there's still a bit missing, which I failed to mention in my original suggestion: the ability to also specify the start time.

beniwohli: added in 47cbeee

PR #551, opened 2019-08-06
respect the NO_PROXY environment variable

### LINKED ISSUE #458: NO_PROXY not passed
The environment variable NO_PROXY is not processed. An APM server in front of a proxy is therefore not found, after a while you get http connection error. Only the environment variables HTTP_PROXY and HTTPS_PROXY are processed.  I`m using flask and the environment variables HTTP & HTTPS_PROXY are needed for oauth.

**To Reproduce**

1. Define APM Server in front of a Proxy
2. Define HTTP_PROXY and HTTPS_PROXY environment variable
3. Define NO_PROXY
4. Try to initiate app
5. Seconds after initate you get an HTTP Connection error.

**Expected behavior**: 
Check if IP or URL is defined in NO_PROXY, maybe part of an URL for wildcard usage.

Change this [line](https://github.com/elastic/apm-agent-python/blob/b3b0ba42c260935b1d3c3a1db244c1ebc14d40ad/elasticapm/transport/http.py#L62) and add a second condition.

- OS: [e.g. Linux]
- Python version: 3.6.8
- Framework and version [e.g. Django 2.1]: Flask 1.0.2
- APM Server version: 7.0.0
- Agent version: 4.2.2

beniwohli:
also, use stdlib functions to get the proxy url

fixes #458

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -711,7 +711,7 @@ Some environment variables that are not specific to the APM agent can be used to
 Using `HTTP_PROXY` and `HTTPS_PROXY`, the agent can be instructed to use a proxy to connect to the APM Server.
 If both are set, `HTTPS_PROXY` takes precedence.
 
-NOTE: the environment variables are case-insensitive.
+NOTE: The environment variables are case-insensitive.
 
 [float]
 [[config-generic-no-proxy]]
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -111,7 +111,7 @@ The URL must be fully qualified, including protocol (`http` or `https`) and port
 |============
 
 
-The transport class to use when sending events to the APM server.
+The transport class to use when sending events to the APM Server.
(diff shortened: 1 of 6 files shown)


docs/configuration.asciidoc
```diff
@@ -698,6 +698,32 @@ If you want to use the route instead of the view name as the transaction name, y
...
+Some environment variables that are not specific to the APM agent can be used to configure the agent.
+
+[float]
+[[config-generic-http-proxy]]
+==== `HTTP_PROXY` and `HTTPS_PROXY`
+
+Using `HTTP_PROXY` and `HTTPS_PROXY`, the agent can be instructed to use a proxy to connect to the APM Server.
+If both are set, `HTTPS_PROXY` takes precedence.
+
+NOTE: the environment variables are case-insensitive.
```
bmorelli25:
```suggestion
NOTE: The environment variables are case-insensitive.
```

PR #563, opened 2019-08-19
Handle empty response from APM Server's config endpoint more gracefully

### LINKED ISSUE #562: Exception json.decoder.JSONDecodeError with empty payload
**Describe the bug**:
It seems that ElasticAPM causes an exception if the payload in the body is empty:
```
Exception in thread eapm conf updater:
  File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/home/hidio/git/api/env/lib/python3.7/site-packages/elasticapm/utils/threading.py", line 79, in run
    rval = self._function(*self._args, **self._kwargs)
  File "/home/hidio/git/api/env/lib/python3.7/site-packages/elasticapm/conf/__init__.py", line 402, in update_config
    new_version, new_config, next_run = transport.get_config(agent.config.config_version, keys)
  File "/home/hidio/git/api/env/lib/python3.7/site-packages/elasticapm/transport/http.py", line 135, in get_config
    return response.headers.get("Etag"), json_encoder.loads(body.decode("utf-8")), max_age
  File "/home/hidio/git/api/env/lib/python3.7/site-packages/elasticapm/utils/json_encoder.py", line 67, in loads
    return json.loads(value, object_hook=better_decoder)
  File "/usr/local/lib/python3.7/json/__init__.py", line 361, in loads
    return cls(**kw).decode(s)
  File "/usr/local/lib/python3.7/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/local/lib/python3.7/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

**To Reproduce**

1. Not sure where the payload actually comes from but if you make it empty (b'') this will cause the issue.

**Expected behavior**:
Do not run json_encoder.loads() on an empty body and continue successfully without raising an exception.

- OS: CentOS 7
- Python version: 3.7.4
- Framework and version: Flask
- APM Server version: 7.3.0
- Agent version: 5.1.0

**Additional context**
An easy fix would be to check if the body is empty and either skip it completely, or continue parsing a valid JSON string instead:
```
return response.headers.get("Etag"), json_encoder.loads(body.decode("utf-8") or "{}"), max_age
```

beniwohli:
also, add a bunch of tests that went forgotten in the original PR

fixes #562

### DIFF
--- a/elasticapm/transport/http.py
+++ b/elasticapm/transport/http.py
@@ -117,7 +117,8 @@ class Transport(HTTPTransportBase):
                         "environment": "bar"
                     }
                 }
-        :return: a three-tuple of new version, config dictionary and validity in seconds, or None
+        :return: a three-tuple of new version, config dictionary and validity in seconds.
+                 Any element of the tuple can be None.
         """
         url = self._config_url
         data = json_encoder.dumps(keys).encode("utf-8")
--- a/elasticapm/transport/http.py
+++ b/elasticapm/transport/http.py
@@ -106,6 +106,19 @@ class Transport(HTTPTransportBase):
                 response.close()
 
     def get_config(self, current_version=None, keys=None):
+        """
+        Gets configuration from a remote APM Server
+
+        :param current_version: version of the current configuration
+        :param keys: a JSON-serializable dict to identify this instance, e.g.
+                {
+                    "service": {
+                        "name": "foo",
+                        "environment": "bar"
+                    }
+                }
+        :return: a three-tuple of new version, config dictionary and validity in seconds, or None
+        """
         url = self._config_url
         data = json_encoder.dumps(keys).encode("utf-8")
         headers = self._headers.copy()
@@ -124,13 +137,17 @@ class Transport(HTTPTransportBase):
             try:
                 max_age = int(next(re.finditer(r"max-age=(\d+)", response.headers["Cache-Control"])).groups()[0])
             except StopIteration:
-                logger.debug("Could not parse Cache-Control header: %s", response["Cache-Control"])
+                logger.debug("Could not parse Cache-Control header: %s", response.headers["Cache-Control"])
         if response.status == 304:
             # config is unchanged, return
             logger.debug("Configuration unchanged")
             return current_version, None, max_age
         elif response.status >= 400:
             return None, None, max_age
+
+        if not body:
+            logger.debug("APM Server answered with empty body and status code %s", response.status)
+            return current_version, None, max_age
 
         return response.headers.get("Etag"), json_encoder.loads(body.decode("utf-8")), max_age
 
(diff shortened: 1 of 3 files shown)


elasticapm/transport/http.py
```diff
@@ -106,6 +106,19 @@ def send(self, data):
...
+
+        :param current_version: version of the current configuration
+        :param keys: a JSON-serializable dict to identify this instance, e.g.
+                {
+                    "service": {
+                        "name": "foo",
+                        "environment": "bar"
+                    }
+                }
+        :return: a three-tuple of new version, config dictionary and validity in seconds, or None
```
beniwohli: this is actually wrong, we *always* return a three-tuple, the elements of which can be `None`

elasticapm/transport/http.py
```diff
@@ -124,14 +137,18 @@ def get_config(self, current_version=None, keys=None):
...
         if response.status == 304:
             # config is unchanged, return
             logger.debug("Configuration unchanged")
             return current_version, None, max_age
         elif response.status >= 400:
             return None, None, max_age
 
+        if not body:
+            logger.debug("APM Server answered with empty body and status code %s", response.status)
+            return current_version, None, max_age
```
jalvz: I don't see in this PR what the caller(s) of this function do with the body, if they handle `None` properly then this looks good to me (or: if this would return `{}` instead maybe the caller logic would be simpler?)

I am also not sure if you should return `current_version` or `response.headers.get("Etag")`... I think what you did makes sense, but I don't understand yet how this scenario can happen in apm-server :/

tests/transports/test_urllib3.py
```diff
@@ -215,8 +216,91 @@ def test_ssl_cert_pinning_fails(waiting_httpsserver):
...
+
+def test_get_config_cache_headers_304(waiting_httpserver, caplog):
+    waiting_httpserver.serve_content(code=304, content=b"", headers={"Cache-Control": "max-age=5"})
+    url = waiting_httpserver.url
+    transport = Transport(url + "/" + constants.EVENTS_API_PATH)
+    try:
+        with caplog.at_level("DEBUG", "elasticapm.transport.http"):
+            version, data, max_age = transport.get_config("1", {})
+        assert waiting_httpserver.requests[0].headers["If-None-Match"] == "1"
+        assert version == "1"
```
jalvz: great tests :100:

elasticapm/transport/http.py
beniwohli: Yes, the caller understands `None`.

As for version, I chose `current_version` because a new version with empty payload does not make much sense IMO, and I don't think we'd handle that properly (we do handle the config going away completely, but then it doesn't have a version either)

PR #579, opened 2019-08-30
Execution context - correctly check if gevent has patched threading.local

ramshaw888:
Typo in call to `is_object_patched`. Correct module to check is `threading`, not `_threading` - https://docs.python.org/3/library/threading.html#threading.local

### DIFF
--- a/.ci/.jenkins_exclude.yml
+++ b/.ci/.jenkins_exclude.yml
@@ -167,3 +167,6 @@ exclude:
     FRAMEWORK: zerorpc-0.4
   - PYTHON_VERSION: python-3.8-rc
     FRAMEWORK: zerorpc-0.4
+  # gevent
+  - PYTHON_VERSION: pypy-2  # see https://github.com/gevent/gevent/issues/1380, fix will be in gevent 1.5
+    FRAMEWORK: gevent-newest
--- a/.ci/.jenkins_framework.yml
+++ b/.ci/.jenkins_framework.yml
@@ -28,4 +28,4 @@ FRAMEWORK:
   - cassandra-newest
   - psutil-newest
   - eventlet-newest
-
+  - gevent-newest
--- a/.ci/.jenkins_framework_full.yml
+++ b/.ci/.jenkins_framework_full.yml
@@ -59,3 +59,4 @@ FRAMEWORK:
   - elasticsearch-5
   - elasticsearch-6
   - elasticsearch-7
+  - gevent-newest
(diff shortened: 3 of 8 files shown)


beniwohli:
@ramshaw888 good catch, this looks indeed like a bug

    In [1]: from gevent.monkey import is_object_patched                                                                                                                        
    
    In [2]: is_object_patched("threading", "local")                                                                                                                            
    Out[2]: False
    
    In [3]: is_object_patched("_threading", "local")                                                                                                                           
    Out[3]: False
    
    In [4]: from gevent import monkey 
       ...: monkey.patch_all()                                                                                                                                                 
    Out[4]: True
    
    In [5]: is_object_patched("threading", "local")                                                                                                                            
    Out[5]: True
    
    In [6]: is_object_patched("_threading", "local")                                                                                                                           
    Out[6]: False

Do you remember why we used `_threading` in the original PR (https://github.com/elastic/apm-agent-python/pull/454)?

beniwohli:
jenkins run tests please

ramshaw888:
@beniwohli It was a typo on my part, looks like I was getting confused the "internal" python `_threading_local` primitive - `_threading` and `_threading.local` do not even exist. https://github.com/python/cpython/blob/master/Lib/threading.py#L1408

We're using the "public api" `threading.local` in `elasticapm.context.threadlocal` so `is_object_patched("threading", "local")` should be the most correct.

beniwohli:
Makes sense! Thanks for the update!

(also, jenkins run the tests please)

beniwohli:
jenkins run the tests please

ramshaw888:
Added more robust tests for gevent and eventlet thread monkeypatching. (relies on #581)

Note that these tests should probably only run in isolation (i.e. when specifying the `gevent` or `eventlet` framework), as the monkeypatching is not easily reversible, e.g. once monkeypatched, it will be so for the remainder of the process. `pytest.mark.gevent` and `pytest.mark.eventlet` applied for that reason.

beniwohli:
You're on a roll :D I think requiring the eventlet/gevent tests to run in isolation is fine (and was already the case for the eventlet tests). There's [pytest-forked](https://github.com/pytest-dev/pytest-forked) which might help if this limitation becomes a burden, although I suspect it would introduce some overhead, and it doesn't work on Windows...

beniwohli:
jenkins run the tests please

beniwohli:
jenkins run the tests please

beniwohli:
@ramshaw888 looks like the failures are caused by the somewhat ancient Logbook version that we have pinned in requirements-base.txt. Let's remove that pinning for now. I'll see if we can introduce matrix testing for logbook itself later on.

ramshaw888:
Ahh thanks, beat me to it :)

beniwohli:
@ramshaw888 hope that's ok with you! We're planning to ship this in a minor release in a few hours, so I thought I'll fix these two testing issues real quick

ramshaw888:
No worries @beniwohli, looks like the checks are all passing now as far as I can tell.

PR #586, opened 2019-09-09
Add logging integrations with logging and structlog

### LINKED ISSUE #520: Integration with log libraries to be able to tag log messages with trace/transaction

basepi:
* Added new entries to the public API as helpers to retrieve transaction, trace, and span IDs
* Added new ways to add APM tracing fields to logs:
  * `logging` filter
  * `structlog` processor
  * automatic log_record_factory for python 3.2+
* Added new "Logging Integrations" documentation
* Added documentation around ingest for log correlation
* Added tests for the filter, processor, and log_record_factory

Resolves elastic/apm-agent-python#520

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -212,7 +212,7 @@ transaction_id = elasticapm.get_transaction_id()
 [[api-get-trace-id]]
 ==== `elasticapm.get_trace_id()`
 
-Get the parent_id of the current transaction's trace. Example:
+Get the `trace_id` of the current transaction's trace. Example:
 
 [source,python]
 ----
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -174,7 +174,6 @@ elasticapm.set_transaction_name('myapp.billing_process')
  * `override`: if `True` (the default), overrides any previously set transaction name.
     If `False`, only sets the name if the transaction name hasn't already been set.
 
-
 [float]
 [[api-set-transaction-result]]
 ==== `elasticapm.set_transaction_result()`
@@ -193,6 +192,35 @@ elasticapm.set_transaction_result('SUCCESS')
  * `result`: (*required*) A string describing the result of the transaction, e.g. `HTTP 2xx` or `SUCCESS`
  * `override`: if `True` (the default), overrides any previously set result.
     If `False`, only sets the result if the result hasn't already been set.
+
+
+[float]
+[[api-get-transaction-id]]
+==== `elasticapm.get_transaction_id()`
+
+Get the id of the current transaction. Example:
+
+[source,python]
+----
+import elasticapm
+
+transaction_id = elasticapm.get_transaction_id()
+----
+
+
+[float]
+[[api-get-transaction-trace-parent-id]]
+==== `elasticapm.get_transaction_trace_parent_id()`
+
+Get the parent_id of the current transaction's trace. Example:
+
+[source,python]
+----
+import elasticapm
+
+trace_parent_id = elasticapm.get_transaction_trace_parent_id()
+----
+
 
 [float]
 [[api-set-custom-context]]
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -218,7 +218,7 @@ Get the parent_id of the current transaction's trace. Example:
 ----
 import elasticapm
 
-trace_parent_id = elasticapm.get_trace_id()
+trace_id = elasticapm.get_trace_id()
 ----
 
 
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -209,8 +209,8 @@ transaction_id = elasticapm.get_transaction_id()
 
 
 [float]
-[[api-get-transaction-trace-parent-id]]
-==== `elasticapm.get_transaction_trace_parent_id()`
+[[api-get-trace-id]]
+==== `elasticapm.get_trace_id()`
 
 Get the parent_id of the current transaction's trace. Example:
 
@@ -218,7 +218,7 @@ Get the parent_id of the current transaction's trace. Example:
 ----
 import elasticapm
 
-trace_parent_id = elasticapm.get_transaction_trace_parent_id()
+trace_parent_id = elasticapm.get_trace_id()
 ----
 
 
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -223,6 +223,20 @@ trace_parent_id = elasticapm.get_trace_id()
 
 
 [float]
+[[api-get-span-id]]
+==== `elasticapm.get_span_id()`
+
+Get the id of the current span. Example:
+
+[source,python]
+----
+import elasticapm
+
+span_id = elasticapm.get_span_id()
+----
+
+
+[float]
 [[api-set-custom-context]]
 ==== `elasticapm.set_custom_context()`
 
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -680,6 +680,18 @@ Any labels set by application via the API will override global labels with the s
 NOTE: This feature requires APM Server >= 7.2.
 
 [float]
+[[config-generic-disable-log-record-factory]]
+==== `disable_log_record_factory`
+|============
+| Environment                              | Django/Flask                 | Default
+| `ELASTIC_APM_DISABLE_LOG_RECORD_FACTORY` | `DISABLE_LOG_RECORD_FACTORY` | `False`
+|============
+
+By default in python 3, the agent will install a <<logging,LogRecord factory>> that
+automatically adds tracing fields to your log records. You can disable this
+behavior by setting this to `True`.
+
+[float]
 [[config-django-specific]]
 === Django-specific configuration
 
@@ -764,11 +776,3 @@ The unit is provided as suffix directly after the number, without and separation
  * `gb` (gigabytes)
 
 NOTE: we use the power-of-two sizing convention, e.g. `1 kilobyte == 1024 bytes`
-
-[float]
-[[config-generic-disable-log-record-factory]]
-==== `DISABLE_LOG_RECORD_FACTORY
-
-By default in python 3, the agent will install a <<logging,LogRecord factory>> that
-automatically adds tracing fields to your log records. You can disable this
-behavior by setting `DISABLE_LOG_RECORD_FACTOR` to `True`.
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -764,3 +764,11 @@ The unit is provided as suffix directly after the number, without and separation
  * `gb` (gigabytes)
 
 NOTE: we use the power-of-two sizing convention, e.g. `1 kilobyte == 1024 bytes`
+
+[float]
+[[config-generic-disable-log-record-factory]]
+==== `DISABLE_LOG_RECORD_FACTORY
+
+By default in python 3, the agent will install a <<logging,LogRecord factory>> that
+automatically adds tracing fields to your log records. You can disable this
+behavior by setting `DISABLE_LOG_RECORD_FACTOR` to `True`.
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -1,0 +1,52 @@
+[[logging-integrations]]
+=== Logging Integrations
+
+`elasticapm` provides integrations with both the default python logging library,
+as well as http://www.structlog.org/en/stable/[`structlog`].
+
+[[logging]]
+==== `logging`
+
+We provide a https://docs.python.org/3/library/logging.html#filter-objects[filter]
+which will add two new attributes to any filtered `LogRecord`:
+
+* `elasticapm_transaction_id`
+* `elasticapm_transaction_trace_parent_id`
+
+Note that because https://docs.python.org/3/library/logging.html#filter-objects[filters
+are not propagated to descendent loggers], you should add the filter to each of
+your log handlers, as handlers are propagated, along with their attached filters.
+
+[source,python]
+----
+import logging
+from elasticapm.handlers.logging import LoggingFilter
+
+console = logging.StreamHandler()
+console.addFilter(LoggingFilter())
+# add the handler to the root logger
+logging.getLogger("").addHandler(console)
+----
+
+
+[[structlog]]
+==== `structlog`
+
+We provide a http://www.structlog.org/en/stable/processors.html[processor] for
+http://www.structlog.org/en/stable/[`structlog`] which will add two new keys
+to the event_dict of any processed event:
+
+* `"elasticapm_transaction_id"`
+* `"elasticapm_transaction_trace_parent_id"`
+
+[source,python]
+----
+from structlog import PrintLogger, wrap_logger
+from elasticapm.handlers.logging import structlog_processor
+
+wrapped_logger = PrintLogger()
+logger = wrap_logger(wrapped_logger, processors=[structlog_processor])
+log = logger.new()
+log.msg("some_event")
+----
+
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -79,7 +79,7 @@ identifiers:
 If you're using structured logging, either https://docs.python.org/3/howto/logging-cookbook.html#implementing-structured-logging[with a custom solution]
 or with http://www.structlog.org/en/stable/[structlog] (recommended), then this
 is fairly easy. Throw the http://www.structlog.org/en/stable/api.html#structlog.processors.JSONRenderer[JSONRenderer]
-in, and use {blog-ref}/structured-logging-filebeat[Filebeat]
+in, and use {blog-ref}structured-logging-filebeat[Filebeat]
 to pull these logs into Elasticsearch.
 
 Without structured logging the task gets a little trickier. Here we
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -18,9 +18,9 @@ each LogRecord object:
 This factory also adds these fields to a dictionary attribute,
 `elasticapm_labels`, using the official ECS https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html[tracing fields].
 
-You can disable this automatic behavior by setting
-<<config-generic-disable-log-record-factory,`"DISABLE_LOG_RECORD_FACTORY": True`>>
-in the config.
+You can disable this automatic behavior by using the
+<<config-generic-disable-log-record-factory,`disable_log_record_factory`>> setting
+in your configuration
 
 For Python versions <3.2, we also provide a
 https://docs.python.org/3/library/logging.html#filter-objects[filter] which will
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -18,7 +18,8 @@ each LogRecord object:
 This factory also adds these fields to a dictionary attribute,
 `elasticapm_labels`, using the official ECS {ecs-ref}/ecs-tracing[tracing fields].
 
-You can disable this automatic behavior by setting `DISABLE_LOG_RECORD_FACTORY=True`
+You can disable this automatic behavior by setting
+<<config-generic-disable-log-record-factory,`"DISABLE_LOG_RECORD_FACTORY": True`>>
 in the config.
 
 For Python versions <3.2, we also provide a
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -8,10 +8,11 @@ as well as http://www.structlog.org/en/stable/[`structlog`].
 ==== `logging`
 
 We provide a https://docs.python.org/3/library/logging.html#filter-objects[filter]
-which will add two new attributes to any filtered `LogRecord`:
+which will add three new attributes to any filtered `LogRecord`:
 
 * `elasticapm_transaction_id`
-* `elasticapm_transaction_trace_parent_id`
+* `elasticapm_trace_id`
+* `ealsticapm_span_id`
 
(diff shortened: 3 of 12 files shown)


docs/logging.asciidoc
```diff
@@ -0,0 +1,52 @@
...
+`elasticapm` provides integrations with both the default python logging library,
+as well as http://www.structlog.org/en/stable/[`structlog`].
+
+[[logging]]
+==== `logging`
+
+We provide a https://docs.python.org/3/library/logging.html#filter-objects[filter]
+which will add two new attributes to any filtered `LogRecord`:
+
+* `elasticapm_transaction_id`
```
axw: We should be using these fields: https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html
There's also another field that's not part of ECS: `span.id`.

It should go like this:

- if there's an active span, then yield `span.id` and `trace.id`
- else if there's an active transaction, then yield `transaction.id` and `trace.id`

I think it would be helpful to provide a function to obtain the current trace IDs, so users don't have to implement this logic. This discussion over in the Node.js agent is relevant: https://github.com/elastic/apm-agent-nodejs/pull/1335#discussion_r322518554

elasticapm/traces.py
```diff
@@ -642,6 +642,26 @@ def set_transaction_result(result, override=True):
...
+    """
+    Returns the current transaction ID
+    """
+    transaction = execution_context.get_transaction()
+    if not transaction:
+        return
+    return transaction.id
+
+
+def get_transaction_trace_parent_id():
```
axw: I think this could just be called `get_trace_id`? Transaction is implied, since you can't have a trace without a transaction; and it's still useful to get the trace ID when you're in the context of a span.

docs/logging.asciidoc
beniwohli: @axw side note on the naming: these are attributes on an object, so we can't use `.` in the name. The idea is to then use these attributes further down the line to include them in your log output, something like `trace.id=%(elasticapm_trace_id)s`

elasticapm/handlers/logging.py
```diff
@@ -155,3 +156,51 @@ def _emit(self, record, **kwargs):
...
+        Unused (logger instance in structlog)
+    :param __:
+        Unused (wrapped method_name)
+    :param event_dict:
+        Event dictionary for the event we're processing
+    :return:
+        `event_dict`, with two new entries.
+    """
+    transaction = execution_context.get_transaction()
+    event_dict["elasticapm_transaction_id"] = transaction.id
```
beniwohli: here we should probably use the names suggested by @axw, although I'm not sure if the right thing happens if we use dotted names (as in, does the value end up in the correct field in ES if we configure structlog to output JSON and pipe it into ES via filebeat or somesuch).

elasticapm/handlers/logging.py
```diff
@@ -155,3 +156,51 @@ def _emit(self, record, **kwargs):
...
+
+
+def structlog_processor(_, __, event_dict):
+    """
+    Add two new entries to the event_dict for any processed events:
+
+    * elasticapm_transaction_id
+    * elasticapm_transaction_trace_parent_id
+
+    :param _:
```
beniwohli: I'd still use the argument names as they seem to be convention in [structlog](http://www.structlog.org/en/0.4.0/processors.html), even if they are unused.

elasticapm/traces.py
basepi: Yeah, most of this naming is trying to be true to the underlying objects. I noticed traces don't have an ID directly, it's the `trace_parent.id`, thus the name. I'll rename them if `trace_parent.id` is considered the trace ID.

docs/logging.asciidoc
basepi: I did provide public API entries for both of these IDs. I didn't spot `span.id` but I'll add it as well, and simplify the names. It will have to be `trace_id` and `transaction_id` as @beniwohli mentioned, and I'll still prepend `elasticapm_` because we're silently adding attributes and should namespace them as such.

elasticapm/handlers/logging.py
basepi: I skipped them only because in `structlog`'s examples, they use these same names when the underlying processor doesn't care about them. I can name them as what they are, I'm not picky here.

basepi: Sorry, should have left a link: http://www.structlog.org/en/stable/examples.html#processors

Note that their sample processors use the same convention.

beniwohli: Oh, didn't see that, interesting!

I've just never seen this convention before, using underscores as argument names (I've seen it for variable names in multiple assignments, something like, `val, _ = some_tuple`), and it's something I'd do a double take every time I happen to read it. You alleviate that somewhat by documenting what the arguments are in the docstring, but still.

docs/logging.asciidoc
basepi: Done.

elasticapm/traces.py
basepi: Done.

elasticapm/handlers/logging.py
basepi: Still haven't tested, but as far as I can tell [dots in field names are turned into properties in ES](https://www.elastic.co/guide/en/elasticsearch/reference/2.4/dots-in-names.html). So this should be good.

elasticapm/handlers/logging.py
basepi: I just named them normally. I wasn't invested in either outcome. :)

basepi:
I still want to add a LogRecord factory for Python 3's [setLogRecordFactory](https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory) -- it will be enabled by default but you can turn it off via a config value.

I also still need to do some more documentation work around json logging and the like, for getting these new fields into ES in a useful way.

Otherwise I've implemented the other review comments, and will do the above ^ tomorrow.

docs/logging.asciidoc
axw: Ahh I see, sorry. Perhaps it would be useful to add a fourth field then, which is a dict with the ECS field names? i.e. something like:

```python
elasticapm_labels = {
    'trace.id': '...',
    'transaction.id': '...',
    'span.id': '...',
}
```

docs/api.asciidoc
```diff
@@ -194,6 +193,49 @@ elasticapm.set_transaction_result('SUCCESS')
...
+[[api-get-trace-id]]
+==== `elasticapm.get_trace_id()`
+
+Get the parent_id of the current transaction's trace. Example:
+
+[source,python]
+----
+import elasticapm
+
+trace_parent_id = elasticapm.get_trace_id()
```
beniwohli: nitpicking here, but let's call the variable `trace_id` as well :)

elasticapm/handlers/logging.py
```diff
@@ -155,3 +156,55 @@ def _emit(self, record, **kwargs):
...
+            transaction.trace_parent.trace_id if transaction and transaction.trace_parent else None
+        )
+        span = execution_context.get_span()
+        record.elasticapm_span_id = span.id if span else None
+        return True
+
+
+def structlog_processor(logger, method_name, event_dict):
+    """
+    Add two new entries to the event_dict for any processed events:
```
beniwohli:
```suggestion
    Add three new entries to the event_dict for any processed events:
```

elasticapm/handlers/logging.py
```diff
@@ -155,3 +156,55 @@ def _emit(self, record, **kwargs):
...
+        record.elasticapm_transaction_id = transaction.id if transaction else None
+        record.elasticapm_trace_id = (
+            transaction.trace_parent.trace_id if transaction and transaction.trace_parent else None
+        )
+        span = execution_context.get_span()
+        record.elasticapm_span_id = span.id if span else None
+        return True
+
+
+def structlog_processor(logger, method_name, event_dict):
```
beniwohli: I wonder if we should create a `elasticapm.contrib.structlog` module and move the processor there, since this isn't really related to the `logging` stdlib module.

tests/handlers/logging/logging_tests.py
```diff
@@ -245,3 +247,47 @@ def test_arbitrary_object(logger):
...
     assert event["log"]["param_message"] == "['a', 'list', 'of', 'strings']"
+
+
+def test_logging_filter_no_span():
+    requests_store = Tracer(lambda: [], lambda: [], lambda *args: None, Config(), None)
+    requests_store.begin_transaction("test")
+    f = LoggingFilter()
+    record = logging.LogRecord(__name__, logging.DEBUG, __file__, 252, "dummy_msg", [], None)
+    f.filter(record)
+    assert record.elasticapm_transaction_id
```
beniwohli: `requests_store.begin_transaction("test")` returns the `Transaction` object, so you could assert that the IDs are identical instead of "just" their existence. Same in the other tests below

docs/logging.asciidoc
basepi: My only concern with that format is how difficult it will make using those attributes in a [logging Formatter](https://docs.python.org/3/library/logging.html#formatter-objects). I'll have to do a little investigation.

basepi: Since the formatters just use simple printf-style string formatting, I don't think it's a good idea to nest all of these attributes. So, I did both. That should provide ease of access for users wanting to format the transaction into normal log messages, and consistency for those logging to JSON.

docs/api.asciidoc
basepi: Done.

elasticapm/handlers/logging.py
basepi: Done.

tests/handlers/logging/logging_tests.py
basepi: Done.

elasticapm/handlers/logging.py
basepi: Done.

elasticapm/handlers/logging.py
```diff
@@ -36,9 +36,11 @@
 import traceback
 
 from elasticapm.base import Client
+from elasticapm.traces import execution_context
 from elasticapm.utils import compat
 from elasticapm.utils.encoding import to_unicode
 from elasticapm.utils.stacks import iter_stack_frames
+from utils import wrapt
```
beniwohli:
```suggestion
from elasticapm.utils import wrapt
```

PyCharm tends to import things without the base module, depending how you set it up (I think it has something to do with the working directory, but not sure)

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+
+wrapped_logger = PrintLogger()
+logger = wrap_logger(wrapped_logger, processors=[structlog_processor])
+log = logger.new()
+log.msg("some_event")
+----
+
+
+[[log-correlation]]
+=== Log Correlation in Elasticsearch
```
bmorelli25: I need to go through the docs and fix this elsewhere, but we should use sentence case in the titles.

```suggestion
=== Log correlation in Elasticsearch
```

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
+[[logging-integrations]]
+=== Logging Integrations
```
bmorelli25:
```suggestion
=== Logging integrations
```

bmorelli25:
I just had another thought about this PR. I'm currently working on documentation for the integration between the APM UI and the Logs UI. The idea for those docs was to include the Log correlation instructions in a central location. You can see the draft I have here: https://github.com/elastic/apm-server/pull/2672/files#diff-e5d2431c1637befb5513a06f10499bfb. Right now I haven't put any work into it. It's just copy/pasted from the logs documentation, and probably doesn't make any sense in this context.

I'm wondering what your thoughts would be on the log correlation documentation living there vs in the Agent repos. It looks like for structured logs you use a logging package specific to python, but for unstructured the solution is more flexible and might apply to other agents as well? Please correct me if I'm wrong. Would like @beniwohli's thoughts as well.

beniwohli:
@bmorelli25 I think everything starting from https://github.com/elastic/apm-agent-python/pull/586/files#diff-b9aa18d4f6625af66a2ba37d4263e914R118 could live in a central documentation, given that all other agents can replicate the format that @basepi proposed.

basepi:
Yeah similar documentation could work for other languages. Obviously getting the unstructured log lines to the parseable state will differ based on the language/logging library the user is working with. If we ended up centralizing it, I'd just replace ours with a link to the central docs.

bmorelli25:
Great, thanks. Let's stick with this format for now and I'll update if/when we get to that point.

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+
+[source, json]
+----
+{
+  "description" : "...",
+  "processors": [
+    {
+      "grok": {
+        "field": "message",
+        "patterns": ["%{GREEDYDATA:msg} | elasticapm transaction.id=%{DATA:transaction.id} trace.id=%{DATA:trace.id} span.id=%{DATA:span.id}"]
```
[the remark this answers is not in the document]
basepi: This is why I made sure to use [DATA](https://github.com/elastic/logstash/blob/v1.4.2/patterns/grok-patterns#L14) instead of [WORD](https://github.com/elastic/logstash/blob/v1.4.2/patterns/grok-patterns#L11) or similar, because it's `.*` instead of `.+`. I think it should just resolve to an empty string for `None`. (Note, I always set the attribute, just set it to `None` if there is not transaction/trace/span ID.)

elasticapm/base.py
```diff
@@ -103,6 +103,15 @@ def __init__(self, config=None, **inline):
             config.disable_send = True
         self.config = VersionedConfig(config, version=None)
 
+        # Insert the log_record_factory into the logging library
+        # The LogRecordFactory functionality is only available on python 3.2+
+        if sys.version_info >= (3, 2) and self.config.disable_log_record_factory is False:
```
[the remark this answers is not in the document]
basepi: You're right, `not` would also work fine. I'm in the habit of `is <bool>` ever since I learned how much faster it is than `== <bool>`. But `not` is more readable and just as fast.

(I didn't realize we were 3.4+, good to know!)

elasticapm/base.py
```diff
@@ -103,6 +103,15 @@ def __init__(self, config=None, **inline):
...
+            record_factory = logging.getLogRecordFactory()
+            if not isinstance(record_factory, elasticapm.utils.wrapt.FunctionWrapper):
```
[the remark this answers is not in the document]
basepi: Actually this is pretty solvable. I think I can just add an attribute on the wrapped function object and check for that attribute (as well as the `isinstance`). I did think about the double wrap and figured life would go on, but I think the attribute will add another failsafe.

elasticapm/handlers/logging.py
basepi: Grrr. Should have checked it when I let pycharm import it for me.

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+to decorate the default LogRecordFactory to automatically add new attributes to
+each LogRecord object:
+
+* `elasticapm_transaction_id`
+* `elasticapm_trace_id`
+* `elasticapm_span_id`
+
+This factory also adds these fields to a dictionary attribute,
+`elasticapm_labels`, using the official tracing fields names as documented
+here: https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html
```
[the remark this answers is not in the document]
basepi: This is awesome! So much better than the long links. Thanks!

Done.

docs/logging.asciidoc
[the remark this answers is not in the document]
beniwohli: Ah, gotcha. I'll retract my comment then 😊

Regarding missing attributes, I had the case in mind when a user configures the formatter, but not the filter or factory.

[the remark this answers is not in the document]
basepi: Ah, I see. Yeah, because I always set it, we should be good. Thanks for asking!

elasticapm/base.py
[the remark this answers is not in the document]
basepi: Ended up adding a change to create a throwaway LogRecord. Ideally `Client` will not be instantiated very often (it's supposed to be a singleton) so we'll only ever have to do this once. I think avoiding an extra step for creating every LogRecord is worth the tradeoff.

elasticapm/base.py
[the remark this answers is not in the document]
basepi: Done.

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+
+You can disable this automatic behavior by setting `DISABLE_LOG_RECORD_FACTORY=True`
```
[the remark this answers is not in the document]
basepi: Done.

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+For Python 3.2+, we use https://docs.python.org/3/library/logging.html#logging.setLogRecordFactory[`logging.setLogRecordFactory()`]
+to decorate the default LogRecordFactory to automatically add new attributes to
+each LogRecord object:
+
+* `elasticapm_transaction_id`
+* `elasticapm_trace_id`
+* `elasticapm_span_id`
+
+This factory also adds these fields to a dictionary attribute,
+`elasticapm_labels`, using the official ECS {ecs-ref}/ecs-tracing[tracing fields].
```
basepi: @bmorelli25 this is failing the docs build, due to `current` vs `1.0` (the attribute resolves to 1.0, not current):

https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html

https://www.elastic.co/guide/en/ecs/1.0/ecs-tracing.html

bmorelli25: Ahh good find. This is one of the problems we have with links to/from documentation books that are versioned differently. I've asked on Slack, but I don't see a way to link there with attributes right now. I say go back to the original link and I'll make a note to clean this up once we have a new `current` reference.

basepi: Done.

basepi:
Alright, I've incorporated all feedback and have the tests back to green. This is ready for final approval.

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+identifiers:
+
+* `transaction.id`
+* `trace.id`
+* `span.id`
+
+If you're using structured logging, either https://docs.python.org/3/howto/logging-cookbook.html#implementing-structured-logging[with a custom solution]
+or with http://www.structlog.org/en/stable/[structlog] (recommended), then this
+is fairly easy. Throw the http://www.structlog.org/en/stable/api.html#structlog.processors.JSONRenderer[JSONRenderer]
+in, and use {blog-ref}/structured-logging-filebeat[Filebeat]
```
axw:
```suggestion
in, and use {blog-ref}structured-logging-filebeat[Filebeat]
```

I was copying some of your docs and noticed that the blog link 404s because {blog-ref} has a trailing slash, so this ends up with a double slash.

bmorelli25: Good spot! It looks like it's the only link attribute with a trailing slash 🤦‍♂

docs/api.asciidoc
```diff
@@ -194,6 +193,49 @@ elasticapm.set_transaction_result('SUCCESS')
...
+
+transaction_id = elasticapm.get_transaction_id()
+----
+
+
+[float]
+[[api-get-trace-id]]
+==== `elasticapm.get_trace_id()`
+
+Get the parent_id of the current transaction's trace. Example:
```
beniwohli:
```suggestion
Get the `trace_id` of the current transaction's trace. Example:
```

Pretty sure I commented on this in a previous review, maybe it was lost in a refactor or something

docs/configuration.asciidoc
```diff
@@ -764,3 +764,11 @@ The unit is provided as suffix directly after the number, without and separation
  * `gb` (gigabytes)
 
 NOTE: we use the power-of-two sizing convention, e.g. `1 kilobyte == 1024 bytes`
+
+[float]
```
beniwohli: Can you move this to right before "Generic Environment variables", and then add a table similar to the other "non-generic" config options?

(My previous comment of "append it at the end" didn't consider that we have a bunch of "special" stuff at the end, like the generic environment variables and the configuration format docs, sorry about that)

docs/logging.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+each LogRecord object:
+
+* `elasticapm_transaction_id`
+* `elasticapm_trace_id`
+* `elasticapm_span_id`
+
+This factory also adds these fields to a dictionary attribute,
+`elasticapm_labels`, using the official ECS https://www.elastic.co/guide/en/ecs/current/ecs-tracing.html[tracing fields].
+
+You can disable this automatic behavior by setting
```
beniwohli: How about

    You can disable this automatic behavior by using the
    <<config-generic-disable-log-record-factory,`disable_log_record_factory`>> setting
    in your configuration

The reason for this being that how exactly the configuration looks depends on the usage (environment variable, Django/Flask config, inline config).

docs/api.asciidoc
basepi: Yeah, weird, not sure how I missed this one. I only resolved each conversation after applying the recommendation. Maybe there was just another instance besides the one you pointed out previously.

docs/logging.asciidoc
basepi: Done.

docs/configuration.asciidoc
basepi: Done.

PR #591, opened 2019-09-17
use structlog's get_logger if it is available

beniwohli:
If the user has set up structlog, this will add some additional context
to each log line, like level and timestamp. E.g.

    {"event": "Read traceparent header 00-42d5718b9ac8289e439a688afb17cbae-d6f9a1dad4c90dec-01"}

becomes

    {"event": "Read traceparent header 00-42d5718b9ac8289e439a688afb17cbae-d6f9a1dad4c90dec-01",
    "timestamp": "2019-09-17T15:37:00.049627Z", "logger": "elasticapm.traces", "level": "debug"}

This can be overridden by setting an environment variable
`ELASTIC_APM_DISABLE_STRUCTLOG` to "true".


basepi:
I'm seriously considering flip-flopping on this again (after all my work on the log correlation testing).

On the one hand, off by default is safer. But if they have `structlog` installed it's not like they'll be surprised to see different style log messages from us. And if I'm doing structured logging it would sure be convenient for the apm agent to switch automatically.

Basically, what I'm trying to say is I am fine either way on this, but honestly I'm leaning toward your first implementation, on by default. ¯\_(ツ)_/¯

beniwohli:
@basepi as a compromise, we could check [`structlog.is_configured()`](http://www.structlog.org/en/stable/api.html#structlog.is_configured), but I'm not a terribly huge fan of that approach, because then the behavior becomes dependent of when exactly `elasticapm` is imported first. It would only use structlog if it is configured before `elasticapm` is imported. Or even worse, some of the loggers could use structlog, while others wouldn't, depending on when exactly each module was first imported

basepi:
Yeah that seems like the worst of both worlds. What's the quote? "A good compromise is when both parties are dissatisfied"? ;)

PR #596, opened 2019-09-24
added support for chained exceptions

### LINKED ISSUE #526: Implement support for chained exceptions
Python 3 supports chained exceptions. Its purpose is to handle cases like this:

    try:
        something_that_breaks()
    except BreakingException:
        do_something_else_that_breaks()  # raises KaboomException

Without chained exceptions, only the `KaboomException` is recorded, the exception information of the `BreakingException` is lost. Python 3 stores the original exception in the `__context__` attribute of the exception value.

Python 3 also introduced an explicit way to raise a chained exception:

    try:
        something_that_breaks()
    except BreakingException as exc:
        raise RetryException('Retry later') from exc

In this case, the `BreakingException` is stored as `__cause__`.

An implementation has been [proposed](https://github.com/elastic/apm/issues/40#issuecomment-508619451) by @axw.

beniwohli:
If an exception has a `__context__` set, we use its traceback for the
chained exception. The exception type and value are either taken from
`__cause__` (if set) or `__context__`.

closes #526


basepi:
Ref elastic/apm#40

beniwohli:
hmm, turns out we can't use `raise XYZ from e` even if we gate it with `if compat.PY3`, as it is a `SyntaxError` on Python 2...

basepi:
Hmmm, it doesn't appear to be excluding for py2 based on the failures...

basepi:
Oh, I see #597 now. Good deal.

beniwohli:
now pytest doesn't complain about not knowing the `--ignore-glob` flag anymore, but it doesn't apply it. Guess I haven't gotten the syntax quite right

Makefile
```diff
@@ -8,9 +8,9 @@ flake8:
 	flake8
 
 test:
-	if [[ "$$PYTHON_VERSION" =~ ^(3.5|3.6|nightly|pypy3)$$ ]] ; then \
+	if [[ "$$PYTHON_VERSION" =~ ^(3.5|3.6|3.7|3.8|nightly|pypy3)$$ ]] ; then \
 	py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT); \
-	else py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT) --ignore=tests/asyncio; fi
+	else py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT) --ignore=tests/asyncio --ignore-glob=py3_*.py; fi
```
basepi: Probably needs quotes:

`--ignore-glob='py3_*.py'`

The shell is likely consuming the `*`

beniwohli: Yep, that was one issue :) the other one was that I need to prefix it with `*/` because the file isn't in the root

PR #602, opened 2019-10-04
Fix logging integrations for DroppedSpan spans

basepi:
The logging integrations assumed (erroneously) that all spans would have an ID. This fixes that assumption, so that `DroppedSpan`s don't cause exceptions.

I also fixed up some tests to use the `elasticapm_client` fixture instead of manually creating a Tracer object.

### DIFF
--- a/elasticapm/handlers/logging.py
+++ b/elasticapm/handlers/logging.py
@@ -217,7 +217,7 @@ def _add_attributes_to_log_record(record):
     record.elasticapm_trace_id = trace_id
 
     span = execution_context.get_span()
-    span_id = span.id if span else None
+    span_id = span.id if span and hasattr(span, "id") else None
     record.elasticapm_span_id = span_id
 
     record.elasticapm_labels = {"transaction.id": transaction_id, "trace.id": trace_id, "span.id": span_id}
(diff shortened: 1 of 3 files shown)


beniwohli:
@basepi maybe it would make sense to add an `id` attribute to `DroppedSpan` that is always `None`. That would make it less of a cognitive overhead to always have to remember that there are multiple types of spans. WDYT?

basepi:
Seems reasonable to me. I'll make that change today.

basepi:
@beniwohli Ready for review. Note that I do not include `None` IDs in the structlog integration. I'm a little unsure on that decision -- in most cases `None` would mean there isn't a transaction/span/trace but in this case we do have a span but it doesn't have an ID...

Probably not a terribly important distinction but I wanted to call it out explicitly in case you think I should include the field in this case.

elasticapm/traces.py
```diff
@@ -451,6 +451,7 @@ class DroppedSpan(BaseSpan):
     def __init__(self, parent, leaf=False):
         self.parent = parent
         self.leaf = leaf
+        self.id = None
```
beniwohli: `id` also needs to be added to `__slots__`

basepi: Whoops, that was a silly mistake. Thanks for catching it!

PR #607, opened 2019-10-10
move processors to run on background processing thread

beniwohli:
processing can be somewhat heavey, moving it out of the main thread
should help with overhead.

This is the first step to move source code collection onto the background
thread. The next step will be to turn the current source code collection
into a processor.


beniwohli:
@basepi can you have another look at this one? Would love to hear your opinion specifically on the new processor and how the hardcoded processors work

https://github.com/elastic/apm-agent-python/pull/607/files#diff-45cc225314a432104edab9aeed0b5c2aR234-R252

https://github.com/elastic/apm-agent-python/pull/607/files#diff-a16d47fe7d0f7cec237390038b9cd86bR122-R123

elasticapm/processors.py
```diff
@@ -229,6 +231,27 @@ def sanitize_http_request_body(client, event):
...
+    #
+    # TODO: further optimize by only opening each file once and reading all needed source
+    # TODO: blocks at once.
+    per_file = defaultdict(list)
+    _process_stack_frames(
+        event, lambda frame: per_file[frame["context"][0]].append(frame) if "context" in frame else None
+    )
+    for filename, frames in compat.iteritems(per_file):
+        for frame in frames:
+            pre_context, context_line, post_context = get_lines_from_file(*frame.pop("context"))
```
basepi: This feels a little...inscrutable. Maybe if we renamed `"context"` to something like `"context_markers"` or `"context_metadata"` or something? It's not the actual context, we just use it here to gather the actual context.

Due to the dynamic way that processors are loaded and applied it's very difficult to tie this to its source at https://github.com/elastic/apm-agent-python/pull/607/files#diff-60f53ea78e11d351c1c8a0f15f1fe609R281 -- maybe that will improve for me over time, and I'm not sure how to improve it, but I just wanted to point it out.

elasticapm/base.py
```diff
@@ -119,6 +119,9 @@ def __init__(self, config=None, **inline):
                 new_factory = elastic_logging.log_record_factory(record_factory)
                 logging.setLogRecordFactory(new_factory)
 
+        processors = [import_string(p) for p in self.config.processors] if self.config.processors else []
+        processors.extend(import_string(p) for p in constants.HARDCODED_PROCESSORS)
```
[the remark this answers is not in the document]
beniwohli: Great point, we don't want duplicates in this list. However, using a set loses the ordering, which might be problematic (no processors rely on it currently, but I don't want to preclude that option). The list of processors should be small enough so we can run through it and create a de-duplicated list.

elasticapm/processors.py
beniwohli: You're totally right, this needs a) more comments, and b) less fancy tricks like `get_lines_from_file(*frame.pop("context"))`

beniwohli: Done in a3ec3b6

elasticapm/base.py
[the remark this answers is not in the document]
beniwohli: Done in 0b23d56

elasticapm/processors.py
basepi: This is a big improvement, thank you!

Issue #610, opened 2019-10-14
ZeroDivisionError in cpu_linux.py


ivasic:
**Describe the bug**:

When deploying my Django app on AWS Beanstalk  the APM agent throws `ZeroDivisionError` with the trace: 
``` python
ZeroDivisionError: float division by zero
  File "elasticapm/utils/threading.py", line 79, in run
    rval = self._function(*self._args, **self._kwargs)
  File "elasticapm/metrics/base_metrics.py", line 97, in collect
    for data in metricset.collect():
  File "elasticapm/metrics/base_metrics.py", line 195, in collect
    self.before_collect()
  File "elasticapm/metrics/sets/cpu_linux.py", line 72, in before_collect
    cpu_usage_ratio = delta["cpu_usage"] / delta["cpu_total"]
```

**To Reproduce**

I'm not able to consistently reproduce this behavior but the problem obviously seems that the `cpu_total` read from the stats file is zero. It may be due to the newly launched docker container that has just started up.

One way to reproduce this (although in my local tests it rarely happens):
- Launch a new `python:3.7` container
- Run any django `manage.py` command as `CMD` via Dockerfile 

**Expected behavior**:

No `ZeroDivisionError` exception is thrown

**Environment (please complete the following information)**
- OS: Linux
- Python version: 3.7.4 (happens on 3.7.3 as well)
- Framework and version [e.g. Django 2.1]:  Django 2.2.6
- APM Server version: v7.4.0
- Agent version: 5.2.1

**Additional context**

I managed to fix this by catching this exception and setting the ration to 0 which I think makes sense. 

``` python
try:
    cpu_usage_ratio = delta["cpu_usage"] / delta["cpu_total"]
except ZeroDivisionError:
    cpu_usage_ratio = 0
```

Similar problem happens with `cpu_process_percent` which I solved the same way.
I can open up a pull request (with a corresponding test case in `cpu_linux_test.py`?) if that makes it easier.

Thanks,
Ivan

beniwohli:
Hi @ivasic 

great catch, thanks for opening the issue! If you're up for it, a pull request would be great! The approach you outlined above looks sensible to me.

basepi:
Fixed in #611

PR #611, opened 2019-10-15
Handle exceptions when `cpu_total` is zero

ivasic:
Ref #610 

I wasn't able to run tests from `cpu_linux_test.py` for some reason, if anyone is willing to help with that I can include a simple test as well.


basepi:
Hmm, as far as I can tell there are no non-standard requirements to run those tests. Did you install [the base test requirements](https://github.com/elastic/apm-agent-python/blob/d0f2a392fb52d0e09d92aa7036a0353e819ba2f1/tests/requirements/requirements-base.txt)? If so, what error are you seeing when you try to run those tests?

ivasic:
@basepi no errors were thrown, the tests from `cpu_linux_tests.py` were just skipped but I did some digging and figured out that they only run on Linux (I was running them on macOS). Duh... :)  

Just added a simple test that exposes `ZeroDivisionError` without the previous fix

tests/metrics/cpu_linux_tests.py
```diff
@@ -155,3 +155,27 @@ def test_mem_free_from_memfree_when_memavailable_not_mentioned(tmpdir):
...
+    metricset = CPUMetricSet(
+        MetricsRegistry(0, lambda x: None),
+        sys_stats_file=proc_stat,
+        process_stats_file=proc_stat_self,
+        memory_stats_file=proc_meminfo,
+    )
+    data = next(metricset.collect())
+
+    cpu_total_expected = 0
+    assert data["samples"]["system.cpu.total.norm.pct"]["value"] == cpu_total_expected
```
beniwohli: How about also asserting on `system.process.cpu.total.norm.pct` being `0` here?

ivasic: Makes sense! Thanks for the tip

tests/metrics/cpu_linux_tests.py
```diff
@@ -155,3 +155,28 @@ def test_mem_free_from_memfree_when_memavailable_not_mentioned(tmpdir):
...
+    assert data["samples"]["system.process.cpu.total.norm.pct"]["value"] == cpu_total_expected
```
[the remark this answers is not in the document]
ivasic: Right, let's see if the update helps

basepi:
@ivasic I released 5.2.2 with this fix: https://github.com/elastic/apm-agent-python/releases/tag/v5.2.2

Thank you again for the contribution!

PR #614, opened 2019-10-21
include `rows_affected` count in dbapi2 instrumentations

### LINKED ISSUE #613: Capture `rows_affected` in DB-API2 integrations
See https://github.com/elastic/apm/issues/112

beniwohli:
fixes #613

depends on https://github.com/elastic/apm-server/issues/2802


elasticapm/instrumentation/packages/dbapi2.py
```diff
@@ -226,11 +227,15 @@ def _trace_sql(self, method, sql, params, action=QUERY_ACTION):
...
-        ):
+        ) as span:
             if params is None:
-                return method(sql)
+                result = method(sql)
             else:
-                return method(sql, params)
+                result = method(sql, params)
+            # store "rows affected", but only for insert/update/delete
+            if span and self.rowcount not in (-1, None) and signature[:6] in DML_QUERIES:
```
axw: maybe extract `6` to the top of the file, and compute it from the lengths of the strings in `DML_QUERIES`? we might end up adding additional dialect-specific statement types

beniwohli: yeah, this seemed like a cute little optimization at the time ("oh look, all 3 keywords are 6 characters long, nice!"), but probably not worth the hassle. I'll modify to `signature.startswith(DML_QUERIES)`, which, funny story, according to `timeit` is about 10% faster ¯\\\_(ツ)\_/¯.

But you bring up an interesting point regarding dialects. `DML_QUERIES` should probably live on the `CursorProxy`, so that driver-specific subclasses can override it.

PR #623, opened 2019-10-31
implement stack_trace_limit config option

beniwohli:
with this config option, it is possible to truncate excessively large 
stack traces. The setting applies to both spans and errors.


fxdgear:
ultimately looks good to me. BUT I lack background knowledge to not know better :(

elasticapm/utils/stacks.py
```diff
@@ -163,23 +163,31 @@ def to_dict(dictish):
...
         frame = tb.tb_frame
-        f_locals = getattr(frame, "f_locals", {})
-        if not _getitem_from_frame(f_locals, "__traceback_hide__"):
-            yield frame, getattr(tb, "tb_lineno", None)
+        f_locals = getattr(frame, "f_locals", None)
+        if f_locals is None or not _getitem_from_frame(f_locals, "__traceback_hide__"):
+            frames.append((frame, getattr(tb, "tb_lineno", None)))
         tb = tb.tb_next
+    if max_frames != -1:
+        frames = frames[-max_frames:]
```
axw: Is it worth counting the number of frames in `tb` first, so you can skip and then yield as before? Then you wouldn't need to create a list only to discard some of the entries.

beniwohli: Sorry for the late answer. I'd rather not traverse the linked list of frames twice (once for counting, once for yielding, unless I'm missing something), as attribute lookup in Python can be expensive. Creating the list should be OK, as most tracebacks don't have an unreasonable amount of frames.

PR #631, opened 2019-11-12
fix pod ID for kubernetes when using the systemd cgroup driver

beniwohli:
When the systemd cgroup driver is used, cgroup paths will have hyphens (e.g. in pod UIDs) escaped to underscores. When we're parsing cgroups, we must therefore unescape them.

refs elastic/apm#165

### DIFF
--- a/elasticapm/utils/cgroup.py
+++ b/elasticapm/utils/cgroup.py
@@ -100,6 +100,8 @@ def parse_cgroups(filehandle):
             pod_id = kubepods_match.group(1)
             if not pod_id:
                 pod_id = kubepods_match.group(2)
+                if pod_id:
+                    pod_id = pod_id.replace("_", "-")
             return {"container": {"id": container_id}, "kubernetes": {"pod": {"uid": pod_id}}}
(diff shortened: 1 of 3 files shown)


elasticapm/utils/cgroup.py
```diff
@@ -100,6 +100,6 @@ def parse_cgroups(filehandle):
             pod_id = kubepods_match.group(1)
             if not pod_id:
                 pod_id = kubepods_match.group(2)
```
axw: should only be replacing in this case, when we're matching the systemd pattern (kubepods.slice, etc.)

eyalkoren: That's what I did as well. Not sure if this is a real problem though, meaning- not sure which has more chances of requiring us to reiterate in the future. But for the sake of consistency...

beniwohli: Good call. I was unsure if we can apply the replacement to pod IDs in general or only if systemd is involved. Better to be safe and consistent tho :+1:

PR #636, opened 2019-11-14
initialize event processor thread on demand

beniwohli:

Instead of starting the processor thread at startup,
we do it on demand. Additionally, we do a PID check
on every queue event. If the PID changed, we assume
that the process has been forked, and create new 
background threads.

## Why is it important?

We continue having issues with forked worker processes
not getting processor threads. We tried several approaches
in the past, like listening to uwsgi's postfork event.

This approach should work independent of the environment
(celery, uwsgi, gunicorn, etc.) and the exact forking logic.
It has slightly higher overhead due to the regular PID check.
Theoretically, we could reduce the check to only occur every
N queue events. But let's keep it simple for now.


elasticapm/transport/base.py
```diff
@@ -97,14 +98,16 @@ def __init__(
...
-            compat.postfork(lambda: self._start_event_processor())
+        self._pid = None
+        self._thread_starter_lock = threading.Lock()
 
     def queue(self, event_type, data, flush=False):
+        with self._thread_starter_lock:
+            # check if self._pid is the same as os.getpid(). If they are not the same, it means that our
+            # process was forked at some point, so we need to start another processor thread
+            if self._pid != os.getpid():
+                self._start_event_processor()
```
sundarv85: Since the `_start_event_processor()` already has the check

`if (not self._event_process_thread or not self._event_process_thread.is_alive()) and not self._closed:`

would it not be sufficient just to have it, instead of the pid checking.

Or if pid check is more reliable, in case a dispatcher can fork an already running process, then should this check from the `_start_event_processor` be removed?

sundarv85:
Thanks. We are also facing the same issue and are testing out this fix. One question is:

Since the `_start_event_processor()` already has the check

`if (not self._event_process_thread or not self._event_process_thread.is_alive()) and not self._closed:`

would it not be sufficient just to have it, instead of the pid checking.

Or if pid check is more reliable, in case a dispatcher can fork an already running process (where the event processor is already running), then should this check from the `_start_event_processor` (if event processor is available and running) be removed?

basepi:
@beniwohli What's the status here? I think we should prioritize this solution -- I mostly didn't address multiprocessing scaling concerns in my tornado implementation, assuming that this fix would solve that issue. I'd love to have this in the 5.4 release with all of our new frameworks.

elasticapm/base.py
```diff
@@ -188,15 +194,23 @@ def __init__(self, config=None, **inline):
...
 
+        self.start_threads()
+
+    def start_threads(self):
+        with self._thread_starter_lock:
+            current_pid = os.getpid()
+            if self._pid != current_pid:
+                for manager in self._thread_managers.items():
+                    manager.start_thread()
+                self._pid = current_pid
```
basepi: My brain is a bit fried so I might be wrong, but I think this has potential to introduce a race condition.

Let's say we fork after the `os.getpid()` but before we start all of our threads. This means one or more of our threads will be started twice, once now and once the next time we `start_threads()` because the pid will have changed. Fairly low risk, but since all of our code is designed to execute in a single thread, none of it is multi-thread safe.

Would it be worth attempting to stop the thread before starting it?

basepi:
Also, test failures are likely due to https://github.com/elastic/apm-agent-python/blob/f007f22b44d051d197aa7b54de8fc9e6deec3e4b/tests/config/central_config_tests.py#L34 since `update_config` isn't module-level anymore.

elasticapm/base.py
beniwohli: As far as I understand the way forking is used by gunicorn, celery etc, it doesn't happen at any "random" moment like a thread context switch would. Instead, it happens at a defined place, e.g. after the application is imported.

The problem with prophylactically stopping threads is that at least for the `Transport`, closing it is destructive, and a new transport objects needs to be initiated. That would complicate things a lot AFAICT.

basepi: Perfect. Thanks for the clarification, I know less than I should about their forking models and just wanted to make sure.

PR #638, opened 2019-11-14
Add support for truncating dicts in local vars

### LINKED ISSUE #632: Truncate large dictionaries is local vars
**Summary**
If you have large dicts in local vars, the messages to be sent to the server can be too large for the server to accept.

There should be an option to truncate dicts to a maximum length.

When capturing exceptions using `client.capture_exception()` the stack trace, including local variables, is sent to the server. The size of the message sent to the server depends on the size of the local variables. When the message is larger than a configurable limit, the server rejects it with a 400 error. This can lead to errors being silent for weeks (we learned this the hard way).

When a local variable is a list or a string, it is truncated https://github.com/elastic/apm-agent-python/blob/6113dc2c7683ca64c4677991f7dec91d2fb1caa3/elasticapm/utils/encoding.py#L185-L189
This prevents large lists from being sent and reduces the message size.

When a local var is a dict, it is not truncated. There is a `TODO` comment about it in the `shorten` function above. But that function never gets a dict as input. Instead, dicts are recursively scanned before even calling the `shorten` function.

The recursive scanning seems to happen in the `varmap` function here
https://github.com/elastic/apm-agent-python/blob/6113dc2c7683ca64c4677991f7dec91d2fb1caa3/elasticapm/utils/__init__.py#L63-L64

This leads to huge messages being sent to server when there's an exception in a function with huge dicts.

Ideally, for our use case at least, dicts would be shortened as it happens with lists or strings.

We have achieved this by monkey patching the var map function like this
```python
import elasticapm.utils


# Monkey patch APM to only report the 20 first items of dicts
def first_dict_elements(var, max_length):
    stop = max_length if len(var) <= max_length else max_length - 1
    ret = {}
    length = 0
    for k, v in elasticapm.utils.compat.iteritems(var):
        ret[k] = v
        length += 1
        if length >= stop:
            if len(var) > length:
                ret['...'] = "(%d more elements)" % (len(var) - length,)
            break
    return ret


def varmap(func, var, context=None, name=None):
    """
    Executes ``func(key_name, value)`` on all values,
    recursively discovering dict and list scoped
    values.
    """
    if context is None:
        context = set()
    objid = id(var)
    if objid in context:
        return func(name, "<...>")
    context.add(objid)
    if isinstance(var, dict):
        cropped = first_dict_elements(var, 20)
        ret = dict((k, varmap(func, v, context, k)) for k, v in elasticapm.utils.compat.iteritems(cropped))
    elif isinstance(var, (list, tuple)):
        ret = func(name, [varmap(func, f, context, name) for f in var])
    else:
        ret = func(name, var)
    context.remove(objid)
    return ret


elasticapm.utils.varmap = varmap
```

This truncates all dicts to 20 elements and adds a last element with the number of skipped elements.
Note that there is some ugly length counting needed because the `varmap` function is called twice for each dict so we sometimes process dicts that have already been truncated and getting the number of skipped elements becomes tricky.

Would a solution like this be implementable by the library?

There might be config options available to achieve this that we did not see. Or maybe is possible to provide some pre- or post-processor of the stack trace to achieve the same. We did not investigate that route.

basepi:

* Adds support to the `shorten()` function for shortening dicts
* Adds new config value, `local_var_dict_max_length` for use with the above
* Tweaks `varmap()` to run the mapped func on dict objects
* Tweaks `_sanitize()` processor for safely processing dict objects (since it is used with `varmap()` as well)
* Adds new `shorten()` and `_sanitize()` tests for dicts

## Why is it important?

Details in #632, but the short version is that large dict objects in the local variable list would not be shortened, resulting in large payloads that the apm server would drop. This was especially bad because errors could be silenced inadvertently as a result.

## Related issues
Closes #632


basepi:
@beniwohli I opted for the less-impactful "band-aid" solution rather than rewriting this code or adding the TODO'd API for metadata. This felt fairly critical to me and I think there are higher priorities than this rewrite.

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,10 @@ https://github.com/elastic/apm-agent-python/compare/v5.3.0\...master[Check the d
 // Unreleased changes go here
 // When the next release happens, nest these changes under the "Python Agent version 5.x" heading
 
+===== New Features
```
basepi: @bmorelli25 the docs build is failing: 
```
22:09:16 INFO:build_docs:asciidoctor: WARNING: ../CHANGELOG.asciidoc: line 28: section title out of sequence: expected level 3, got level 4
```
However, I don't want to just make it a level three heading, or else we'll have to modify all the heading levels every time we release. Is there a way to tell it to ignore this issue? Alternatively, we could add a level 3 heading between `Unreleased` and `New Features`, though that seems silly. Thoughts?

bmorelli25: > Is there a way to tell it to ignore this issue

Yes! This needs a float tag: `[float]`. This tells the build to "float" the heading to the current page and not include it in the section hierarchy. That will fix this error.

bmorelli25: As an example: https://github.com/elastic/apm-agent-python/blame/ebe1c4a6ca057cadac15ac6726527428967d1679/CHANGELOG.asciidoc#L13-L14

tests/utils/encoding/tests.py
```diff
@@ -227,6 +227,12 @@ def test_shorten_tuple():
     #     assert result[-1] == '(450 more elements)'
 
 
+def test_shorten_dict():
+    result = shorten({k: v for k in range(500) for v in range(500)}, dict_length=50)
```
paulinus:
```suggestion
    result = shorten({k: k for k in range(500)}, dict_length=50)
```
The double loop produces a dict where all values are `499`

CHANGELOG.asciidoc
```diff
@@ -25,6 +25,10 @@ https://github.com/elastic/apm-agent-python/compare/v5.3.0\...master[Check the d
...
+
+* Added support for shortening dicts in local variables (#638)
```
bmorelli25: Also, changelog links will now need to be formatted like this:

```suggestion
* Added support for shortening dicts in local variables {pull}638[#638 ]
```

elasticapm/utils/encoding.py
```diff
@@ -178,15 +178,41 @@ def to_string(value):
...
         var = var[: string_length - 3] + "..."
     elif isinstance(var, (list, tuple, set, frozenset)) and len(var) > list_length:
         # TODO: we should write a real API for storing some metadata with vars when
         # we get around to doing ref storage
-        # TODO: when we finish the above, we should also implement this for dicts
         var = list(var)[:list_length] + ["...", "(%d more elements)" % (len(var) - list_length,)]
+    elif isinstance(var, dict) and len(var) > dict_length:
+        trimmed_tuples = [(k, v) for (k, v) in compat.iteritems(var)][:dict_length]
+        if "..." not in var:
+            trimmed_tuples += [("...", "(%d more elements)" % (len(var) - dict_length))]
```
paulinus: Since this is adding an extra element to the dict. If ones calls `shorten` twice, this element gets removed.  So, `shorten(long_dict)` will include the `...` key but `shorten(shorten(long_dict))` will not.
In my rudimentary tests, I've seen `varmap` being called twice for dicts. So the `...` key may not be present in the final message sent to the server.
A possible solution is to pop to replace the last element of the array with the dots instead of adding an extra element.
```suggestion
            trimmed_tuples[-1] = ("...", "(%d more elements)" % (len(var) - dict_length + 1))
```

elasticapm/utils/encoding.py
```diff
@@ -178,15 +178,41 @@ def to_string(value):
...
+        var = dict(trimmed_tuples)
```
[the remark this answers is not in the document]
paulinus: The order is not maintained. In the UI the keys are sorted alphabetically (which lets the dots in the first position)

beniwohli:
Soo, funny story regarding insertion order stability and getting our placeholder element shown last in the UI...

First of all, Go maps purposefully randomize order when iterating over a map, to drive the message home that order can't be relied upon. So that's out of the window, and we need not even try to maintain order on our side.

But, turns out that the UI displays dicts in alphabetical key order anyway. Which means `...` would come first in most cases.

We could either

- ask the UI folks to implement a special case when rendering dicts in the local vars component and always render the `...` element last. This would be shipped with 7.6 at the earliest.
- be sneaky and instead of `...` use `…` (the unicode ellipsis character), which loses, alphabetically speaking, against most common ASCII characters. But we're out of luck for any other character sets, e.g. Chinese (see http://s.woh.li/uYyN3iMaDO.png for an example rendering).

Or we could do both. Use `…`, get correct order for many cases now, and ask the UI team to render the `…` element last, so we get correct rendering in all cases with 7.6 (or later, depending on priorities).

elasticapm/utils/encoding.py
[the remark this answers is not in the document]
beniwohli: yup, just realized the same, see comment below :) resolving this, as using `OrderedDict` won't help us.

elasticapm/utils/encoding.py
```diff
@@ -178,15 +178,41 @@ def to_string(value):
...
     var = transform(var)
     if isinstance(var, compat.string_types) and len(var) > string_length:
         var = var[: string_length - 3] + "..."
     elif isinstance(var, (list, tuple, set, frozenset)) and len(var) > list_length:
         # TODO: we should write a real API for storing some metadata with vars when
         # we get around to doing ref storage
-        # TODO: when we finish the above, we should also implement this for dicts
         var = list(var)[:list_length] + ["...", "(%d more elements)" % (len(var) - list_length,)]
+    elif isinstance(var, dict) and len(var) > dict_length:
+        trimmed_tuples = [(k, v) for (k, v) in compat.iteritems(var)][:dict_length]
```
beniwohli: It might be worth not turning the whole dict into a list, only to throw the majority away, especially for very large dicts. Something like this would be less resource intensive

```suggestion
        trimmed_tuples = [(k, v) for (k, v) in itertools.islice(compat.iteritems(var), dict_length)]
```

According to `timeit`, with a dict of 5000 elements and trimming to 20 elements, it's about a 100x difference

basepi: I like it, thanks!

elasticapm/utils/encoding.py
basepi: It's being called twice, but I think once with `_sanitize` and once with `shorten` -- that's why it was important to apply the patch here, instead of in `varmap` like in your monkey-patch on your issue. `varmap` is not a single-use mapping function, unfortunately.

tests/utils/encoding/tests.py
basepi: Whoops, I don't know what I was thinking. To accomplish what I meant to accomplish, I'm actually going to do this:

```
{k: v for k, v in enumerate(range(500))}
```

CHANGELOG.asciidoc
basepi: Perfect. I'll make that change.

basepi:
@beniwohli Alternatively, we could name the new key in a more human-understandable way. I used the `...` because it matched the list, but we could easily name it `<keys_removed>` or something, which would make ordering less of an issue. I don't know how the UI folks feel about those kinds of exceptions.

beniwohli:
I like that. I'll also throw `<truncated>` into the ring, but I'll leave it to you to decide :)

We can still ask the UI team to move that key to the end, they are usually quite open to implement such things if it makes the experience better.

basepi:
I like `<truncated>` and just pushed that change. What's the best way to make that request from UI?

beniwohli:
Probably best to open an issue in https://github.com/elastic/kibana, prefix the label with `[APM]` and label it with `Team:apm`. If you want to go the extra mile, you can update Opbeans to have a dict that gets truncated e.g. here: https://github.com/elastic/opbeans-python/blob/b5aa81b6a5f3007a92e2d842d2f14d2684f0095e/opbeans/tasks.py#L18

Not sure what's up with Codecov reporting weird coverage numbers, but they are so far off reality that we can ignore them I think :D

tests/utils/encoding/tests.py
basepi: @beniwohli pointed out that my version is actually the same as your version, but way less efficient. I blame newborn sleep deprivation! Fixed in https://github.com/elastic/apm-agent-python/pull/638/commits/8827348f8d37302c1c1bc6b5a7104dbe78660b67

basepi:
> ah, looks like coverage statuses get updates as more and more matrix builds finish. That's new I think...

Am I missing something or is that a terrible way to do this? I don't want you to have a big red fail if it's still in-progress...

Anyway, updated opbeans here: https://github.com/elastic/opbeans-python/pull/37

And opened a new Kibana issue here: https://github.com/elastic/kibana/issues/50818

PR #639, opened 2019-11-18
Integration for aiohttp web

beniwohli:

Adds integration for the aiohttp framework. The start of the transaction
is triggered in a middleware, while the end is triggered in a signal
handler for the `on_reponse_prepare` signal sent by aiohttp

## Why is it important?

We need to support some commonly used asyncio frameworks.


elasticapm/contrib/aiohttp/__init__.py
```diff
@@ -0,0 +1,65 @@
...
+@asyncio.coroutine
+def on_response_prepare(request, response):
+    elasticapm_client = request.get(CLIENT_KEY)
+    if elasticapm_client:
+        elasticapm.set_transaction_result("HTTP {}xx".format(response.status // 100), override=False)
+        resource = request.match_info.route.resource
+        if resource:
+            name = "{} {}".format(request.method, resource.canonical)
+            elasticapm.set_transaction_name(name, override=False)
+        elasticapm_client.end_transaction()
```
axw: What's the reason for ending the transaction in `on_response_prepare`, rather than in the middleware? Doesn't `on_response_prepare` fire before the handler returns to the middleware?

beniwohli: Great point, fixed!

beniwohli:
Closing this, will open a new PR that contains instrumentation for both aiohttp client and web

PR #644, opened 2019-11-20
added auto-instrumentation of Django management commands

### LINKED ISSUE #627: Add instrumented Django management command base class
We should make it easy to instrument Django management commands. It is not a lot of work now, but it could be made even easier by providing a base class that calls the right APIs

Currently, the user would need to do something like

    from django.core.management.base import BaseCommand, CommandError
    
    import elasticapm
    from elasticapm.contrib.django.client import get_client
    
    class Command(BaseCommand):
        def handle(self, *args, **options):
            elasticapm.instrument()
            client = get_client()
            client.begin_transaction(transaction_type="command")
            try:
                # do some work
                status = "success"
            except Exception as e:
                client.capture_exception()
                status = "failure"
                raise e  # Python 3
                # for Python 2, use this instead:
                #
                # import sys
                # t, v, tb = sys.exc_info()
                # raise t, v, tb
            finally:
                # use module name as transaction name
                client.end_transaction(name=__name__, result=status)

beniwohli:

It adds an instrumentation module for Django management commands.

Other than all other instrumentation modules, we override
`call_if_sampling`, not `call`. The reason for this is that
`call` excpects an already running transaction, which we don't
have in this case.

To give the user some influence on which commands are instrumented,
a new configuration option `django_commands_exclude` is introduced.

## Why is it important?

Management commands are often used in cron jobs or during deployment.
Capturing this data can be very helpful.

## Related issues

fixes #627


elasticapm/utils/compat.py
```diff
@@ -107,7 +107,26 @@ def iteritems(d, **kwargs):
...
+        exec("""exec _code_ in _globs_, _locs_""")
+
+    exec_(
+        """def reraise(tp, value, tb=None):
+    try:
+        raise tp, value, tb
+    finally:
+        tb = None
+"""
+    )
```
basepi: My brain really dislikes this. Any reason we're not using `six` instead?

docs/configuration.asciidoc
```diff
@@ -750,6 +750,17 @@ To trace Django requests, the agent uses a middleware, `elasticapm.contrib.djang
...
+[[config-django-commands-exclude]]
+==== `django_commands_exclude`
+|============
+| Environment                            | Django                    | Default
+| `ELASTIC_APM_DJANGO_COMMANDS_EXCLUDE`  | `DJANGO_COMMANDS_EXCLUDE` | `runserver*,migrate,createsuperuser,\*shell*,testserver`
+|============
+
+By default, Elastic APM instruments Django management commands.
+You can supply a list of commands that should not be instrumented.
+To disable instrumenting of management commands, set it to `*`.
```
basepi: What would you think about making this a whitelist (with globbing included)? That would also make this much safer. (More thoughts in the review body)

docs/django.asciidoc
```diff
@@ -93,6 +93,15 @@ the agent also collects fine grained metrics on template rendering,
 database queries, HTTP requests, etc.
 You can find more information on what we instrument in the <<automatic-instrumentation, Automatic Instrumentation>> section.
 
+Lastly, the agent will also collect performance data for Django management commands.
+You can disable instrumenting for certain commands using the
```
bmorelli25:
```suggestion
You can disable instrumentation for certain commands using the 
```

docs/django.asciidoc
```diff
@@ -93,6 +93,15 @@ the agent also collects fine grained metrics on template rendering,
 database queries, HTTP requests, etc.
 You can find more information on what we instrument in the <<automatic-instrumentation, Automatic Instrumentation>> section.
 
+Lastly, the agent will also collect performance data for Django management commands.
+You can disable instrumenting for certain commands using the 
+<<config-django-commands-exclude,`django_commands_exclude`>> setting.
+Transactions for management commands can be accessed in the APM app in Kibana by choosing `django_command` in the "transaction type" filter.
```
bmorelli25: **APM app** 😍

docs/django.asciidoc
```diff
@@ -93,6 +93,15 @@ the agent also collects fine grained metrics on template rendering,
...
+
+NOTE: The agent collects command line arguments as additional meta data for transactions.
```
bmorelli25:
```suggestion
NOTE: The agent collects command line arguments as additional metadata for transactions.
```

elasticapm/utils/compat.py
beniwohli: Depending on six is somewhat problematic, as it is a very widely used dependency. Meaning we can't require any specific version of six, as that might clash with requirements of another module that depends on six.

We used to vendor the complete six module, but at some point removed the vast majority of it because we only use a very small subset. The result is the `elasticapm.utils.compat` module.

We can totally discuss if declaring six as a dependency would be the better approach, but if it's ok with you, let's do that in a different issue.

docs/configuration.asciidoc
beniwohli: A whitelist would definitely be the safest method, but it creates a lot of work for the user. Large Django projects can easily have dozens of management commands, and gain new ones if new 3rd party Django apps are added to the project. Keeping the white list up-to-date could be tedious work.

Another draw-back of a whitelist is that we can't provide a blacklist of commands that we should absolutely not instrument in code.

How about adding a second option, which is a flag, and disable it by default? Add some specific documentation in the Django docs, and include the `NOTE` there. That way, people will see the note when they read up on how to enable management command instrumentation.

elasticapm/utils/compat.py
basepi: Honestly this is likely what `six` is doing under the covers anyway, so I'm fine. I assumed there was a reason but was just curious. :)

docs/configuration.asciidoc
basepi: Sounds reasonable to me. 👍

elasticapm/utils/compat.py
beniwohli: This is a straight copy from `six` :) there's a copyright note in the source, so it should be kosher

PR #649, opened 2019-11-24
Add support for ipv6 address format when parsing url [#648]

### LINKED ISSUE #648: Support for ipv6 address format
**Describe the bug**: 

Exception is thrown when python agent is trying to parse url, that contains ipv6 address.
```
ValueError: too many values to unpack (expected 2)
  File "elasticapm/contrib/django/middleware/__init__.py", line 162, in process_response
    "request",
  File "elasticapm/traces.py", line 333, in set_context
    data = data()
  File "elasticapm/contrib/django/middleware/__init__.py", line 160, in <lambda>
    request, capture_body=self.client.config.capture_body in ("all", "transactions")
  File "elasticapm/contrib/django/client.py", line 137, in get_data_from_request
    result["url"] = get_url_dict(url)
  File "elasticapm/utils/__init__.py", line 84, in get_url_dict
    hostname, port = netloc.split(":")
```

**To Reproduce**

Not sure how can I manually replicate it, but below is the request details that is send to our webapp.

```
curl \
 --compressed \
 -H "Accept: */*" \
 -H "Accept-Encoding: gzip, deflate" \
 -H "Connection: close" \
 -H "Host: [::ffff:a9fe:a9fe]:80" \
 -H "User-Agent: AWS Security Scanner" \
 -H "X-Forwarded-For: 44.224.22.196, 10.255.0.2" \
 -H "X-Forwarded-Port: 443" \
 -H "X-Forwarded-Proto: https" \
 -H "X-Real-Ip: 10.255.0.2" \
 "http://[::ffff:a9fe:a9fe]/latest/dynamic/instance-identity/document"
``` 

**Expected behavior**:

The url containing hostname in ipv6 address, should be parsed correctly and python agent should  successfully process request.

- OS: Linux
- Python version: 3.5
- Framework and version [e.g. Django 2.1]: 2.1
- APM Server version: 6.7.1
- Agent version: 5.3.0

mariodev:

The changes are improving `get_url_dict` util function, to correctly parse ipv6 address.

## Why is it important?

Allows to handle request without throwing exception. More details in related issue. 

## Related issues
closes #648

### DIFF
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -103,7 +103,10 @@ def is_master_process():
 
 def get_url_dict(url):
     scheme, netloc, path, params, query, fragment = compat.urlparse.urlparse(url)
-    if ":" in netloc:
+    if '[' in netloc and ']' in netloc:
+        hostname = netloc.split(']')[0][1:]
+        port = (netloc.split(']:') + [None])[1]
+    elif ":" in netloc:
         hostname, port = netloc.split(":")
     else:
         hostname, port = (netloc, None)
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -119,7 +119,6 @@ def get_url_dict(url):
         url_dict["search"] = encoding.keyword_field("?" + parse_result.query)
     return url_dict
 
-
 def sanitize_url(url):
     if "@" not in url:
         return url
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -119,6 +119,7 @@ def get_url_dict(url):
         url_dict["search"] = encoding.keyword_field("?" + parse_result.query)
     return url_dict
 
+
 def sanitize_url(url):
     if "@" not in url:
         return url
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -103,7 +103,7 @@ def is_master_process():
 
 def get_url_dict(url):
     parse_result = compat.urlparse.urlparse(url)
-    
+
     url_dict = {
         "full": encoding.keyword_field(url),
         "protocol": parse_result.scheme + ":",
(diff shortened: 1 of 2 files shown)


elasticapm/utils/__init__.py
```diff
@@ -103,7 +103,10 @@ def is_master_process():
 
 def get_url_dict(url):
     scheme, netloc, path, params, query, fragment = compat.urlparse.urlparse(url)
-    if ":" in netloc:
+    if '[' in netloc and ']' in netloc:
```
beniwohli: I had a quick look at `ParseResult` (the object that `urlparse` returns). It has properties for `hostname` and `port`, which properly support IPv6 URLs since Python 2.7. Only drawback I see is that it'll lowercase IPv6 addresses, but that seems like it isn't a huge problem.

So we would do something like this:

    parse_result = compat.urlparse.urlparse(url)

And then use `parse_result.hostname`, `parse_result.port`, `parse_result.scheme` etc. What do you think?

mariodev: @beniwohli You're right. I refactored the code to use `ParseResult`. Since the port is returned as integer, I had to add small hack to map into string, so not to introduce any regressions. Thanks for the help.

beniwohli:
jenkins run the tests please

elasticapm/utils/__init__.py
```diff
@@ -102,24 +102,23 @@ def is_master_process():
...
+
     if port:
         url_dict["port"] = port
-    if query:
-        url_dict["search"] = encoding.keyword_field("?" + query)
+    if parse_result.query:
+        url_dict["search"] = encoding.keyword_field("?" + parse_result.query)
     return url_dict
 
-
```
beniwohli: Can you add this empty line back as it is required by our code autoformatter black? See https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-linting-mbp/detail/PR-649/5/tests

beniwohli:
@mariodev I shipped this in version 5.3.2 of the agent. Thanks again for reporting and fixing the bug!

PR #650, opened 2019-11-25
update changelog for #649

beniwohli:
also, use proper parametrization for test_get_url_dict test cases

Adds changelog entry for #649, and uses parametrization for the related
test.

## Why is it important?

With parametrization, all test cases are run independently. That means
that if one of the cases fails, the other cases will still be run.


beniwohli:
Merging this without review, as it is only a changelog change and a refactor of a test

PR #651, opened 2019-11-25
ci(jenkins): enable benchmark run

v1v:

* Run benchmark by default for each commit into the master branch. It does enable us to run the benchmark on a PR basis but only if using the Jenkins UI. 

* Besides, it does enable us to customize what CI stages should be launch using the UI. 

## Why is it important?

APM agents do run benchmark by default.

## e2e tests

- See [build without tag flag and as-is](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/PR-651/15/pipeline)

- See [build with latest changes](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/PR-651/20/)

## UI view

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,3 +27,11 @@ repos:
         exclude: "(elasticapm/utils/wrapt/.*|tests/utils/stacks/linenos.py)"
         language: script
         types: [python]
+
+-   repo: git@github.com:elastic/apm-pipeline-library
+    rev: current
+    hooks:
+    -   id: check-bash-syntax
+    -   id: check-jenkins-pipelines
+        files: ^(.ci/(.*\.groovy|Jenkinsfile)|Jenkinsfile)$
+    -   id: check-jjbb
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -27,11 +27,3 @@ repos:
         exclude: "(elasticapm/utils/wrapt/.*|tests/utils/stacks/linenos.py)"
         language: script
         types: [python]
-
--   repo: git@github.com:elastic/apm-pipeline-library
-    rev: current
-    hooks:
-    -   id: check-bash-syntax
-    -   id: check-jenkins-pipelines
-        files: ^(.ci/(.*\.groovy|Jenkinsfile)|Jenkinsfile)$
-    -   id: check-jjbb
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -151,6 +163,41 @@ pipeline {
...
+          anyOf {
+            branch 'master'
+            expression { return params.Run_As_Master_Branch }
+          }
+          expression { return params.bench_ci }
+        }
+      }
+      steps {
+        withGithubNotify(context: 'Benchmarks', tab: 'artifacts') {
+          dir(env.BUILD_NUMBER) {
```
v1v: This is a way to ensure we can run in parallel as many builds as we want without clashing with the same job/branch over and over. Or in case, the baremetal got issues when deleting the workspace. It might not be required ideally, but that might give us more control in case of any issues with the baremetal as it's a permanent worker in the CI

.pre-commit-config.yaml
```diff
@@ -27,3 +27,11 @@ repos:
         exclude: "(elasticapm/utils/wrapt/.*|tests/utils/stacks/linenos.py)"
         language: script
         types: [python]
+
+-   repo: git@github.com:elastic/apm-pipeline-library
+    rev: current
+    hooks:
+    -   id: check-bash-syntax
+    -   id: check-jenkins-pipelines
+        files: ^(.ci/(.*\.groovy|Jenkinsfile)|Jenkinsfile)$
```
v1v: This should be gone as soon as a new release for the library gets out.

scripts/run-benchmarks.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+set -ex
```
kuisathaverat:
```suggestion
set -e
```

Jenkinsfile
```diff
@@ -151,6 +163,51 @@ pipeline {
...
       }
     }
+    stage('Benchmarks') {
+      agent { label 'metal' }
+      options { skipDefaultCheckout() }
+      environment {
+        HOME = "${env.WORKSPACE}"
+        PATH = "${env.WORKSPACE}/.local/bin:${env.PATH}"
+        PIP_CACHE = "${env.WORKSPACE}/.cache"
+        AGENT_WORKDIR = "${env.WORKSPACE}/${env.BUILD_NUMBER}/${env.BASE_DIR}"
```
mdelapenya: Like it! Let's look into this when using GOPATH for the Go agent

scripts/run-benchmarks.sh
v1v: The sensitive data is already obfuscated within the step `sendBenchmarks.prepareAndRun`, but I agree let's be less verbose to avoid any disclosure of data.

PR #654, opened 2019-11-26
implemented async_capture_span and instrumentation for the aiohttp client

beniwohli:

It adds an async version of `capture_span`, surprisingly named
`async_capture_span`, as well as instrumentation for the
aiohttp HTTP client.

## Why is it important?

Having an asyncio aware version of `capture_span` allows us
to implement instrumentation of asyncio libraries


beniwohli:
closing this, as it'll be split up in two separate PRs, the first one being #657

Issue #598, opened 2019-09-26
Add support to asyncio — Asynchronous I/O


lreuven:
Add support to asyncio — Asynchronous I/O

- [x] Move blocking code to background thread (#607)
- [x] Add async `capture_span` variant (#657)
- [x] Add support for aiohttp client/web (#659)
- [x] Add support for tornado (#661)
- [x] Investigate support for starlette, FastAPI, sanic
- [x] Investigate async support for Django 3.0

beniwohli:
Closing this, as general asyncio support has been implemented

PR #657, opened 2019-11-28
added async_capture_span for decorating coroutines and `async with`

beniwohli:

It adds a new API, `async_capture_span`, which can be used to decorate
coroutines and asynchronous context managers (`async with`).

Additionally, it adds instrumentation for `asyncio.tasks.sleep`.

## Why is it important?

Providing these APIs allows ourselves and our users to instrument
asynchronous applications.

## Related issues

refs #598

### DIFF
--- a/elasticapm/contrib/asyncio/traces.py
+++ b/elasticapm/contrib/asyncio/traces.py
@@ -1,6 +1,5 @@
 #  BSD 3-Clause License
 #
-#  Copyright (c) 2012, the Sentry Team, see AUTHORS for more details
 #  Copyright (c) 2019, Elasticsearch BV
 #  All rights reserved.
 #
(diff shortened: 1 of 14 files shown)


elasticapm/contrib/asyncio/traces.py
```diff
@@ -0,0 +1,69 @@
...
+                self.name,
+                self.type,
+                context=self.extra,
+                leaf=self.leaf,
+                labels=self.labels,
+                span_subtype=self.subtype,
+                span_action=self.action,
+                sync=False,
+                start=self.start,
+            )
```
axw: Just an idea: this+capture_span could be made a bit more DRY by one of these approaches:
 - add a sync slot to capture_span, and setting it to False in the async_capture_span constructor; then `__aenter__`/`__aexit__` could delegate to `__enter__`/`__exit__` as they don't do anything synchronous
 - split the code in `__enter__`/`__exit__` into separate methods that take a `sync` arg, and calling them with `sync=False` from `__aenter__`/`__aexit__`

elasticapm/instrumentation/packages/asyncio/sleep.py
```diff
@@ -0,0 +1,46 @@
...
+
+class AsyncIOSleepInstrumentation(AsyncAbstractInstrumentedModule):
+    name = "asyncio_sleep"
+
+    instrument_list = [("asyncio.tasks", "sleep")]
+
+    async def call(self, module, method, wrapped, instance, args, kwargs):
+        delay = args[0] if args else kwargs["delay"]
+
+        signature = "asyncio.sleep %.3f" % delay
```
axw: Is it useful to have the delay in there? You can already see how long it takes through the duration.

tests/instrumentation/asyncio/base_tests.py
```diff
@@ -0,0 +1,66 @@
...
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from asyncio import tasks
+
+import pytest
+
+import elasticapm
+from elasticapm.conf import constants
+
+pytestmark = [pytest.mark.asyncio]
```
axw: doesn't appear to be used

elasticapm/instrumentation/packages/asyncio/sleep.py
beniwohli: It is not, I'll remove it :)

tests/instrumentation/asyncio/base_tests.py
beniwohli: This is a marker to indicate to pytest that the test functions should be run as coroutines

https://github.com/pytest-dev/pytest-asyncio#pytestmarkasyncio

axw: Ah, neat - TIL. Thanks.

elasticapm/contrib/asyncio/traces.py
beniwohli: I'm a bit wary of introducing more layers into this code path, as every layer ends up being captured as span stack frame. We could refactor this a bit to capture the stack trace directly in the context manager and pass it into the `begin_span` methods.

If it's ok with you, I'll leave this as is, and do the refactor in a follow-up PR. I'd like to merge this soonish, as I currently have to do a lot of git dances locally to merge this into other branches that depend on it, only to back it out again before pushing.

axw: > I'm a bit wary of introducing more layers into this code path, as every layer ends up being captured as span stack frame. We could refactor this a bit to capture the stack trace directly in the context manager and pass it into the begin_span methods.

Couldn't you just add 1 to skip_frames when calling `end`?

> If it's ok with you, I'll leave this as is, and do the refactor in a follow-up PR. I'd like to merge this soonish, as I currently have to do a lot of git dances locally to merge this into other branches that depend on it, only to back it out again before pushing.

Sounds good. Up to you anyway, I just thought it might help keep maintenance costs down.

PR #660, opened 2019-12-01
Implement Trace Context standard

### LINKED ISSUE #628: Support tracecontext W3C header
See elastic/apm#71

- [ ] implement `tracecontext`
- [ ] forward `tracestate` header

beniwohli:
- [x] support `traceparent` header
 - [x] support `tracestate` header propagation

On incoming requests, we first check for the presence of `traceparent`,
and fall back to `elastic-apm-traceparent`.

For outgoing requests, `traceparent` is always set. `elastic-apm-traceparent`
is set if the config option `use_elastic_traceparent_header` is set to
True. For now, this option defaults to True, but will be changed to False
after a grace period.

## Why is it important?

TraceContext is now a recommendation candidate, time to support it
fully.

## Related issues

closes #628
refs elastic/apm#71

### DIFF
--- a/elasticapm/utils/disttracing.py
+++ b/elasticapm/utils/disttracing.py
@@ -30,19 +30,21 @@
 
 import ctypes
 
+from elasticapm.conf import constants
 from elasticapm.utils.logging import get_logger
 
 logger = get_logger("elasticapm.utils")
 
 
 class TraceParent(object):
-    __slots__ = ("version", "trace_id", "span_id", "trace_options")
+    __slots__ = ("version", "trace_id", "span_id", "trace_options", "is_legacy")
 
-    def __init__(self, version, trace_id, span_id, trace_options):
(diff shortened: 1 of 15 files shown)


elasticapm/utils/disttracing.py
```diff
@@ -79,7 +83,42 @@ def from_string(cls, traceparent_string):
...
+        This method is provided as a hook for framework integrations to provide their own
+        TraceParent implementation. The implementation should return a single string. Multiple
+        values for the same key should be merged using a comma as separator.
+
+        :param headers: a dict-like header object
+        :param key: header name
+        :return: a single string value
+        """
+        # this works for all known WSGI implementations
+        return headers.get(key)
```
basepi: Good work being so thorough with your investigation here. And for providing something to override if Murphy's Law gives us a WSGI implementation with different behavior.

PR #659, opened 2019-11-30
Add support for aiohttp client and server

beniwohli:
## What does this pull request do?

This PR adds integration for the aiohttp web framework, as well as the aiohttp
HTTP client library.

## Why is it important?

~Adding support for more asyncio frameworks and libraries is part of the roadmap.~ BECAUSE ASYNC ALL THE THINGS! (edited for excitement by @basepi)

PR #661, opened 2019-12-06
Add support for tornado.web

basepi:
Figured I'd open a draft even though this is barely more than a skeleton so far.

Adds Tornado support (or it will, anyway, once it's done)

## Related issues

Refs #598


elasticapm/contrib/tornado/__init__.py
```diff
@@ -0,0 +1,62 @@
...
+        if not client:
+            config.setdefault("framework_name", "tornado")
+            config.setdefault("framework_version", tornado.version)
+            client = Client(config)
+        self.app = app
+        self.client = client
+        app.elastiapm_client = client
+
+        # Don't instrument if debug=True in tornado
+        if not self.app.settings.get("debug") and client.config.instrument:
```
beniwohli: IIRC, in other framework integrations, we still instrument the framework in debug mode if `client.config.debug is True`

elasticapm/contrib/tornado/utils.py
```diff
@@ -0,0 +1,74 @@
...
+    pass
+
+
+def get_data_from_request(request, capture_body=False, capture_headers=True):
+    """
+    Capture relevant data from a tornado.httputil.HTTPServerRequest
+    """
+    result = {
+        "method": request.method,
+        "socket": {"remote_address": request.uri, "encrypted": request.protocol == "https"},
```
beniwohli: this should be the IP address (`request.remote_ip`)

elasticapm/contrib/tornado/utils.py
```diff
@@ -0,0 +1,74 @@
...
+            # TODO what if request.body is a future?
+            # https://github.com/tornadoweb/tornado/blob/18d7026853900c56a1879141235c6b6bd51b0a6a/tornado/web.py#L1691-L1699
+            body = tornado.escape.json_decode(request.body)
+        except Exception:
+            pass
+
+        if body is not None:
+            result["body"] = body if capture_body else "[REDACTED]"
+
+    result["url"] = request.get_url()
```
beniwohli: does `request.get_url()` return a string? If so, you need to explode it into its components using the [`get_url_dict()`](https://github.com/elastic/apm-agent-python/blob/ac3b6ffb5893f1b989507eaa671ff8fbcaa53685/elasticapm/utils/__init__.py#L104) helper function

elasticapm/instrumentation/packages/tornado.py
```diff
@@ -0,0 +1,92 @@
...
+from elasticapm.instrumentation.packages.asyncio.base import AbstractInstrumentedModule, AsyncAbstractInstrumentedModule
+from elasticapm.utils.disttracing import TraceParent
+
+
+class TornadoRequestExecuteInstrumentation(AsyncAbstractInstrumentedModule):
+    name = "tornado_request_execute"
+
+    instrument_list = [("tornado.web.RequestHandler", "_execute")]
+
+    async def call(self, module, method, wrapped, instance, args, kwargs):
```
beniwohli: this won't be called, because `call` is only called if there already is a transaction, and it is being sampled. This is the API deficiency for using instrumentation for creating transactions. Basically, you need to also override [`call_if_sampling`](https://github.com/elastic/apm-agent-python/blob/ac3b6ffb5893f1b989507eaa671ff8fbcaa53685/elasticapm/instrumentation/packages/base.py#L133), and create the transaction there.

It might make sense to think about introducing another bit of API to make this less ugly... any ideas?

elasticapm/instrumentation/packages/tornado.py
```diff
@@ -0,0 +1,92 @@
...
+        request = instance.request
+        if constants.TRACEPARENT_HEADER_NAME in request.headers:
+            trace_parent = TraceParent.from_string(request.headers[constants.TRACEPARENT_HEADER_NAME])
```
beniwohli: can you use the new [`TraceParent.from_headers`](https://github.com/elastic/apm-agent-python/blob/ac3b6ffb5893f1b989507eaa671ff8fbcaa53685/elasticapm/utils/disttracing.py#L89) here?

elasticapm/instrumentation/packages/tornado.py
```diff
@@ -0,0 +1,92 @@
...
+            lambda: get_data_from_request(
+                request,
+                capture_body=self.client.config.capture_body in ("transactions", "all"),
+                capture_headers=self.client.config.capture_headers,
+            ),
+            "request",
+        )
+        elasticapm.set_transaction_name("{} {}".format(request.method, request.path), override=False)
+
+        ret = await wrapped(*args, **kwargs)
```
beniwohli: Do you know if Tornado bubbles up any exceptions? If yes, we could catch it here, call `client.capture_exception()` and then re-raise it. For reference: https://github.com/elastic/apm-agent-python/pull/659/files#diff-13cd41618748a89578bdced1f7884738R69-R90

That might make `TornadoHandleExceptionInstrumentation` superfluous if we're lucky :D

basepi: It does not, unfortunately: https://github.com/tornadoweb/tornado/blob/18d7026853900c56a1879141235c6b6bd51b0a6a/tornado/web.py#L1707-L1719

(It calls `_handle_request_exception()` which sends the error and doesn't raise.)

basepi: One of many things that makes Tornado unfriendly to instrumentation.

elasticapm/instrumentation/packages/tornado.py
basepi: I added `AbstractInstrumentedModule.creates_transactions` to solve this issue. I think it's pretty clean unless I'm missing something obvious.

elasticapm/instrumentation/packages/base.py
```diff
@@ -131,6 +172,15 @@ def uninstrument(self):
         self.originals = {}
 
     def call_if_sampling(self, module, method, wrapped, instance, args, kwargs):
+        """
+        This is the function which will wrap the instrumented method/function.
+
+        By default, will call the instrumented method/function, via `call()`,
+        only if a transaction is active. This behavior can be overridden by
```
beniwohli: Might be worth pointing out here what happens if a transaction is active but `is_sampled` set to `False`.

elasticapm/instrumentation/packages/tornado.py
```diff
@@ -0,0 +1,88 @@
...
+        client.begin_transaction("request", trace_parent=trace_parent)
+        elasticapm.set_context(
+            lambda: get_data_from_request(
+                request,
+                capture_body=client.config.capture_body in ("transactions", "all"),
+                capture_headers=client.config.capture_headers,
+            ),
+            "request",
+        )
+        elasticapm.set_transaction_name("{} {}".format(request.method, request.path), override=False)
```
beniwohli: Ideally, we'd use a parametrized path here, so we can group transactions based on e.g. the route. If that's not available, I think the type name of the handler would be a better fit than the path.

basepi: Yeah this is definitely not ideal. Unfortunately the RequestHandler object doesn't get any reference to the rule that routed the request to it. I could potentially dive the Application object and try to match up Rules, but that could get hairy and imprecise, as there's nothing stopping a user from using the same RequestHandler for multiple routes.

I do think `GET {type(RequestHandler)}` is the better way to handle this, though. Good suggestion. I'll also see if I can come up with any better way to refer to the routing rule itself.

basepi: Yeah this is definitely not ideal. Unfortunately the RequestHandler object doesn't get any reference to the rule that routed the request to it. I could potentially dive the Application object and try to match up Rules, but that could get hairy and imprecise, as there's nothing stopping a user from using the same RequestHandler for multiple routes.

I do think `GET {type(RequestHandler)}` is the better way to handle this, though. Good suggestion. I'll also see if I can come up with any better way to refer to the routing rule itself.

basepi:
This is code complete. I need to work out our support matrix, fix up the test suite (for `async` errors in py2 mostly), and write up some documentation.

basepi:
This is ready for review! We only support tornado 6.0+ (see reasons [here](https://github.com/elastic/apm-agent-python/pull/661/commits/bfe354635ab481e5580d05453636d6cbe4ec2ec9).

@beniwohli I'm not sure what's going wrong with the tests -- I added all the exclusions to the jenkins-exclude file, but I'm still getting a bunch of syntax errors in non-3.7 tests due to `async def`. What am I missing?

beniwohli:
@basepi the SyntaxErrors are due to pytest's test discovery. Even when excluded via Jenkins excludes, pytest will still cause Python to parse the code. You have to tell pytest to ignore the the code. We currently do this for all files that are prefixed with `py3_`, as well for all files that are in an `asyncio` folder, including subfolders.

https://github.com/elastic/apm-agent-python/blob/721601eec8abbdcbabf4d2567386012fdf9ba1bd/Makefile#L13

If you move the Tornado tests into `tests/contrib/asyncio`, it should work.

I'll do a proper code review on Monday :)

.flake8
```diff
@@ -1,4 +1,11 @@
 [flake8]
-exclude=elasticapm/utils/wrapt,build,src,tests,dist,conftest.py,setup.py
+exclude=
```
beniwohli: any reason for this change? It is more readable/scannable for sure, so I'm good with merging it, just wondering if there's any other reason behind it.

basepi: Yes, since vscode invokes flake8 with a full path on save, the excludes weren't being observed. The `/**` format seems to fix the issue and is still semantically the same.

(Splitting onto different lines was just additional readability)

basepi:
Pretty sure the remaining test failures are the same ones as are failing on `master`, but I'm going to wait until we get them fixed because I like green. :)

beniwohli:
@basepi I think most failures are due to an import error: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/PR-661/19/pipeline/158#step-1296-log-1482

Looks like you might need to move the import of `elasticapm.contrib.tornado.utils ` into the `call` methods. Not pretty, but I don't see a better solution.

basepi:
Good catch. I saw those client errors and stopped reading before I got to the "good ones."

Fixed now, I think.

docs/supported-technologies.asciidoc
```diff
@@ -6,7 +6,7 @@ The Elastic APM Python Agent comes with support for the following frameworks:
 
  * <<django-support,Django>>
  * <<flask-support,Flask>>
- 
+
```
bmorelli25: Tornado is a framework, right? Let's add it here if so. It looks like Aiohttp isn't a framework, but we should probably add it here too. If it makes more sense, we can change the preceding sentence to remove framework. Something like: "The Elastic APM Python Agent comes with support for the following:"
```suggestion
 * <<supported-aiohttp,Aiohttp Server>>
 * <<supported-tornado,Tornado>>
```

docs/tornado.asciidoc
```diff
@@ -0,0 +1,125 @@
...
+[float]
+[[tornado-performance-metrics]]
+==== Performance metrics
+
+If you've followed the instructions above, the agent has installed our
+instrumentation within the base RequestHandler class in tornado.web. This will
+measure response times, as well as detailed performance data for all supported
+technologies.
+
+NOTE: due to the fact that `asyncio` drivers are usually separate from their
```
bmorelli25:
```suggestion
NOTE: Due to the fact that `asyncio` drivers are usually separate from their
```

docs/supported-technologies.asciidoc
```diff
@@ -6,7 +6,8 @@ The Elastic APM Python Agent comes with support for the following frameworks:
 
  * <<django-support,Django>>
  * <<flask-support,Flask>>
-
+ * <<supported-aiohttp,Aiohttp Server>>
+ * <<supported-tornado,Tornado>>
```
bmorelli25: Sorry, I forgot the blank line after the final bullet.
```suggestion
 * <<supported-tornado,Tornado>>
 
```

basepi:
jenkins run the tests please

beniwohli:
Awesome work getting this done @basepi!

PR #666, opened 2019-12-11
move django specific stuff in conftest to subdirectory

beniwohli:
also, use fixtures as pytest plugins instead of importing them

It uses pytest's ability to pick up folder-specific conftests.
This allows us to move django-specific setup to where it belongs.

## Why is it important?

Currently, when running the django/flask matrix builds,
all non-specific tests (as in, anything that doesn't have
any specific marks) also runs. This leads to a lot of repeated
tests.

Ideally, we'd introduce new marks for Django/Flask, and a new
matrix row that only runs non-marked tests.
Having the conftest separate is the first step for this.

### DIFF
--- a/.ci/.jenkins_framework.yml
+++ b/.ci/.jenkins_framework.yml
@@ -1,6 +1,7 @@
 # this is a limited list of matrix builds to be used for PRs
 # see .jenkins_framework_full.yml for a full list
 FRAMEWORK:
+  - none
   - django-1.11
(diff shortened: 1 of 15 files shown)


beniwohli:
I compared the list of passed tests per Python version with current master, and except for `tests.utils.compat_tests.test_multivalue_dict`, all were present and accounted for. That was easily fixed in 5ad4bc7.

PR #667, opened 2019-12-12
updated our support matrix

beniwohli:

Marks some old versions as deprecated, while adding some new versions.

## Why is it important?

Some versions have reached EOL a long time ago (Django 1.10 and older),
others a bit more recently (Python 3.4 went EOL on March 16 2019),
and lastly, some Flask versions are just really old (Flask doesn't
have EOL policies).

The plan is to mark these deprecated in 5.4, and drop them from the
support matrix in 6.0. We'd also stop testing these versions,
and remove any old compatibility shims with 6.0.


CHANGELOG.asciidoc
```diff
@@ -25,10 +25,20 @@ https://github.com/elastic/apm-agent-python/compare/v5.3.1\...master[Check the d
...
+
+ * Python 3.4
+ * Django 1.8, 1.9 and 1.11
+ * Flask < 1.0
+
 [float]
 ===== New Features
 
  * Added support for W3C `traceparent` and `tracestate` headers {pull}660[#660]
+ * Added Django 3.0 and Flask 1.1 to the support matrix {pull}666[#]
```
beniwohli: 667 :man_facepalming:

beniwohli:
@nehaduggal @lreuven please weigh in on these plans if you'd like. We don't have any usage statistics for how many users are still on old flask/django versions, but at least for Django, the versions we intend to stop supporting went EOL years ago.

As for Python 3.4, [download statistics](https://pypistats.org/packages/elastic-apm) show it below 1%.

PR #668, opened 2019-12-12
added instrumentation support for aiopg

beniwohli:

Adds instrumentation for the aiopg Cursor, which wraps
psycopg2's cursor. While we would capture spans from
psycopg2 without this instrumentation, they all would
have a duration of (close to) 0, as we'd only measure
the creation of the future object.

## Why is it important?

More asyncio instrumentations!!


tests/scripts/envs/aiopg.sh
```diff
@@ -0,0 +1,7 @@
+export PYTEST_MARKER="-m aiopg"
+export DOCKER_DEPS="postgres"
+export POSTGRES_HOST="postgres"
+export POSTGRES_USER="postgres"
+export POSTGRES_DB="elasticapm_test"
+export POSTGRES_HOST="postgres"
+export POSTGRES_PORT="5432"
```
basepi: Where is this file used? I can't find references to it.

beniwohli: great question, and a good sign that this needs to be documented properly. This is part of our Jenkins matrix build. Based on the `FRAMEWORK` axis, we look for `.sh` files, first based on the full framework name (e.g. `aiohttp-newest.sh`), then for one without version (`aiohttp.sh`). These are then sourced both by the process that boots up the docker images (this allows us to define additional docker containers to be booted with the test run, `postgres` in this case), as well as in the docker container that executes the test (to define which pytest markers to filter by, DB credentials etc).

https://github.com/elastic/apm-agent-python/blob/14988bdb9560862342a05a117f03649cfc9622bb/tests/scripts/docker/run_tests.sh#L26-L36

https://github.com/elastic/apm-agent-python/blob/d5df7620654b4dcdcadde764fef38e316c5a36f6/tests/scripts/run_tests.sh#L11-L22

basepi: Perfect, thanks for the clarification.

PR #673, opened 2019-12-13
PoC ci(jenkins): let's rebuild if checkout issues

v1v:

Propose a workaround to retry until the implicit git checkout works. 
 
## Why is it important?

Let's do the git checkout fails tolerant with the current CI infrastructure.

## Test cases
For this particular test cases, we have used the `aborted` post action.

- If manually aborted

If the build got aborted [#3](https://apm-ci.elastic.co/job/apm-agent-python/job/apm-agent-python-mbp/job/PR-673/3/)

image

Then, [#4](https://apm-ci.elastic.co/job/apm-agent-python/job/apm-agent-python-mbp/job/PR-673/4/) got executed automatically.

- If automatically canceled.

If the build  [#7](https://apm-ci.elastic.co/job/apm-agent-python/job/apm-agent-python-mbp/job/PR-673/7/) got triggered

image

Then [#6](https://apm-ci.elastic.co/job/apm-agent-python/job/apm-agent-python-mbp/job/PR-673/6/)
got canceled 

image

The customised git checkout happened so it's not required to rebuild it.

## Tasks
- [ ] Avoid infinite loops somehow.

## Related issues
closes #ISSUE


Jenkinsfile
```diff
@@ -298,6 +298,19 @@ pipeline {
       }
       notifyBuildResult()
     }
+    aborted {
```
kuisathaverat: we are overengineering to resolve an issue that is not here if it is something that we will remove as soon as the networks issues disappear I am ok with it, but the network issue is there and will appear on other steps

kuisathaverat: also, it would not bee the same build, the commit can change

kuisathaverat: this gives me evil ideas to recover from known failures 😈

v1v: Yep, I just wanted to see whether this particular approach in real. I do not like it though

v1v: > also, it would not bee the same build, the commit can change

Actually that's not really an issue though as the cancel running old builds will be in place anyway. I initially thought it could be potentially an issue but not in our case as far as I'm concerned. 

> this gives me evil ideas to recover from known failures 😈

I initially thought to create a step to help with this, but the only issue that I could not solve is the one regarding the parameters when using the build step, if we could have a way to dynamically know what parameters are available for the given `JOB_NAME` then we could encapsulate the logic in a shared step.

v1v:
Let me add a bit of context since I closed this PR in favor of our shared library.

## What did I do?
- If a git timeout happens then the pipeline will be triggered only if the last two previous builds did not fail, in other words, let's ensure no infinite loops.

You will receive notifications as normal, that's something we might need to figure out whether we would like to restrict the notifications if there is a git timeout, to avoid spamming for some unrelated code issues but environmental ones. Let's see whether this approach mitigates the current issues .

Cheers

PR #679, opened 2019-12-18
added ApiKey option

beniwohli:

Adds support for the ApiKey authorization method

## Why is it important?

This will be supported in a future version of APM Server


tests/bdd/features/api_key.feature
```diff
@@ -0,0 +1,18 @@
...
+    Given an agent
+    When an api key is set in the config
+    When a secret_token is set in the config
+    Then the api key is sent in the Authorization header
+
+  Scenario: A configured secret token is sent if no api key is configured
+    Given an agent
+    When a secret_token is set in the config
+    When an api key is not set in the config
+    Then the secret token is sent in the Authorization header
```
basepi: Do we have a plan for how to keep these up to date? Not very DRY (and has potential for defeating the purpose of gherkin specs if we accidentally diverge at all from the central version).

elasticapm/transport/base.py
```diff
@@ -62,6 +62,7 @@ class Transport(object):
 
     def __init__(
         self,
+        client,
```
basepi: Any particular reason this is a required arg rather than `client=None`?

basepi:
Note: this should be labeled `experimental` in the docs.

tests/bdd/features/api_key.feature
beniwohli: added download script in 9bfc402

elasticapm/transport/base.py
beniwohli: I made it a positional argument to indicate that it is required for proper functionality. It adds a bit of code churn in some tests where we now have to explicitly pass `None` as client, but I'm personally fine with that. But no strong opinions, I can change it if you want

basepi: Sounds fine.

PR #680, opened 2019-12-19
[DOC] Provide an example when no framework is used

lucabelluccini:

We provide an example when no framework is used.
Feel free to edit or enrich with additional references to other documentation pages.

## Why is it important?

It was not clear the spans from supported libraries were collected, but they required a transaction.


docs/how-the-agent-works.asciidoc
```diff
@@ -26,6 +26,40 @@ Examples of this are
...
 E.g. for Django, you need to add `elasticapm.contrib.django` to `INSTALLED_APPS`.
 
+[float]
+[[how-it-works-no-framework]]
+==== What if you are not using a framework
+
+In case you are not using a supported framework, for example a simple Python script, you can still
+leverage the <<automatic-instrumentation,Automatic Instrumentation>>.
+
+In order to collect the Spans generated by the supported libraries, you will need to create a Transaction.
```
basepi: Can we add a note here about calling `elasticapm.instrument()` in addition to creating a transaction? That's a key requirement for actually getting spans from instrumented libraries.

docs/how-the-agent-works.asciidoc
```diff
@@ -26,6 +26,40 @@ Examples of this are
 Framework integrations require some limited code changes in your app.
 E.g. for Django, you need to add `elasticapm.contrib.django` to `INSTALLED_APPS`.
 
+[float]
+[[how-it-works-no-framework]]
+==== What if you are not using a framework
+
+In case you are not using a supported framework, for example a simple Python script, you can still
```
bmorelli25:
```suggestion
If you're not using a supported framework, for example, a simple Python script, you can still
```

docs/how-the-agent-works.asciidoc
```diff
@@ -26,6 +26,40 @@ Examples of this are
...
+leverage the <<automatic-instrumentation,Automatic Instrumentation>>.
```
bmorelli25:
```suggestion
leverage the agent's <<automatic-instrumentation,automatic instrumentation>>.
```

docs/how-the-agent-works.asciidoc
```diff
@@ -26,6 +26,40 @@ Examples of this are
...
+
+In order to collect the Spans generated by the supported libraries, you will need to create a Transaction.
+It is up to you to define what do you consider a Transaction within your application: it can be the whole execution of the
```
bmorelli25:
```suggestion
It is up to you to determine what you consider a Transaction within your application -- it can be the whole execution of the
```

docs/how-the-agent-works.asciidoc
```diff
@@ -26,6 +26,40 @@ Examples of this are
...
+script or a part of it.
+
+The example below will consider the whole execution as a single transaction and you'll see 2 HTTP request Spans in it.
```
bmorelli25:
```suggestion
The example below will consider the whole execution as a single transaction with two HTTP request Spans in it.
```

docs/how-the-agent-works.asciidoc
lucabelluccini: Hello @basepi ! I've added the requirement to invoke `elasticapm.instrument()` just once.
I am not 100% sure of the wording used.

lucabelluccini:
Hello @bmorelli25 ! Thank you a lot for helping out and the suggested changes.

PR #689, opened 2020-01-13
use the configured hostname in get_system_info

### LINKED ISSUE #688: HOSTNAME value ignored and socket.gethostname() sent
I've trouble in configuring the HOSTNAME, If I set it in the django settings or in the enviorment nothing changes. the `socket.gethostname()` is always the value sent.

**To Reproduce**
- setup project with APM
- add 
```
ELASTIC_APM = {
  ..
    'HOSTNAME': 'myhostname'
...
}
```

**Expected behavior**: 
To see `myhostname` in the host list in kibana

- OS: Linux/Mac
- Python version: 3.7
- Framework and version Django 2.2
- APM Server version: 7.5
- Agent version: elastic-apm==5.3.2

**Additional context**

Debugging the code, in the `config` the correct value is there, however the 
socket.gethostname() value is sent

beniwohli:

We currently ignore the hostname configuration in the
system info collection. That's no good.

Thanks to @esseti for finding the bug and suggesting a fix!

## Related issues

fixes #688

### DIFF
--- a/tests/client/client_tests.py
+++ b/tests/client/client_tests.py
@@ -34,6 +34,7 @@ from __future__ import absolute_import
 
 import os
 import platform
+import socket
 import sys
 import time
 from collections import defaultdict
@@ -79,6 +80,7 @@ def test_system_info(elasticapm_client):
         mocked.return_value = {}
         system_info = elasticapm_client.get_system_info()
     assert {"hostname", "architecture", "platform"} == set(system_info.keys())
+    assert system_info["hostname"] == socket.gethostname()
 
 
 @pytest.mark.parametrize("elasticapm_client", [{"hostname": "my_custom_hostname"}], indirect=True)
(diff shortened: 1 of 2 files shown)


tests/client/client_tests.py
```diff
@@ -81,6 +81,13 @@ def test_system_info(elasticapm_client):
     assert {"hostname", "architecture", "platform"} == set(system_info.keys())
 
 
+@pytest.mark.parametrize("elasticapm_client", [{"hostname": "my_custom_hostname"}], indirect=True)
```
basepi: Does it make sense to add a test that checks `hostname` but without an explicitly configured hostname? To make sure that we're populating it either way? (This may be covered by another test, I haven't checked.)

beniwohli: Good idea, added (somewhat sub optimally, see commit message) in 5b1b81e0

PR #692, opened 2020-01-16
Fix test suite when aiohttp is not installed

basepi:

Due to the reference in conftest.py, the test suite would fail if aiohttp were not installed, even if those tests were not targeted. This fixes that issue by instead requiring `pytest-aiohttp`.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -59,11 +59,7 @@ collect_ignore = ["build", "src"]
 
 pytest_plugins = ["tests.fixtures"]
 
-for module, fixtures in {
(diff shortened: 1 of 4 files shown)


basepi:
Huh. You're totally right. It shouldn't ever be added to the pytest_plugins. Maybe I'll do a bit more experimenting.

basepi:
Figured it out. Apparently `pip uninstall aiohttp` did not remove the `aiohttp/` directory in site-packages in my virtualenv. So the import was successful, but empty? I don't know, it's working now that I deleted that folder. Closing.

PR #694, opened 2020-01-20
Add Starlette / FastAPI middleware

### LINKED ISSUE #686: Add support for FastAPI
Currently, it seems FastAPI is unsupported. Would be great to add the Python Framework to the list of those supported.

Am I correct in assuming we could write our own middleware for this Framework using this library?

Sparkycz:
Closes #686


Makefile
```diff
@@ -10,7 +10,7 @@ flake8:
 test:
 	if [[ "$$PYTHON_VERSION" =~ ^(3.5|3.6|3.7|3.8|nightly|pypy3)$$ ]] ; then \
 	py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT); \
-	else py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT) --ignore-glob='*/py3_*.py' --ignore-glob='*/asyncio/*'; fi
+	else py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT) --ignore-glob='*/py3_*.py' --ignore-glob='*/aiohttp/*'; fi
```
basepi: Oh, also revert this as part of my previous review. We're going to put everything async under `tests/contrib/asyncio` for now.

elasticapm/contrib/starlette/utils.py
```diff
@@ -0,0 +1,162 @@
...
+    Args:
+        request (Request)
+
+    Returns:
+        str
+    """
+    body = await request.body()
+    await set_body(request, body)
+
+    return body.decode("utf-8")
```
basepi: I don't think this is quite right.

The problem with this solution is that when you call `request.body()`, that function [uses `request.stream()` to get the body](https://github.com/encode/starlette/blob/20973f2ed31f4fe7faca988f8590fe3eaf70541e/starlette/requests.py#L194-L195). Because you haven't reset the state around streams, if the user then calls `request.stream()` they will get an error.

I think if you set `request._stream_consumed = False` that will make this work.

Alternatively, you could just set `request._body` and not touch `_receive` at all. As far as I can tell, everything will work, as `request._body` [short-circuits streaming](https://github.com/encode/starlette/blob/20973f2ed31f4fe7faca988f8590fe3eaf70541e/starlette/requests.py#L169-L172). That removes the need for `set_body` entirely, unless I'm missing something.

elasticapm/contrib/starlette/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+        trace_parent = TraceParent.from_headers(dict(request.headers))
+        self.client.begin_transaction("request", trace_parent=trace_parent)
+
+        request_data = await get_data_from_request(
+            request,
+            capture_body=self.client.config.capture_body in ("transactions", "all"),
+            capture_headers=self.client.config.capture_headers,
+        )
+
+        elasticapm.set_context(request_data, "request")
```
basepi: We like to wrap these types of processing functions in a lambda, which `set_context()` only calls if the transaction is actually sampled. This way for unsampled transactions we can skip a lot of work, which is particularly relevant if we're capturing streaming bodies or the like.

Here's an example: https://github.com/elastic/apm-agent-python/blob/721601eec8abbdcbabf4d2567386012fdf9ba1bd/elasticapm/contrib/aiohttp/middleware.py#L67-L74

elasticapm/contrib/starlette/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+        elasticapm.set_transaction_name("{} {}".format(request.method, request.url.path), override=False)
+
+    async def _request_finished(self, response: Response):
+        """Captures the end of the request processing to APM.
+
+        Args:
+            response (Response)
+        """
+        response_data = await get_data_from_response(response, capture_headers=self.client.config.capture_headers)
+        elasticapm.set_context(response_data, "response")
```
basepi: Same as above, should use a lambda

elasticapm/contrib/starlette/utils.py
Sparkycz: I admit I hadn't tested as much as I got that problem.
Loading of request body out of the method of business logic (e.g. a middleware) seems an antipattern to me. We shouldn't do it in async world at all. However, this is great information for APM tracking, especially in POST methods.

Thanks for advice I'll try the setting of `request._stream_consumed = False`.

elasticapm/contrib/starlette/__init__.py
Sparkycz: I noticed that but `get_data_from_request()` is a coroutine in my case.
Updated..

tests/contrib/asyncio/starlette_tests.py
```diff
@@ -0,0 +1,143 @@
...
+#  FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+#  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+#  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+#  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+#  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+import pytest  # isort:skip
+
+aiohttp = pytest.importorskip("starlette")  # isort:skip
```
basepi: s/aiohttp/starlette/

Gotta love copy pasta! :)

elasticapm/contrib/starlette/utils.py
basepi: What do you think about my alternate solution, just setting `request._body` with the body? That way you could avoid overriding `_receive` completely. Seems cleaner to me, but I want to make sure I'm not missing some reason that's not the best solution.

basepi:
jenkins run the tests please

kuisathaverat:
jenkins run the tests please

basepi:
@Sparkycz Please remember to [sign the CLA](https://www.elastic.co/contributor-agreement). I've noticed that your commits are not associating with your Github account. I recommend adding your commit email (`vladimir.kaspar@heureka.cz`) to your Github account, or you may have to sign the CLA twice. (I'm not 100% sure how that validation operates)

Alternatively you could rebase the commits with a Github-associated email.

Sparkycz:
Rebased all commits due to CLA only..

elasticapm/contrib/starlette/utils.py
Sparkycz: Honestly, I think both of the solutions are wrong and it doesn't matter which one.
however, you're the owner of this repository ;-) and if you want me to change it I'll do it...

Sparkycz:
Should I squash all the commits to one?

elasticapm/contrib/starlette/utils.py
Sparkycz: Updated..

Sparkycz: @basepi I've been testing the middleware in my app and the last solution of set `request._body` doesn't work.  The object of type `starlete.requests.Request` is instanced more times in Starlete (I didn't look for where). If I invoke the request to my app the Starlete framework waits to load (stream) body but it's not able to. 
I'll revert it to the first solution ;-)

Sparkycz: Heh, it doesn't work due to the sequence of operations in async probably.

basepi:
jenkins run the tests please

basepi:
No, we use squash and merge so it should be fine.

elasticapm/contrib/starlette/utils.py
basepi: Fascinating. Thank you for testing my solution, I'm ok with the first solution.

We'll see if any other users run into issues with this. We may have to put it behind a configuration gate if they do. It's definitely not pretty, but like you said, it's "wrong" anyway, but useful.

basepi: Fascinating. Thank you for testing my solution, I'm ok with the first solution.

We'll see if any other users run into issues with this. We may have to put it behind a configuration gate if they do. It's definitely not pretty, but like you said, it's "wrong" anyway, but useful.

PR #697, opened 2020-01-23
Fix elasticsearch tests

basepi:

Fixes elasticsearch tests (failing as of the release of 7.5.1), and removes some brittleness around argument ordering, since that API is dynamically generated and order is not guaranteed.

Also removed our extensive matrix tracking the position of the `body` argument. `elasticsearch-py` [actually doesn't support non-kwarg use of the API](https://elasticsearch-py.readthedocs.io/en/master/api.html#api-documentation), they are only args because there's no way in py2 to have a required kwarg. Made a note in the docs and updated the tests to match.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -34,13 +34,18 @@ This will be the last minor release to support the following versions:
  * Django 1.8, 1.9 and 1.10
  * Flask < 1.0
 
+ In addition, as of this release we no longer support positional args for
+ elasticsearch-py. This is in keeping with the
+ https://elasticsearch-py.readthedocs.io/en/master/api.html#api-documentation[upstream policy]
+ of positional args being unsupported. {pull}697[#697]
+
 [float]
 ===== New Features
 
  * Added support for aiohttp client and server {pull}659[#659]
  * Added support for W3C `traceparent` and `tracestate` headers {pull}660[#660]
  * Added Django 3.0 and Flask 1.1 to the support matrix {pull}667[#667]
- * Added support for aiopg {pull}668[#668] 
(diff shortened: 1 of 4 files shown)


basepi:
@beniwohli I'd prefer to have a review on this since it makes some executive decisions, but I want to get the tests green so I'm going to merge it.

When you're back next week, I would still appreciate a review of this, and can make any changes you recommend at that point.

beniwohli:
@basepi I had a cursory look earlier today, looks good! I'll have another look tomorrow

CHANGELOG.asciidoc
```diff
@@ -34,13 +34,18 @@ This will be the last minor release to support the following versions:
  * Django 1.8, 1.9 and 1.10
  * Flask < 1.0
 
+ In addition, as of this release we no longer support positional args for
```
beniwohli: Wondering if we should tone down this a little bit, it sounds like things would break if the user uses position args.

Something like

> In addition, as of this release we only supported capturing extended information on elasticsearch queries when using keyword arguments with the elasticsearch-py API. This is in keeping with the
 https://elasticsearch-py.readthedocs.io/en/master/api.html#api-documentation[upstream policy]
 of positional arguments being unsupported. {pull}697[#697]

docs/supported-technologies.asciidoc
```diff
@@ -93,6 +93,11 @@ Collected trace data:
  * the query string (if available)
  * the `query` element from the request body (if available)
 
+We recommend using keyword args only with elasticsearch-py, as recommended by
```
beniwohli:
```suggestion
We recommend using keyword arguments only with elasticsearch-py, as recommended by

PR #700, opened 2020-01-24
Enable docker image cache in the CI workers

v1v:
## What does this pull request do?

Build docker images and cache them in the CI workers using the `.ci/.jenkins_python.yml` as the reference to the docker images to be built

## Why is it important?

Speed up the CI overall execution time

## How to test this

```bash
hub checkout pr 700
sh -x .ci/packer_cache.sh
```
_NOTE_: `/usr/local/bin/bash_standard_lib.sh` is required to exist. That particular requirement is already in place for the CI workers.

## Question
- Using a docker volume when running the docker compose could help with the caching? Or that's something already solved when building the docker image?

PR #701, opened 2020-01-27
Enable docker image cache in the CI workers

v1v:

Use the just built docker images from [pipeline](https://apm-ci.elastic.co/job/apm-shared/job/apm-docker-images-pipeline/) that are already available in the internal docker registry and retag them to use them in the CI pipeline.

## Why is it important?

Speed up the CI overall execution time

## How to test this

```bash
hub checkout pr 701
sh -x .ci/packer_cache.sh
```
_NOTE_: `/usr/local/bin/bash_standard_lib.sh` is required to exist. That particular requirement is already in place for the CI workers.

## Issues

superseed #700

### DIFF
--- a/.ci/packer_cache.sh
+++ b/.ci/packer_cache.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+source /usr/local/bin/bash_standard_lib.sh
+
+grep "-" .ci/.jenkins_python.yml | cut -d'-' -f2- | \
+while read -r version;
+do
+    (retry 2 ./tests/scripts/docker/docker_build.sh "${version}") || echo "Error building ${version} Docker image, we continue"
+done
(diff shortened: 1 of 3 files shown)


v1v:
> Are those images in the registry private? If they aren't, we could use their tags directly in the docker-compose file, so we wouldn't need to do all that re-tagging

They are intentionally pushed to the internal docker registry, the idea is to get more control of the process to generate the docker build images, and only fetch them with the packer_cache.sh script which it's under the control of the infra team.

.ci/packer_cache.sh
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env bash
+
+source /usr/local/bin/bash_standard_lib.sh
+
+grep "-" .ci/.jenkins_python.yml | cut -d'-' -f2- | \
+while read -r version;
+do
+    imageName="apm-agent-python:${version}"
+    registryImageName="docker.elastic.co/observability-ci/${imageName}"
```
v1v:
```
[2020-01-27T14:27:03.874Z] + docker build --build-arg PYTHON_IMAGE=python:3.7 -t docker.elastic.co/observability-ci/apm-agent-python:python-3.7 .
```

PR #705, opened 2020-01-28
Support capturing spans from code run in an event loop executor

beniwohli:

To make the current transaction/span available, we wrap the actual
callable with one of our own that sets the transaction/span in the
execution context of the thread.

## Why is it important?

Using an executor (e.g. ThreadPoolExecutor) is a common way to use
blocking calls in an event loop.


elasticapm/instrumentation/packages/asyncio/base.py
```diff
@@ -29,8 +29,33 @@
...
 
 
 class AsyncAbstractInstrumentedModule(AbstractInstrumentedModule):
     async def call(self, module, method, wrapped, instance, args, kwargs):
         raise NotImplementedError()
+
+
+class RunInExecutorInstrumentation(AsyncAbstractInstrumentedModule):
+    name = "executor"
+    instrument_list = [("asyncio.base_events", "BaseEventLoop.run_in_executor")]
```
beniwohli: Might we worth checking if this would also work with uvloop (https://github.com/MagicStack/uvloop/blob/8beacd265651429a558d1b22e200d6d81016f1db/uvloop/loop.pyx#L2646), but I have my doubts, as wrapt can't patch objects defined in C AFAIK

Issue #428, opened 2019-03-19
[META 352] AWS Lambda support


bobthemighty:
I'm currently building a serverless architecture and was hoping to use elastic cloud as a one-stop repository for logs and metrics. Per elastic/apm-agent-python#320 it seems that Lambda is not a supported environment for Elastic-APM.

I'd be happy to put in some work to make this happen, but I'm not entirely clear on where the problems lie. Is there a reason why this is more complex than calling `collect` on the  metric repository and `flush` on the transport? 

If I had a MetricRegistry with a collect_interval of 0ms, and an unthreaded transport that flushed only on demand, I could flush explicitly at the end of the lambda invocation. It would be simple to extend that to a decorator that flushed on exit from the wrapped function.
Is this an enormous simplification of the requirements?

beniwohli:
Hi @bobthemighty. Yes, that would probably work, however it might incur quite a heavy overhead as basically every lambda invocation would result in an HTTP request to your APM Server. That might be ok if you have relatively long-running lambda functions, but for very short running functions (< 100ms or thereabout), the overhead might start to dominate the actual execution time.

I'm not very familiar with how exactly the life cycle of an invoked lambda function is, but if they shut down the interpreter cleanly after the invocation, you might not even have to flush the queue, as we use [`atexit`](https://docs.python.org/3/library/atexit.html) to flush the queue on shutdown.

beniwohli:
That goes in the general direction of what we think might be a workable solution, basically use CloudWatch or possibly SQS as a transport.

Note however that the events that the agent generates are not elasticsearch documents. They first need to be processed by the APM Server, which turns them into ES documents. 

If you want to write your own transport, note that we just merged a pretty large change to the inner workings of transports (#411). This shouldn't make it much more complicated to implement such a transport, though (we might have to move the `init_buffer` inner function to a method though, so you can avoid the GZip compression).

bobthemighty:
> Note however that the events that the agent generates are not elasticsearch documents. They first need to be processed by the APM Server, which turns them into ES documents.

Roger. Is the format you're writing identical between apm-node/python/go, or is there additional language-specific munging? I'm thinking that if we were to write to CloudWatch, and the format is the same, then it's not too hard to provide a library for parsing those logs out and forwarding them to an APM instance.

> note that we just merged a pretty large change to the inner workings of transports (#411). 

[...]

PR #716, opened 2020-02-13
Serverless instrumentation

basepi:

Adds support for instrumenting serverless functions. Logs to CloudWatch so that those logs can be collected and sent to the APM Server later. Disables all background threads.

Note: This "works", but isn't useful without the connecting lambda from CloudWatch to the APM Server. Thus the lack of docs or references to this code.

## Related issues

Ref #428


elasticapm/contrib/serverless/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+
+def get_data_from_request(event, capture_body=False, capture_headers=True):
+    """
+    Capture context data from API gateway event
+    """
+    result = {}
+    if capture_headers and "headers" in event:
+        result["headers"] = event["headers"]
+    if capture_headers and "requestContext" in event:
+        result["requestContext"] = event["requestContext"]
```
basepi: This is not standard fare for the `request` piece of the context dictionary. In [the original data structure](https://docs.aws.amazon.com/lambda/latest/dg/services-apigateway.html) it's named `requestContext`. On the one hand, the camelCase makes me uncomfortable and doesn't match our spec at all. On the other hand, it's easy to identify where it comes from, if you're used to dealing with AWS lambdas. Should I rename it or leave it?

elasticapm/contrib/serverless/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+import json
+
+import elasticapm
+from elasticapm.base import ServerlessClient
+from elasticapm.conf import constants
+from elasticapm.utils import compat, encoding, get_name_from_func
+from elasticapm.utils.disttracing import TraceParent
+
+
+class capture_serverless(object):
```
basepi: @beniwohli Any thoughts on testing strategy here? I could write useless tests against dummy `event` data to get coverage, but that's not an actual useful test, and I'm not sure how to get real data from AWS lambda in our test framework.

On the ServerlessClient/ServerlessTransport end of things, I suppose I could test to make sure no threads are created, or that it correctly `print()`s the spans instead of trying to send them. Are those useful tests?

elasticapm/contrib/serverless/__init__.py
beniwohli: Good question. I'm tending towards underscores to be in sync with all other fields, but I see your point. Maybe worth bringing it up with the rest of the team, as this probably needs to go through our "adding new fields to context" process sooner or later.

elasticapm/contrib/serverless/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+    if capture_headers and "requestContext" in event:
+        result["requestContext"] = event["requestContext"]
+    if "httpMethod" not in event:
+        # Not API Gateway
+        return result
+
+    result["method"] = event["httpMethod"]
+    if event["httpMethod"] in constants.HTTP_WITH_BODY and "body" in event:
+        body = event["body"]
+        if event.get("isBase64Encoded"):
```
beniwohli: is there a way to avoid base64 decoding / JSON loading if `capture_body` is set to False anyway?

elasticapm/__init__.py
```diff
@@ -29,21 +29,31 @@
 #  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 import sys
 
+from elasticapm.base import Client  # noqa: F401
+from elasticapm.conf import setup_logging  # noqa: F401
+from elasticapm.contrib.serverless import capture_serverless  # noqa: F401
+from elasticapm.instrumentation.control import instrument, uninstrument  # noqa: F401
+from elasticapm.traces import (  # noqa: F401
```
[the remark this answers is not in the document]
basepi: As far as I can tell, both are running on that file in my IDE and both are happy. Took some experimentation as to the location of the `# noqa: F401` line (which is why I know both were running).

Could you try checking it out and see if your IDE gives you problems?

elasticapm/contrib/serverless/__init__.py
```diff
@@ -0,0 +1,190 @@
+#  BSD 3-Clause License
```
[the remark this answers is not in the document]
basepi: Definitely. Will do this.

elasticapm/contrib/serverless/__init__.py
basepi: My thought yesterday was that using sample data meant that we couldn't check for regressions against AWS, which is true. What I've realized since then is that it's still useful for checking for regressions in our code, so I'll write a test using sample data.

elasticapm/contrib/serverless/__init__.py
```diff
@@ -0,0 +1,190 @@
...
+                return func(*args, **kwds)
+
+        return decorated
+
+    def __enter__(self):
+        """
+        Transaction setup
+        """
+        trace_parent = TraceParent.from_headers(self.event.get("headers"))
+        self.transaction = self.client.begin_transaction("request", trace_parent=trace_parent)
```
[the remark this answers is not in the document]
basepi: Not a bad idea. I think `request` is still probably the correct transaction type for when the trigger originates with API Gateway. So maybe I'll just re-type it if I don't see `httpMethod` in the event.

elasticapm/contrib/serverless/__init__.py
basepi: Definitely. I'll make that change.

basepi:
Closing for now. We will probably go a different direction (via Lambda extensions)

PR #727, opened 2020-02-19
update output of `elasticapm test` management command

beniwohli:

We are not able to show an URL to the error, so we explain
how the user can find the error in Kibana themselves.

## Why is it important?

Telling the user to open an URL that isn't there is a bit confusing...


beniwohli:
Test failure is due to psutil/pypy incompatibility, see https://github.com/elastic/apm-agent-python/pull/725

PR #726, opened 2020-02-19
110353 add windows testing to python agent

ElWPenn:
## What does this pull request do?

Adds additional tests for windows on different versions of python

## Why is it important?

Resolves potential flakiness from Appveyor tests by deprecating them

## Related issues
closes

PR #730, opened 2020-02-20
ci(jenkins): install python3 for the given VERSION

v1v:

Install python with the version that's the same that the given VERSION or the closest one. Therefore the VERSION format could be:
- fully
- major
- major.minor
- semver...

## Why is it important?

It will help to install python3 in windows just given the major.minor version rather than the major.minor.patch version

## Related issues

Caused by https://github.com/elastic/apm-agent-python/pull/726

## Tests

```powershell
$Env:VERSION = "3.8"
.\install-python.ps1
```

image

### DIFF
--- a/scripts/install-python.ps1
+++ b/scripts/install-python.ps1
@@ -1,0 +1,9 @@
+# Abort with non zero exit code on errors
+$ErrorActionPreference = "Stop"
+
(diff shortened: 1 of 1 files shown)


v1v:
Most likely we will provide a new step in our shared library to help with the CI pipeline and keep those implementation details in one place.
I'll merge this PR for the time being, then @ElWPenn could potentially use this script in his ongoing PR.

PR #731, opened 2020-02-20
pypy2/psutil also breaks for older versions of psutil, let's exclude them too

beniwohli:

Excludes older psutil versions on pypy2

## Why is it important?

Because it breaks the full matrix test run


beniwohli:
Turns out we haven't set upper limits in the psutil-4.0 and psutil-5.0 requirements files, so they too ran the newest versions :man_facepalming:. Luckily, no other requirements files had this particular snafu.

Closing and opening another one with fixed requirements files.

PR #733, opened 2020-02-21
Fix aiohttp client

alxpy:
Fix for broken `aiohttp.client` tracker.

`url` may be not a string:
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/client.py#L1083
https://github.com/aio-libs/aiohttp/blob/master/aiohttp/typedefs.py#L45

### DIFF
--- a/tests/instrumentation/asyncio/aiohttp_client_tests.py
+++ b/tests/instrumentation/asyncio/aiohttp_client_tests.py
@@ -31,6 +31,7 @@
 import pytest  # isort:skip
 
 aiohttp = pytest.importorskip("aiohttp")  # isort:skip
+yarl = pytest.importorskip("yarl")  # isort:skip
 
 from elasticapm.conf import constants
 from elasticapm.utils.disttracing import TraceParent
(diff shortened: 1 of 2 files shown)


beniwohli:
@alxpy great catch, thanks for opening the PR! If you're up for it, could you write a test for this scenario? You could copy https://github.com/elastic/apm-agent-python/blob/6598165320d6891aa50bb6b178a6a06043efb04a/tests/instrumentation/asyncio/aiohttp_client_tests.py#L41-L59 and turn the `waiting_httpserver.url` into a yarl `URL` object before passing it to `session.get()`. Then, you can use this command to run the tests

    ./tests/scripts/docker/run_tests.sh python-3.8 aiohttp-newest

I can also add the test myself before merging if you prefer :)

alxpy:
I'll write test)

alxpy:
@beniwohli done

tests/instrumentation/asyncio/aiohttp_client_tests.py
```diff
@@ -38,10 +39,14 @@
 pytestmark = [pytest.mark.asyncio, pytest.mark.aiohttp]
 
 
-async def test_http_get(instrument, event_loop, elasticapm_client, waiting_httpserver):
+@pytest.mark.parametrize("use_yarl", [True, False])
```
beniwohli: Great use of `parametrize`, much better than my suggestion of copying the test :+1:

tests/instrumentation/asyncio/aiohttp_client_tests.py
```diff
@@ -29,6 +29,7 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 import pytest  # isort:skip
+import yarl
```
beniwohli: @alxpy we need to move this import below the `pytest.importorskip` line. `yarl` is only available if aiohttp is installed. This is a side effect of how our test matrix build works. It only installs the requirements needed for a certain test module, and we use `importskip` to gate that.

Only noticed this after submitting the review, sorry!

alxpy: no problem, done)

beniwohli:
@alxpy Awesome, thanks!

PR #736, opened 2020-02-24
ci(jenkins): send downstream build data to ES is not required anymore

v1v:

Get rid of the sendData to ES for the downstream jobs.

## Why is it important?

We have changed our process to send data to ES, and the legacy approach is now deprecated.

## Related issues
closes #ISSUE

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -274,11 +274,6 @@ pipeline {
           def processor = new ResultsProcessor()
           processor.processResults(mapResults)
           archiveArtifacts allowEmptyArchive: true, artifacts: 'results.json,results.html', defaultExcludes: false
-          catchError(buildResult: 'SUCCESS') {
-            def datafile = readFile(file: "results.json")
-            def json = getVaultSecret(secret: 'secret/apm-team/ci/apm-server-benchmark-cloud')
-            sendDataToElasticsearch(es: json.data.url, data: datafile, restCall: '/jenkins-builds-test-results/_doc/')
-          }
         }
       }
       notifyBuildResult()


Jenkinsfile
```diff
@@ -274,11 +274,6 @@ pipeline {
           def processor = new ResultsProcessor()
           processor.processResults(mapResults)
           archiveArtifacts allowEmptyArchive: true, artifacts: 'results.json,results.html', defaultExcludes: false
-          catchError(buildResult: 'SUCCESS') {
-            def datafile = readFile(file: "results.json")
-            def json = getVaultSecret(secret: 'secret/apm-team/ci/apm-server-benchmark-cloud')
-            sendDataToElasticsearch(es: json.data.url, data: datafile, restCall: '/jenkins-builds-test-results/_doc/')
```
cachedout: Should we [modify the library](https://github.com/elastic/apm-pipeline-library/blob/master/vars/sendDataToElasticsearch.groovy) to raise a warning if this is called? (Or just remove it?) I don't see any mention of it being deprecated.

v1v: The step is not deprecated but the call in this particular pipeline. `sendDataToElasticsearch` is still a valid step and it's in fact used by the `notifyBuildResult` step to send data to ES.

The `jenkins-builds-test-results` index is not something we are using anymore, and used to collect the status of the downstream jobs for the current pipeline and store their results, but not the process we wanna keep using from now on.

cachedout: Ahh, OK. I misunderstood what you meant originally. Thanks for the clarification!

PR #741, opened 2020-02-27
Update PR and Bug templates

basepi:

We've already discussed removing "Why is this important?" from the
PR template. I went a step further and removed "Expected behavior"
from the Bug_report template, as I think it's almost always addressed in
the "Describe the bug" section.

@beniwohli WDYT?


basepi:
@beniwohli Any thoughts here?

beniwohli:
Oh sorry, forgot to submit my 👍

PR #745, opened 2020-03-03
remove Python 3.4 from the test matrix and add deprecation warning

beniwohli:

Python 3.4 is EOL since March 2019. More and more of our tooling
(pytest, coverage, ...) don't support it anymore, making it
tedious to keep up support on our side.

This PR removes 3.4 from the test matrix, as well as adding a 
deprecation warning when the user runs an unsupported version.

At this time, we're not actively removing compatibility code
for 3.4, or add new code that is not compatible with 3.4,
although this will happen organically going forward.


sethmlarson:
Should probably set this somewhere in `setup.py` as well so the package isn't installed erroneously on an unsupported Python version. See https://github.com/elastic/elasticsearch-py/blob/master/setup.py#L56

beniwohli:
@sethmlarson great call, done in 57ba3bf8

PR #746, opened 2020-03-04
ci(jenkins): use CI Windows workers

v1v:

Replace AppVeyor with CI Jenkins.
Configuration-driven, therefore if a new python version is required then it should be added to the file `.ci/.jenkins_windows.yml`
Python versions that are supported so far:
- `3.5`
- `3.6`
- `3.7`
- `3.8`

## Why is it important?

No more dependencies with AppVeyor.

## Related issues
closes #ISSUE

## Actions
- [x] Define a file that contains the required dependencies, therefore we don't need any changes in the Jenkinsfile but in the yml file.
- [x] Test all the scenarios
- [x] Transform sh step with something else for windows if possible.
- [ ] Remove AppVeyor GitHub check in this repository.

## Follow-ups
- Coverage for Windows
- Package generation for Windows
- Support `2.7`


beniwohli:
Is this a replacement for https://github.com/elastic/apm-agent-python/pull/726?

I think for now, we can set `FRAMEWORK` to `none` for all builds. This will run all the tests that are framework independent (the vast majority of tests).

v1v:
@beniwohli , yes, I'm working on a separate PR for the same requirement.

Jenkinsfile
```diff
@@ -24,7 +24,6 @@ pipeline {
     BENCHMARK_SECRET  = 'secret/apm-team/ci/benchmark-cloud'
     OPBEANS_REPO = 'opbeans-python'
     HOME = "${env.WORKSPACE}"
-    PATH = "${env.WORKSPACE}/.local/bin:${env.WORKSPACE}/bin:${env.PATH}"
```
v1v: Otherwise, there will be a clash in the CI Windows workers

Jenkinsfile
```diff
@@ -112,6 +114,12 @@ pipeline {
                 steps: this
                 )
               def mapPatallelTasks = pythonTasksGen.generateParallelTests()
+
+              // Let's now enable the windows stages
+              readYaml(file: '.ci/.jenkins_windows.yml')['windows'].each { v ->
+                def description = "${v.VERSION}-${v.DISTUTILS_USE_SDK.equals('1') ? 'use-sdk' : ''}-${v.WEBFRAMEWORK}"
```
v1v: This will help to run the same version with different parameters, for such the primary key is a combination of `VERSION`, `DISTUTILS_USE_SDK` and `WEBFRAMEWORK`.

Jenkinsfile
```diff
@@ -386,3 +394,41 @@ def releasePackages(){
...
+    node('windows-2019-docker-immutable'){
+      withEnv(["PYTHON=${pythonPath}",
+               "DISTUTILS_USE_SDK=${distutils}",
+               "ASYNCIO=${asyncio}",
+               "WEBFRAMEWORK=${framework}"]) {
+        try {
+          deleteDir()
+          unstash 'source'
+          dir("${BASE_DIR}"){
+            installTools([ [tool: "python${majorVersion}", version: "${version}" ] ])
```
v1v: This will help to install `python2` and `python3`.

.ci/.jenkins_windows.yml
```diff
@@ -0,0 +1,29 @@
+# This is the limited list of matrix builds in Windows, to be triggered on a PR basis
```
cachedout: This diverges a bit from what we have right now. Presently, we have [Python versions](https://github.com/elastic/apm-agent-python/blob/master/.ci/.jenkins_python.yml) and [frameworks](https://github.com/elastic/apm-agent-python/blob/master/.ci/.jenkins_framework.yml) defined in separate files. Then we're using [the Parallel Tasks Generator](https://github.com/elastic/apm-pipeline-library/blob/master/src/co/elastic/matrix/DefaultParallelTaskGenerator.groovy) to create the matrix and define the steps. This seems as though [it partially re-implements that logic](https://github.com/elastic/apm-agent-python/pull/746/files#diff-58231b16fdee45a03a4ee3cf94a9f2c3R421).

I think that this divergence should be fine so long as we aren't expecting large numbers of combinations, which might be difficult to maintain in a single file such as this.

Do we anticipate this will be a problem in the future? (I don't have really strong preferences about which approach is taken. I am more curious about the reasoning here than anything.)

Jenkinsfile
```diff
@@ -112,6 +114,12 @@ pipeline {
                 steps: this
                 )
               def mapPatallelTasks = pythonTasksGen.generateParallelTests()
+
+              // Let's now enable the windows stages
+              readYaml(file: '.ci/.jenkins_windows.yml')['windows'].each { v ->
+                def description = "${v.VERSION}${v.DISTUTILS_USE_SDK.equals('1') ? '-sdk' : ''}-${v.WEBFRAMEWORK}"
+                mapPatallelTasks["windows-${description}"] = generateStepForWindows(v)
+              }
               parallel(mapPatallelTasks)
```
cachedout: It's not related to this PR but I wouldn't mind at all if we took this chance to correct this misspelling and changed `mapPatallelTasks` to `mapParallelTasks` on this line, and L121 and L116. :)

tests/scripts/download_gherkin_features.bat
```diff
@@ -0,0 +1,12 @@
+: Adding gherkin feature download for windows
+@echo off
+SET TMP_FOLDER=.\tests\tempFeatures
+SET FOLDER=.\tests\bdd\features
+SET APM_BRANCH=master
+
+mkdir %TMP_FOLDER%
+curl -s https://codeload.github.com/elastic/apm/zip/%APM_BRANCH% -o %TMP_FOLDER%\features.zip
+7z x %TMP_FOLDER%\features.zip -o%TMP_FOLDER% *.feature -y -r
```
cachedout: It's possible that this is totally valid and I just don't know what it means, but the `7z x ` at the start of this line is surprising to me. What does it do?

tests/scripts/download_json_schema.bat
```diff
@@ -0,0 +1,12 @@
+: Adding schema feature download for windows
+@echo off
+SET TMP_FOLDER=.\tests\tempSchema
+SET FOLDER=.\tests\.schemacache
+SET APM_SERVER_BRANCH=master
+
+mkdir %TMP_FOLDER%
+curl -s https://codeload.github.com/elastic/apm-server/zip/%APM_SERVER_BRANCH% -o %TMP_FOLDER%\%APM_SERVER_BRANCH%.zip
+7z x %TMP_FOLDER%\%APM_SERVER_BRANCH%.zip -o%TMP_FOLDER% *.json -y -r
```
cachedout: Same question about the `7z x ` here as well. :)

.ci/.jenkins_windows.yml
beniwohli: For Windows, we don't plan to have a matrix anywhere as extensive as the one one Linux. For now, we only have one element on the `FRAMEWORK` axis (`none`, which selects all tests that are framework-independent). We might want to add a couple of Windows-specific ones later on, e.g. MS-SQL, but that's about it.

cachedout: Cool! That resolves my concerns then. :) Thanks, @beniwohli

tests/scripts/download_gherkin_features.bat
v1v: Sure :)

- `7z` stated for `7zip` 
- `x` stated for `extract`

In other words, that's the unzip command, see [7zip](https://www.7-zip.org/)

cachedout: TIL! Thanks!

Jenkinsfile
v1v: Did I say that I even rename the variable in one of the iterations I did run and I didn't even see it. Definitely I'm blind!! Good spot. I'll do it here then :)

PR #756, opened 2020-03-12
Add troubleshooting docs

basepi:
It struck me yesterday when I was helping out with a [discuss post](https://discuss.elastic.co/t/python-apm-agent-for-flask-doesnt-report-metrics/223083) that we don't really have a central place to send people for troubleshooting.

I'd love feedback on the relevance of each of these additions, as well as their content and order. It doesn't flow very well as a doc, but I'm not sure how to make it better since it's a hodge-podge list of tips.

### DIFF
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -4,6 +4,10 @@
 Below are some resources and tips for troubleshooting and debugging the
 python agent.
 
+* <<easy-fixes>>
+* <<django-test>>
+* <<agent-logging>>
+
 [float]
 [[easy-fixes]]
 === Easy Fixes
--- a/docs/troubleshooting.asciidoc
+++ b/docs/troubleshooting.asciidoc
@@ -1,0 +1,142 @@
+[[troubleshooting]]
+== Troubleshooting
+
+Below are some resources and tips for troubleshooting and debugging the
+python agent.
+
+[float]
+[[easy-fixes]]
+=== Easy Fixes
+
+Before you try anything else, go through the following sections to ensure that
+the agent is configured correctly. This is not an exhaustive list, but rather
+a list of common problems that users run into.
+
+[float]
+[[debug-mode]]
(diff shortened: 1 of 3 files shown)


docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,142 @@
...
+Before you try anything else, go through the following sections to ensure that
+the agent is configured correctly. This is not an exhaustive list, but rather
+a list of common problems that users run into.
+
+[float]
+[[debug-mode]]
+==== Debug Mode
+
+Most frameworks support a debug mode. Generally, this mode is intended for
+non-prod environments and provides detailed error messages and logging of
```
beniwohli: Generally, I'd try to avoid abbreviations like "prod" in documentation. You never know if the reader knows the abbreviation or maybe interprets it differently than intended.

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,142 @@
...
+app = Flask(__name__)
+app.config["ELASTIC_APM"] = {"DEBUG": True}
+apm = ElasticAPM(app, service_name="flask-app")
+----
+
+[float]
+[[psutil-metrics]]
+==== `psutil` for Metrics
+
+To get CPU and system metrics, `psutil` must be installed. The agent should
```
beniwohli: This is only true for non-Linux systems. On Linux, we get most metrics by reading a bunch of files in `/proc`

docs/troubleshooting.asciidoc
```diff
@@ -0,0 +1,142 @@
+[[troubleshooting]]
+== Troubleshooting
+
+Below are some resources and tips for troubleshooting and debugging the
+python agent.
+
+[float]
```
bmorelli25: Maybe add a jump table at the top for easier scanability.

```suggestion
python agent.

* <<easy-fixes>>
* <<django-test>>
* <<agent-logging>>

[float]
```

PR #761, opened 2020-03-19
fixed destination issue with old versions of cassandra-driver (<3.18)

beniwohli:

Running the full matrix, an issue with old versions of cassandra
was revealed. This should fix the issue


beniwohli:
jenkins run the full test matrix

beniwohli:
merging so we can run the full matrix manually on master

PR #768, opened 2020-03-23
Windows Cobertura coverage

cachedout:

Enables Cobetura coverage for Windows builds.

Also comments out a failing test.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -289,6 +289,12 @@ pipeline {
           for(vector in matrixDump) {
             unstash("coverage-${vector}")
           }
+          // Windows coverage converge
+          readYaml(file: '.ci/.jenkins_windows.yml')['windows'].each { v ->
+            unstash(
+              name: "coverage-${v.VERSION}-${v.WEBFRAMEWORK}"
+            )
+          }
           sh('python3 -m coverage combine && python3 -m coverage xml')
           cobertura coberturaReportFile: 'coverage.xml'
         }
@@ -439,6 +445,13 @@ def generateStepForWindows(Map v = [:]){
             installTools([ [tool: "python${majorVersion}", version: "${env.VERSION}" ] ])
             bat(label: 'Install tools', script: '.\\scripts\\install-tools.bat')
             bat(label: 'Run tests', script: '.\\scripts\\run-tests.bat')
+            script{
+              stash{
+                name: "coverage-${v.VERSION}-${v.WEBFRAMEWORK}",
+                includes: ".coverage.${v.VERSION}.${v.WEBFRAMEWORK}",
+                allowEmpty: false
+              }
+            }
           }
         } catch(e){
           error(e.toString())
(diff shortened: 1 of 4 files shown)


cachedout:
Hrm. Nothing in life is easy, apparently:

```
[2020-03-23T11:10:00.163Z] No source for code: '/var/lib/jenkins/workspace/thon_apm-agent-python-mbp_PR-768/src/github.com/elastic/apm-agent-python/C:\Users\jenkins\workspace\thon_apm-agent-python-mbp_PR-768\src\github.com\elastic\apm-agent-python\elasticapm\__init__.py'.

[2020-03-23T11:10:00.163Z] Aborting report output, consider using -i.
```

Even if we were to archive up the entire Windows build checkout and then restore it into the Jenkins workspace, I have my doubts that Python would do the right thing and follow the path correctly. Ex:

```
/tmp/c:
❯ mkdir \\Users

/tmp/c:
❯ cd /tmp/c:\Users
-bash: cd: /tmp/c:Users: No such file or directory
```

And in Python:

```
❯ python
Python 3.6.6 (default, Nov 25 2019, 12:53:29)
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir("/tmp/c:\Users")
  File "<stdin>", line 1
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 7-8: truncated \UXXXXXXXX escape
```

I'm happy to jump down this rabbithole, but first I wanted to gather thoughts from folks about whether this is worth the investment in time and any other ideas about how we might mitigate this issue.

cc: @beniwohli and @v1v

beniwohli:
@cachedout `coverage` has a way to [map paths](https://coverage.readthedocs.io/en/v4.5.x/config.html#paths), we already use it here: https://github.com/elastic/apm-agent-python/blob/9360c58b43021dad7eceeacb7c682794955444b5/setup.cfg#L66-L69

Maybe adding something like `C:\Users\jenkins\workspace\*\src\github.com` to that list might help.

beniwohli:
Looks like I got the path slightly wrong. Can you try

    C:\Users\jenkins\workspace\*\src\github.com\elastic\apm-agent-python

?

cachedout:
@beniwohli I believe this is ready to go now.

Jenkinsfile
```diff
@@ -439,6 +445,13 @@ def generateStepForWindows(Map v = [:]){
             installTools([ [tool: "python${majorVersion}", version: "${env.VERSION}" ] ])
             bat(label: 'Install tools', script: '.\\scripts\\install-tools.bat')
             bat(label: 'Run tests', script: '.\\scripts\\run-tests.bat')
+            script{
```
v1v: NIT:

Is this script closure required?

PR #771, opened 2020-03-24
Add public API calls for TraceParent generation

basepi:

Adds public API entries for TraceParent generation. With our [no-framework instructions](https://www.elastic.co/guide/en/apm/agent/python/current/how-the-agent-works.html#how-it-works-no-framework) it's worth exposing the ability to create TraceParent objects for distributed tracing in custom environments.

I decided to expose both `from_headers` and `from_string` even though most custom implementations will just use `from_string`.

### Note

I cannot for the life of me figure out how to get isort to properly skip `elasticapm.__init__.py`. Both VSCode and `pre-commit` appear to run it directly on the absolute file name which overrides the excludes. (I know, because when VSCode wasn't cooperating I made the change in vim, and then `pre-commit` mangled the file too.)

Anyway, I think I have it so that `isort` ran, and pylint is still happy, and semantically the file is still the same.


elasticapm/utils/disttracing.py
```diff
@@ -133,3 +133,29 @@ def __init__(self, **kwargs):
...
+    """
+    This is a wrapper function so we can add traceparent generation to the
+    public API.
+    """
+    return TraceParent.from_string(traceparent_string, tracestate_string=tracestate_string, is_legacy=is_legacy)
+
+
+def trace_parent_from_headers(
+    headers,
+    header_name=constants.TRACEPARENT_HEADER_NAME,
```
beniwohli: I wonder if we need to expose these header names in the public API. They are all standardized in the W3C spec, so I don't see a valid use case to allow users to override them

basepi: Honestly I was torn on whether to expose `from_headers()` at all, but you're right that I can at least simplify it by removing those overrides.

PR #787, opened 2020-04-03
Fix SyntaxWarning on Python 3.8

messense:

Fixes the following warning on Python 3.8

```
/srv/brm-api_staging/venv/lib/python3.8/site-packages/elasticapm/contrib/flask/__init__.py:136: SyntaxWarning: "is" with a literal. Did you mean "=="?
  if self.logging or self.logging is 0:  # noqa F632
```

### DIFF
--- a/elasticapm/contrib/flask/__init__.py
+++ b/elasticapm/contrib/flask/__init__.py
@@ -30,6 +30,7 @@
 
 
 from __future__ import absolute_import
+import logging
 
 import flask
 from flask import request, signals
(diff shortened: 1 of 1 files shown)


elasticapm/contrib/flask/__init__.py
```diff
@@ -133,7 +133,7 @@ def init_app(self, app, **defaults):
             self.client = make_client(self.client_cls, app, **defaults)
 
         # 0 is a valid log level (NOTSET), so we need to check explicitly for it
-        if self.logging or self.logging is 0:  # noqa F632
+        if self.logging or self.logging == 0:
```
beniwohli, line 136: could we instead compare with `logging.NOTSET`? That would be more explicit and also suppresses the `SyntaxWarning`

```suggestion
        if self.logging or self.logging is logging.NOTSET:
```

(`logging` needs to be added to imports for this to work)

messense: Addressed.

beniwohli:
jenkins test this please

beniwohli:
@messense thanks for updating the PR. Unfortunately, our CI seems to have some technical difficulties at the moment. I'll merge your PR on Monday

PR #790, opened 2020-04-05
introduce enabled/recording settings

beniwohli:
Introduce `recording`/`enabled` as per https://github.com/elastic/apm/issues/92#issuecomment-519752096


beniwohli:
@basepi the "[spec](https://github.com/elastic/apm/issues/92#issuecomment-519752096)" (a.k.a. random comment amidst lengthy discussion) says

> Introduce `enabled` config: true by default, setting to false will completely disable the agent, including instrumentation and remote config polling. This can only be set locally to the agent (e.g. in code, config file, environment variable), and is interpreted once at agent initialisation time; once disabled, disabled for the process's lifetime.

That's the only place I'm aware of that defines behavior of these new settings. Is there any other place I missed?

basepi:
Oh! I missed that. I thought enabled was a remote config option as well. We always talk about `enabled`/`recording` together in the remote config issue: https://github.com/elastic/apm/issues/213

That certainly makes things easier, thanks for the explanation.

basepi:
@beniwohli if this is ready let's get it merged on Monday and cut a new release early next week.

PR #791, opened 2020-04-06
Add instrumentation when using Starlette / FastAPI

HenrikOssipoff:
## Add instrumentation when using Starlette / FastAPI

After been puzzled for ages trying to get Starlette set up, not understanding why it didn't really add instrumentation for anything but the request itself, it seems this was never added.

I haven't contributed to this project before, but I assume you're accepting community PRs.

I've looked through the other framework implementation tests, and can't see any tests for this in those, so I haven't touched those.

What should this be categories as in the change log?


HenrikOssipoff:
I realise now that I missed that an issue should be created before submitting a PR; let me know if I should create a separate issue to track this. Sorry!

PR #808, opened 2020-04-23
ci(dependabot): configuration file

v1v:
If we would like to enable the dependabot automation process to bump the version of the packages automatically, then it's required this particular file.

### How does it work?

There is a weekly pipeline that will care for all the projects with this particular file `ci/dependabot.yml` and if so it will run the validation and PR creations.

### For instance

I ran the below command locally

```bash
docker run -ti \
    --rm -e GITHUB_ACCESS_TOKEN=${GITHUB_ACCESS_TOKEN} \
    -e PROJECT_PATH=v1v/apm-agent-python \
    -e PACKAGE_MANAGER=pip docker.elastic.co/observability-ci/dependabot 
NOTE: Inheriting Faraday::Error::ClientError is deprecated; use Faraday::ClientError instead. It will be removed in or after version 1.0
Faraday::Error::ClientError.inherited called from /src/dependabot-script/vendor/ruby/2.6.0/gems/octokit-4.14.0/lib/octokit/middleware/follow_redirects.rb:14.
Fetching pip dependency files for v1v/apm-agent-python
Parsing dependencies information
Done
...
```

### Actions

- [ ] Verify if this particular dependabot is a good candidate to be enabled in this project.
- [ ] Assign?

### DIFF
--- a/.ci/dependabot.yml
+++ b/.ci/dependabot.yml
@@ -1,0 +1,5 @@
+### The list of GitHub accounts that will be used to assign the PRs to (no space and comma separated)
+assign: ''
+
+### The package manager to be used (no space and comma separated if more than one package)
+manager: pip


beniwohli:
Generally speaking, this is not super useful for our use case. We only have 2 dependencies (3 on Python 2.7), all of them are un-pinned (this is so we don't interfere with whatever version of these libraries the user requires in their project).

As for test requirements, for most of them we are either

 * purposefully using old versions, for matrix tests (e.g. to test older Django versions)
 * can't upgrade to newer versions of the libraries due to them not supporting Python 2.7 anymore (e.g. `pytest`)

We might want to take another look once we dropped Python 2.7 support, but then we'd have to somehow limit it to https://github.com/elastic/apm-agent-python/blob/master/tests/requirements/reqs-base.txt

v1v:
I'll close this PR now for the time being, as we are still gathering details about what's the tool to be used.

PR #812, opened 2020-04-27
Add instrumentation for pymemcache

### LINKED ISSUE #706: Add instrumentation for pymemcache (memcache)
Recently released v3.0 https://github.com/pinterest/pymemcache

Similar to python-memcached in that its pure python, but with more features and, apparently, better perf.

basepi:

Adds instrumentation support for [`pymemcache`](https://github.com/pinterest/pymemcache)

## Related issues

Closes #706

### DIFF
--- a/tests/instrumentation/pymemcache_tests.py
+++ b/tests/instrumentation/pymemcache_tests.py
@@ -111,14 +111,11 @@ def test_pymemcache_pooled_client(instrument, elasticapm_client):
         "PooledClient.set",
         "PooledClient.get",
         "PooledClient.get_many",
-        "Client.set",
-        "Client.get",
-        "Client.get_many",
     }
 
     assert {t["name"] for t in spans} == expected_signatures
 
-    assert len(spans) == 7
+    assert len(spans) == 4
 
 
 @pytest.mark.integrationtest
--- a/tests/instrumentation/pymemcache_tests.py
+++ b/tests/instrumentation/pymemcache_tests.py
@@ -1,0 +1,150 @@
+#  BSD 3-Clause License
+#
+#  Copyright (c) 2019, Elasticsearch BV
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#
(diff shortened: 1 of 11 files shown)


tests/instrumentation/pymemcache_tests.py
```diff
@@ -0,0 +1,150 @@
...
+
+    expected_signatures = {
+        "test_pymemcache",
+        "PooledClient.set",
+        "PooledClient.get",
+        "PooledClient.get_many",
+        "Client.set",
+        "Client.get",
+        "Client.get_many",
+    }
```
basepi, line 117: Interestingly, no matter which class is used, all the "work" is done in a `Client` class at the bottom of the call chain. Despite this, I decided to instrument `PooledClient` and `HashClient` -- if they misbehave in the code surrounding the `Client` call, I think the user would want to know about it. That said, it does result in somewhat redundant spans.

beniwohli: you could pass `leaf=True` to the `capture_span` calls, that would make it ignore child spans. Not sure if that's what you're alluding to, though

tests/instrumentation/python_memcached_tests.py
```diff
@@ -42,7 +42,7 @@
 
 
 if "MEMCACHED_HOST" not in os.environ:
-    pytestmark.append(pytest.mark.skip("Skipping memchached tests, no MEMCACHED_HOST environment variable set"))
+    pytestmark.append(pytest.mark.skip("Skipping memcached tests, no MEMCACHED_HOST environment variable set"))
```
beniwohli, line 45: :bow:

basepi: I only noticed because I started from a copy-paste of those tests in the first place xD

tests/instrumentation/pymemcache_tests.py
basepi: I decided to use `leaf=True` for `PooledClient`, but not for `HashClient`. The reason for this is that `HashClient` deals with memcached clusters, and thus multiple `Client` objects per operation, while `PooledClient` only calls a single `Client` from the pool per operation, so the additional spans don't tell much.

PR #825, opened 2020-05-14
Update the pid/ppid in transport metadata if they have changed

### LINKED ISSUE #792: Process pid get master pid
**Process pid get master pid**

**Environment**
- OS:  Linux
- Python version: All versions
- Framework and version : All Versons
- APM Server version: All Versions
- Agent version: All Versions

https://github.com/elastic/apm-agent-python/blob/master/elasticapm/base.py#134
```python
        transport_kwargs = {
            "metadata": self._build_metadata(),
            "headers": headers,
            "verify_server_cert": self.config.verify_server_cert,
            "server_cert": self.config.server_cert,
            "timeout": self.config.server_timeout,
            "processors": self.load_processors(),
        }
```

Because we get pid in master process but not in workers after fork

basepi:
## Related issues
Closes #792


basepi:
@beniwohli I can't get those flaky windows tests to go green but this is ready for review.

basepi:
@beniwohli this could still use a review when you have a moment.

elasticapm/transport/base.py
```diff
@@ -232,6 +232,12 @@ def _flush(self, buffer):
     def start_thread(self, pid=None):
         super(Transport, self).start_thread(pid=pid)
         if (not self._thread or self.pid != self._thread.pid) and not self._closed:
+            # Update our pid in the metadata
+            if self._metadata.get("process", {}).get("pid"):
+                self._metadata["process"]["pid"] = pid
+                if hasattr(os, "getppid"):
+                    # Windows python 2.7 was failing
+                    self._metadata["process"]["ppid"] = os.getppid()
```
beniwohli, line 240: As this should only happen seldomly, we could rebuild the whole metadata object, so we can reuse the logic there (e.g. the check if `os.getppid` exists).

```suggestion
            # update metadata to reflect new process information
            if self.client:
                self._metadata = self.client._build_metadata()
```

beniwohli: might wanna think about renaming `Client._build_metadata` to `Client.build_metadata` as well if we go down this route.

basepi: Excellent suggestions. Done.

basepi:
@beniwohli I realized when writing the test that if we generate metadata at start_thread time, we would lose any metadata explicitly set in the Transport kwargs. This was problematic for testing, but also begged the question: why would we bother generating it at all before we create the Transport, when it's going to be immediately overwritten when we start the thread?

So now we store any metadata that is sent into the Transport as `self._metadata_kwarg`, and any time we build metadata, we update the result with that dictionary to retain those overrides.

This is a much larger-scale change now, and I'm interested in your opinion.

tests/transports/test_base.py
```diff
@@ -240,3 +240,12 @@ def test_compress_level_sanitization():
     assert DummyTransport(compress_level=None, url="", client=None)._compress_level == 0
     assert DummyTransport(compress_level=-1, url="", client=None)._compress_level == 0
     assert DummyTransport(compress_level=10, url="", client=None)._compress_level == 9
+
+
+@mock.patch("elasticapm.transport.base.Transport.send")
+def test_transport_metadata_pid_change(mock_send, elasticapm_client):
+    transport = Transport(client=elasticapm_client, metadata={})
```
beniwohli, line 247: how about putting something into the `metadata` dict and see if that shows up in `transport._metadata`?

basepi: The only reason I left in the ability to provide overrides is that without that addition, any test that provided metadata would lose that metadata at `transport.start_thread()` time. So, for example, [this test](https://github.com/elastic/apm-agent-python/blob/a00659d3d49e89a1e3854f7663409feb79ae2426/tests/transports/test_base.py#L101) would lose the `{"x": "y"}`. 

Though, now that I'm looking through, I don't think any tests actually rely on those values. That test provides those values just to make sure some metadata is sent, but with the new change it definitely will be sent, as we'll generate the whole metadata dictionary at `start_thread()`. Based on that, I'll probably just remove the metadata kwarg altogether and see if anything breaks.

basepi: Done.

elasticapm/transport/base.py
```diff
@@ -75,14 +74,14 @@ def __init__(
         """
         Create a new Transport instance
 
-        :param metadata: Metadata object to prepend to every queue
+        :param metadata: Metadata overrides for metadata object that will be prepended to every queue
```
beniwohli, line 77: this `:param` can be removed, right?

basepi: Yup. I'll remove it.

basepi:
Just flakey tests failing. Merging.

PR #829, opened 2020-05-20
Fix LoggingHandler Client requirements, plus some docs

basepi:
This is a result of some questions on a support ticket.`LoggingHandler` should not be possible to instantiate without a `Client` object (unless `client_cls` is passed in). Otherwise we can end up with a `LoggingHandler` with a default-config `Client` object that it self-creates and that can cause a lot of hard to debug issues.

The user was also trying to use `ElasticAPM(app, logging=logging.DEBUG)` to increase the log level of the agent, when it's actually meant to ship logs to Elasticsearch. I added some clarification to the docs.

### DIFF
--- a/docs/flask.asciidoc
+++ b/docs/flask.asciidoc
@@ -134,10 +134,10 @@ apm.capture_message('hello, world!')
 [[flask-logging]]
 ==== Shipping Logs to Elasticsearch
 
-Passing `logging=LEVEL` to the ElasticAPM constructor will make the agent
-automatically ship all log messages from Python's built-in `logging` module
-to Elasticsearch, with the given level or higher. `LEVEL` has to be one of
-the levels defined in the
+Enable automatic log shipping by passing `logging=LEVEL` to the ElasticAPM
+constructor. The agent will then ship all log messages with the given
+log-level or higher, from Python's built-in `logging` module to
+Elasticsearch. `LEVEL` must be one of the levels defined in the
 https://docs.python.org/3/library/logging.html#logging-levels[`logging`]
 module.
 
@@ -168,7 +168,7 @@ For fine-grained control, you can initialize a logging handler and add it,
 just as you would with any other handler.
 
 For information on how to increase the log verbosity from the APM agent, see
-the <<troubleshooting,`troubleshooting docs`>>.
+<<agent-logging,troubleshooting agent logging>>.
 
 [source,python]
 ----
--- a/docs/flask.asciidoc
+++ b/docs/flask.asciidoc
@@ -132,11 +132,14 @@ apm.capture_message('hello, world!')
 
 [float]
 [[flask-logging]]
-==== Logging
+==== Shipping Logs to Elasticsearch
 
-Passing `logging=LEVEL` to the ElasticAPM constructor will make the agent automatically log all log messages from Python's built-in `logging` module,
-with the given level or higher.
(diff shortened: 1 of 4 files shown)


elasticapm/handlers/logging.py
```diff
@@ -44,22 +44,20 @@
...
-            self.client = kwargs.pop("client")
-        else:
-            self.client = client(*args, **kwargs)
 
+        if not self.client:
+            client_cls = kwargs.pop("client_cls", None)
+            if client_cls:
+                self.client = client_cls(*args, **kwargs)
+            else:
+                raise ValueError("LoggingHandler requires a Client instance. No Client was received.")
```
beniwohli, line 60: I wonder if this constitutes a backwards incompatible change. Code that didn't raise an error now does. Could we maybe issue a warning instead and change it to raising an error in 6.0?

Also, can you add a test for this case?

docs/flask.asciidoc
```diff
@@ -163,6 +167,9 @@ For shipping of less urgent logs, we recommend to use {filebeat-ref}[filebeat].
 For fine-grained control, you can initialize a logging handler and add it,
 just as you would with any other handler.
 
+For information on how to increase the log verbosity from the APM agent, see
+the <<troubleshooting,`troubleshooting docs`>>.
```
bmorelli25, line 171:
```suggestion
<<agent-logging,troubleshooting agent logging>>.
```

docs/flask.asciidoc
```diff
@@ -132,11 +132,14 @@ apm.capture_message('hello, world!')
...
 
-Passing `logging=LEVEL` to the ElasticAPM constructor will make the agent automatically log all log messages from Python's built-in `logging` module,
-with the given level or higher.
-`LEVEL` has to be one of the levels defined in the https://docs.python.org/3/library/logging.html#logging-levels[`logging`] module.
+Passing `logging=LEVEL` to the ElasticAPM constructor will make the agent
+automatically ship all log messages from Python's built-in `logging` module
+to Elasticsearch, with the given level or higher. `LEVEL` has to be one of
+the levels defined in the
+https://docs.python.org/3/library/logging.html#logging-levels[`logging`]
+module.
```
bmorelli25, line 142: I struggled with the initial paragraph. Does this seem cleaner to you? If it doesn't, that's fine. Also, these line breaks are probably not correct.

```suggestion
Enable automatic log shipping by passing `logging=LEVEL` to the ElasticAPM constructor.
The agent will then ship all log messages with the given log-level or higher,
from Python's built-in `logging` module to Elasticsearch.
to Elasticsearch. `LEVEL` must be one of the levels defined in the https://docs.python.org/3/library/logging.html#logging-levels[`logging`] module.
```

basepi: Yes, I liked your suggestion, though it had a typo so I just applied it manually. Thanks!

elasticapm/handlers/logging.py
```diff
@@ -44,22 +45,26 @@
...
+            if client_cls:
+                self.client = client_cls(*args, **kwargs)
             else:
-                raise ValueError(
-                    "The first argument to %s must be a Client instance, "
-                    "got %r instead." % (self.__class__.__name__, arg)
+                # In 6.0, this should raise a ValueError
+                warnings.warn(
+                    "LoggingHandler requires a Client instance. No Client was "
+                    "received. This will result in an error starting in v6.0",
```
beniwohli, line 64: just a small nit, but I'd explicitly say "v6.0 of the agent" to make it clear that we mean the agent version, not stack version or similar

PR #832, opened 2020-05-23
Respect explicit transaction names over generated names

qeternity:

Favors explicit transaction names over generated ones


beniwohli:
Hi @qeternity 

We should already prefer an explicitly set transaction name over the "calculated" one, and there's a test for that:

https://github.com/elastic/apm-agent-python/blob/a06ba1217bee51eb55d7907ee6d76d1035c5f574/tests/instrumentation/transactions_store_tests.py#L363-L375

In fact, that test fails with your changes applied. Can you give a usage example where the agent doesn't behave as you would expect?

qeternity:
Hi @beniwohli  - I think the behavior you identify as intended is what I am arguing is not intuitive. If I pass in a name at end_transaction, then I would expect the transaction to have that name...which it does not. Our exact use case is changing the instrumentation of Django. We graphql in both Koa and Django backends, but the python does not have graphql implementation (graphene to be specific) so we have to monkeypatch the instrumentation in order to set a sensible transaction name so that everything doesn't just appear as POST /graphql.

beniwohli:
@qeternity ok, I get the confusion, thanks for clarifying. Unfortunately, this behaviour is somewhat necessary to allow modifying the transaction name using [`elasticapm.set_transaction_name()`](https://www.elastic.co/guide/en/apm/agent/python/current/api.html#api-set-transaction-name) during the transaction. We also can't really change this behaviour at this point, as it would constitute a breaking change.

Would it be possible for you to use `set_transaction_name()` e.g. using a Django middleware? That way, you wouldn't need to use monkeypatching.

I'll close this PR for now, as we cannot merge it in the current form.

PR #834, opened 2020-05-23
Do not install flask and requets when installing starlette

HenrikOssipoff:
Stops installing dependencies that aren't required, when installing with the `starlette` flag.

I'm not sure if there's a reason, but I haven't been able to find one. I noticed that when installing using `apm-agent[starlette]`, both `Flask` and `requests` were also installed, which seemed unnecessary.
I can't see any place where requests is actually used, but there was a reference to flask in a test. That test does seem flawed though, since it seems to mock `TraceParent` from the flask package, instead of the starlette package.

Not much, but it saves two dependencies that shouldn't be needed when installing.

Note that I haven't been able to run tests locally (it seems like there's a big setup required), so sorry in advance if the packages were actually required somehow, and the CI freaks out :)

### DIFF
--- a/tests/contrib/asyncio/starlette_tests.py
+++ b/tests/contrib/asyncio/starlette_tests.py
@@ -166,7 +166,7 @@ def test_exception(app, elasticapm_client):
 def test_traceparent_handling(app, elasticapm_client, header_name):
     client = TestClient(app)
     with mock.patch(
-        "elasticapm.contrib.flask.TraceParent.from_string", wraps=TraceParent.from_string
+        "elasticapm.contrib.starlettte.TraceParent.from_string", wraps=TraceParent.from_string
     ) as wrapped_from_string:
         response = client.get(
             "/",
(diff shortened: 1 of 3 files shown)


tests/contrib/asyncio/starlette_tests.py
```diff
@@ -166,7 +166,7 @@ def test_exception(app, elasticapm_client):
 def test_traceparent_handling(app, elasticapm_client, header_name):
     client = TestClient(app)
     with mock.patch(
-        "elasticapm.contrib.flask.TraceParent.from_string", wraps=TraceParent.from_string
+        "elasticapm.contrib.starlettte.TraceParent.from_string", wraps=TraceParent.from_string
```
beniwohli, line 169: Typo :)

```suggestion
        "elasticapm.contrib.starlette.TraceParent.from_string", wraps=TraceParent.from_string
```

HenrikOssipoff:
@beniwohli This is what I get for using a Macbook Pro 🤷‍♂️ Thank you for the review, should be fixed now!

beniwohli:
Thanks for the fix @HenrikOssipoff!

PR #838, opened 2020-05-27
Catch and log any exceptions while closing the transport

### LINKED ISSUE #823: APM agent crashes in Celery at random times during the day
I'm not sure how to reproduce but I've had this issue in production for several months. The crash happens randomly during the day, 20-50 times (we run 400k+ tasks per day) 

The crash is:

```
TransportException: HTTP 400: {"accepted":0,"errors":[{"message":"gzip: invalid header"}]}

  File "elasticapm/transport/base.py", line 208, in _flush
    self.send(data)
  File "elasticapm/transport/http.py", line 102, in send
    raise TransportException(message, data, print_trace=print_trace)
```

- OS: Ubuntu 18.04
- Python version: 3.6
- Framework and version: Django 2.2 Celery 4.4.2
- APM Server version: 7.6.0
- Agent version: 5.3.2

basepi:
Previously these would show up as big tracebacks in the logs which
were ugly, considering all sorts of things can go wrong when suddenly
closing the transport (such as when a celery worker terminates)

Fixes #823


elasticapm/transport/base.py
```diff
@@ -257,7 +263,7 @@ def close(self):
         self._closed = True
         self.queue("close", None)
         if not self._flushed.wait(timeout=self._max_flush_time):
-            raise ValueError("close timed out")
+            logger.error("Closing the transport connection timed out.")
```
beniwohli, line 266: I wonder if we additionally also should use a shorter timeout than `max_flush_time`. Maybe 2s?

basepi: Hmmm. I can potentially see the reasoning, but let's assume the user has upped the `api_request_time` because they're on a slow network (or for some other reason). I don't particularly want to lose data due to being impatient at close time.

I think I'm going to leave it, unless you think there's a hole in my reasoning?

basepi:
More flakey tests. Merging.

PR #850, opened 2020-06-03
Support GraphQL

RyanKung:
Support GraphQL

And I'm not sure what `span_type` it should be.

## Related issues
closes #ISSUE


RyanKung:
Just signed Contributor Agreement

beniwohli:
Hi @RyanKung 

this is awesome, thank you very much! We do have some planning work done for GraphQL, but this mostly pertains to the server side. You can find that here: https://github.com/elastic/apm/blob/master/docs/graphql.md. Maybe some of it could be relevant for the client side too.

As for the span type/subtype/action, I'd personally go with external/graphql/query. There is some prior art in the [Node.js agent](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/graphql.js#L44), where they went with db/graphql/execute, though. 

To be honest, I'm not super familiar with GraphQL, so I'll have to read up and play around a bit with it before I can give you a proper review. This is probably going to take a few days, sorry about that!

beniwohli:
Regarding CLA, it still fails because you used a different email address for signing the CLA than what you use for committing. Could you either sign the CLA again with the email address you used for committing, or amend the commits with the email address you signed the CLA with? Sorry about that!

RyanKung:
> Hi @RyanKung
> 
> this is awesome, thank you very much! We do have some planning work done for GraphQL, but this mostly pertains to the server side. You can find that here: https://github.com/elastic/apm/blob/master/docs/graphql.md. Maybe some of it could be relevant for the client side too.
> 
> As for the span type/subtype/action, I'd personally go with external/graphql/query. There is some prior art in the [Node.js agent](https://github.com/elastic/apm-agent-nodejs/blob/master/lib/instrumentation/modules/graphql.js#L44), where they went with db/graphql/execute, though.
> 
> To be honest, I'm not super familiar with GraphQL, so I'll have to read up and play around a bit with it before I can give you a proper review. This is probably going to take a few days, sorry about that!

Thanks for response, this PR is also for server side. And I will try to refine this PR according to the document. But do we really needs to show such detail informations like `lex`, `parse`, `validate`? Maybe just `execute` is good enough?

RyanKung:
@beniwohli Thanks & Merged, According to the document, the TX name should be changed when getting a GraphQL request. So I have to modify those framework middlewares. Here is what I'm working on now.  It may cost few days.

Support setup TX name when `content-type` = `application/graphgl`
- [x] Support Django middleware
- [x] Support Flask
- [ ] Support aioHttp

Unittest for `framework middlewares`

- [ ] Django
- [ ] Flask
- [ ] aioHttp

beniwohli:
@RyanKung I wonder if instead of having to add support for all frameworks, we could hook into a deeper part of the GraphQL core library. Having played around with the debugger a bit. The `document_from_string` methods of the different backends seem to be a good candidate. We could e.g. write an instrumentation like this (`get_graphql_tx_name` would need to be modified to take a document instead of the unparsed document string)

    import elasticapm
    from elasticapm.instrumentation.packages.base import AbstractInstrumentedModule
    from elasticapm.utils.graphql import get_graphql_tx_name
    
    
    class GraphQLInstrumentation(AbstractInstrumentedModule):
        name = "graphql"
    
        instrument_list = [
            ("graphql.backend.core", "GraphQLCoreBackend.document_from_string"),
            ("graphql.backend.cache", "GraphQLCachedBackend.document_from_string"),
        ]
    
        def call(self, module, method, wrapped, instance, args, kwargs):
            graphql_document = wrapped(*args, **kwargs)
            transaction_name = get_graphql_tx_name(graphql_document.document_ast)
            elasticapm.set_transaction_name(transaction_name)
            return graphql_document

Something similar would need to be done for the new graphql-core package, maybe here? https://github.com/graphql-python/graphql-core/blob/master/src/graphql/execution/execute.py#L245

elasticapm/contrib/flask/__init__.py
```diff
@@ -186,6 +187,16 @@ def request_started(self, app):
...
             rule = request.url_rule.rule if request.url_rule is not None else ""
             rule = build_name_with_http_method_prefix(rule, request)
+            try:
+                if request.headers.get("Content-Type", "") == "application/graphql":
+                    if request.method == "GET":
+                        op, query = request.query_string.decode().split("=")
+                        rule += " GraphQL %s" % get_graphql_tx_name(query, op)
+                    if request.method == "POST":
+                        rule += " GraphQL %s" % get_graphql_tx_name(request.data.decode())
+                assert False
```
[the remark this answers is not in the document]
RyanKung: Fixed

elasticapm/instrumentation/packages/graphene.py
```diff
@@ -0,0 +1,68 @@
...
+#  DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+#  SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+#  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+#  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from elasticapm.instrumentation.packages.base import AbstractInstrumentedModule
+from elasticapm.traces import capture_span
+
+class GrapheneInstrumentation(AbstractInstrumentedModule):
```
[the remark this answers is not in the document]
RyanKung: Fixed

RyanKung:
@beniwohli Wow, change TX name in hooks is simpler than modify all framework support. I'll try it, Thanks. Maybe I should create a new PR for clear?

elasticapm/instrumentation/packages/graphql.py
```diff
@@ -0,0 +1,92 @@
...
+    ]
+
+
+    def get_graphql_tx_name(self, graphql_doc):
+        op = graphql_doc.definitions[0].operation
+        fields = graphql_doc.definitions[0].selection_set.selections
+        return "%s %s" % (op.upper(), "+".join([f.name.value for f in fields]))
+
+
+    def call(self, module, method, wrapped, instance, args, kwargs):
```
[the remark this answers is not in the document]
RyanKung: fixed

RyanKung:
@beniwohli For now this PR is only support `graphene2`&`graphql-core2`.

I tried to support `graphql-core3`, but got some wired behavior when adding more than two methods of `graphql.execution.executor.ExecutionContext` to `instrument_list` such as:

```python
class GraphQLExecutorInstrumentation(AbstractInstrumentedModule):
    name = "graphql"

    Instrument_list_2 = [
      ...
    ]

    instrument_list_3 = [
        # For graphql-core 3
        ("graphql.execution.execute", "ExecutionContext.execute_operation"),
        ("graphql.execution.execute", "ExecutionContext.execute_fields"),
    ]

    instrument_list = instrument_list_2 + instrument_list_3


    def call_graphql_2(self, module, method, wrapped, instance, args, kwargs):
       ...

    def call_graphql_3(self, module, method, wrapped, instance, args, kwargs):
        name = "GraphQL"
        info = ""

        if method == "ExecutionContext.execute_operation":
            print(method)
        if method == "ExecutionContext.execute_fields":
            print(method)

        with capture_span(
                "%s.%s" % (name, info),
                span_type="external",
                span_subtype="graphql",
                span_action="query"
        ):

    def call(self, module, method, wrapped, instance, args, kwargs):
        if method in dict(self.instrument_list_2).values():
            return self.call_graphql_2(module, method, wrapped, instance, args, kwargs)
        if method in dict(self.instrument_list_3).values():
            return self.call_graphql_3(module, method, wrapped, instance, args, kwargs)
```

Only one of those method will be executed, and the whole result will be `None` on unittest:

```python
@pytest.mark.integrationtest
def test_fetch_data(instrument, elasticapm_client):
    query_string = "{succ{yeah},err{__typename}}"

    schema = graphene.Schema(query=Query)

    elasticapm_client.begin_transaction("transaction.test")
    with capture_span("test_graphene", "test"):
        result = schema.execute(query_string)
        assert result.data == {"succ": {"yeah": "hello world"}, "err": {"__typename": "Error"}}
```

```bash
E           AssertionError: assert None == {'err': {'__typename': 'Error'}, 'succ': {'yeah': 'hello world'}}
E             +None
E             -{'succ': {'yeah': 'hello world'}, 'err': {'__typename': 'Error'}}
```

RyanKung:
Hi, @beniwohli 

I think graphene3 is in beta stage, so maybe we can only support 2.0 version for now. Graphene3&graph-core3 supporting can be another pull request.

RyanKung:
Hi @beniwohli 

Thanks for reply.

>     * only capture top level graphql executions (the `GraphQL QUERY allProducts` in this case)

GraphQL query may contains complex/composed queries. A execution may cross a lot of Objects, so I think only top level is not enough for showing complex details.

>     * or not capture executions that take very little time. Capturing spans does have some overhead, and if the operation itself is only fractions of a millisecond, we're adding an undue percentage of overhead.

I added a list of basic type, such as `int`, `str` or `float`  (https://github.com/elastic/apm-agent-python/pull/850/files#diff-a0bf214d1c9543ef09f978bf6e29eda7R55).

 It may avoid capture executions that take very little time. How do you think about it?

jamesthomasdavidson:
Is there anything I can do to contribute towards this? We are going to move our GraphQL API out of develop and into production in a few weeks, and having native APM support would be invaluable.

tests/requirements/reqs-graphene-2.txt
```diff
@@ -0,0 +1,2 @@
+graphene>2
```
basepi, line 1: We should either make this `graphene>2,<3` or we should rename this file to `reqs-graphene-newest.txt` to match with the rest of the requirement files.

RyanKung: Fixed

basepi:
Just flakey test failures. Merging.

beniwohli:
@jamesthomasdavidson if you have the time, it would be great if you could give the current master branch a try and see if it fulfills your needs :)

elasticapm/instrumentation/packages/graphql.py
```diff
@@ -0,0 +1,107 @@
...
+class GraphQLBackendInstrumentation(AbstractInstrumentedModule):
+    name = "graphql"
+
+    instrument_list = [
+        ("graphql.backend.core", "GraphQLCoreBackend.document_from_string"),
+        ("graphql.backend.cache", "GraphQLCachedBackend.document_from_string"),
+    ]
+
+    def get_graphql_tx_name(self, graphql_doc):
+        op = graphql_doc.definitions[0].operation
```
ekampf, line 99: There is a bug here - the first item in the definition is not necessarily an operation.
It can be a fragment in which case you'll get an exception here.

@RyanKung

ekampf: (in fact, client libraries like Apollo send fragments first and operation last so this will always fail. Its best to iterate the list and find the operation then assume its first\last)

beniwohli: @ekampf thanks for the heads up! Would you mind opening a new issue for this?

RyanKung: @ekampf Maybe like this? Although I'm not sure how to test that....

```py
    def get_graphql_tx_name(self, graphql_doc):
        op_def = [
            i for i in graphql_doc.definitions
            if type(i).__name__=="OperationDefinition"
        ][0]
        op = op_def.operation
        fields = op_def.selection_set.selections
        return "GraphQL %s %s" % (op.upper(), "+".join([f.name.value for f in fields]))
```

RyanKung: @ekampf related PR here: https://github.com/elastic/apm-agent-python/pull/881

PR #855, opened 2020-06-08
print error message when submit fails in Django test command

beniwohli:

Due to a refactor in the past, the Django test command did
not print an error anymore when sending a message to the
APM Server failed. Added a fix and test.


basepi:
Er, just kidding, I totally misread the intent of this on my first review. Ignore my misplaced comment about `else:`

PR #860, opened 2020-06-14
Config sanitize field names

### LINKED ISSUE #841: Make SANITIZE_FIELD_NAMES constant configurable for custom headers
It would be nice to have the option to extend the list of field names that are considered when sanitizing HTTP headers. This is especially useful if you have custom headers with sensitive data.

A function that lets a user provide header(s) that extend the base `SANITIZE_FIELD_NAMES` set, such that using the `sanitize_http_headers` processor will also sanitize the headers passed by the user.

1. The simplest workaround could just be to write a custom processor and sanitize custom headers yourself, but this feels like a misuse of the library. If there's a function meant to sanitize headers, makes sense to use it.
2. It's also potentially possible to write a custom processor that wraps `sanitize_http_headers` and monkey patch `SANITIZE_FIELD_NAMES` to include the custom headers you want sanitized. Again, doesn't seem ideal.

**Additional context**

AumitLeon:
Makes `SANITIZE_FIELD_NAMES` a configurable option such that users can specify custom fields they want masked.

From the testing code, it looks like the base sanitize code already handles some wildcard-like features (i.e, passing `some-` as a field to be sanitized would mask fields that match `*some-*`).
 
Example usage: 
```python
ELASTIC_APM = {
    'APP_NAME': '<APP-NAME>',
    'SECRET_TOKEN': '<SECRET-TOKEN>',
    'PROCESSORS': (
        'path.to.my_processor',
        'elasticapm.processors.sanitize_stacktrace_locals',
        'elasticapm.processors.sanitize_http_request_cookies',
        'elasticapm.processors.sanitize_http_headers',
        'elasticapm.processors.sanitize_http_wsgi_env',
        'elasticapm.processors.sanitize_http_request_querystring',
        'elasticapm.processors.sanitize_http_request_body',
    ),
   'SANITIZE_FIELD_NAMES': (
        'some-sensitive-field-1',
        'should-be-hidden'
    ),
}
```

I wanted to get a review before updating the docs.  I had some time and was peeking around the code and decided to take a crack at this and see if it would work out. Please do let me know if this approach is overly simplistic or if I'm missing some other context.

## Related issues
closes #841


basepi:
I apologize for our slowness here @AumitLeon -- It's been a crazy week. I probably won't get to this until next week, but know that I haven't forgotten about it!

basepi:
I think your solution is pretty solid, @AumitLeon. However, I don't like the fact that the sanitize field names will be configurable, but the final list isn't actually stored in the Config object. With the increased focus on remote configuration and the like, I think that abstraction layer might cause problems later.

I'm leaning towards just having the list live in the config, using the existing `SANITIZE_FIELD_NAMES` list as the default for the new setting. This means that users will have to reproduce this list as part of their config if they want to retain the defaults (and add new items to the list). This isn't ideal, but it does allow complete user control over the list, which I like. (Maybe someone wants `sessionid` to not be masked for some reason?)

The nice thing about this is that we can then lose the new class that was managing the complete list. The Config object is the single source of truth.

@beniwohli what do you think?

tests/processors/tests.py
```diff
@@ -177,15 +187,17 @@ def test_sanitize_http_response_cookies(http_test_data):
     )
 
 
-def test_sanitize_http_headers(http_test_data):
+@pytest.mark.parametrize(
+    "elasticapm_client, expected",
+    [
+        ({"sanitize_field_names": "some-value"}, {**_base_expected_headers, "some-value": processors.MASK}),
+        ({"sanitize_field_names": "some-"}, {**_base_expected_headers, "some-value": processors.MASK}),
+        ({"sanitize_field_names": "other-val"}, {**_base_expected_headers, "some-value": "some-secret-value"}),
```
basepi, line 195: Unfortunately, the syntax `{**mydict, "foo": "bar"}` isn't valid in python2. You'll probably just want to keep the `"some-value"` part of the dictionary in the parametrize call and then do a `.update(_base_expected_headers)` in the test itself.

beniwohli:
I agree with @basepi. We have a similar "problem" with `processors`, and added a `warning` note in the docs to use the pre-configured list of processors in addition to custom ones:

https://github.com/elastic/apm-agent-python/blob/40a1303b76b6c6060e339260fac79d8bf52ecc0b/docs/configuration.asciidoc#processors

basepi:
@AumitLeon Keep me posted if you need any guidance for making those changes.

AumitLeon:
Thanks for that great feedback! 

I'm wondering what the most appropriate way to pass the `Config` object to the processors is to access the `sanitize_field_names` list. Any thoughts there?

basepi:
The processors already have a copy of `client`. The config is stored in `client.config`. Then you'll just have to pass the sanitize fields list from that config into the `_sanitize` function. Does that make sense?

AumitLeon:
That does make sense -- what I'm having some trouble with is figuring out how the reconcile this change with the `varmap` function. I'm not quite sure how to pass the `_sanitize` the field names (`client.config.sanitize_field_names`)

basepi:
Oh, good call, on my cursory look I missed the `varmap` usage.

I think I would probably update varmap to take `**kwargs`, and change its usage to executing `func(key_name, value, **kwargs)`. Looks like there are about 11 uses of `varmap` in the codebase -- all of the varmap-called functions will need to be updated to accept `**kwargs`. Then the uses of `varmap` with `_sanitize` can just send in a new variable, `sanitize_field_names`.

Does that help?

elasticapm/processors.py
```diff
@@ -327,4 +320,4 @@ def _process_stack_frames(event, func):
     if "log" in event and "stacktrace" in event["log"]:
         for frame in event["log"]["stacktrace"]:
             func(frame)
-    return event
+    return event
```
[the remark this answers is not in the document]
AumitLeon: oops. Forgot to run the pre commit setup. Rebased and fixed now.

AumitLeon:
@basepi  that was super helpful! 

I've made the relevant changes, but now a single test seems to be failing: `tests/middleware/tests.py::test_error_handler`

The trace: 
```
========================================================================================================= FAILURES =========================================================================================================
____________________________________________________________________________________________________ test_error_handler ____________________________________________________________________________________________________

self = <elasticapm.middleware.ElasticAPM object at 0x7fb6accec510>, environ = {'HTTP_HOST': 'localhost:80', 'PATH_INFO': '/an-error', 'QUERY_STRING': 'foo=bar', 'REQUEST_METHOD': 'GET', ...}
start_response = <function test_error_handler.<locals>.<lambda> at 0x7fb6acc7fb00>

    def __call__(self, environ, start_response):
        try:
>           for event in self.application(environ, start_response):

elasticapm/middleware.py:53: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

environ = {'HTTP_HOST': 'localhost:80', 'PATH_INFO': '/an-error', 'QUERY_STRING': 'foo=bar', 'REQUEST_METHOD': 'GET', ...}, start_response = <function test_error_handler.<locals>.<lambda> at 0x7fb6acc7fb00>

    def example_app(environ, start_response):
>       raise ValueError("hello world")
E       ValueError: hello world

tests/middleware/tests.py:41: ValueError

During handling of the above exception, another exception occurred:

elasticapm_client = <tests.fixtures.TempStoreClient object at 0x7fb6acd1e750>

    def test_error_handler(elasticapm_client):
        middleware = ElasticAPM(example_app, client=elasticapm_client)
    
        request = webob.Request.blank("/an-error?foo=bar")
        response = middleware(request.environ, lambda *args: None)
    
        with pytest.raises(ValueError):
>           list(response)

tests/middleware/tests.py:51: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
elasticapm/middleware.py:57: in __call__
    self.handle_exception(exc_info, environ)
elasticapm/middleware.py:73: in handle_exception
    handled=False,
elasticapm/base.py:231: in capture
    self.queue(ERROR, data, flush=not handled)
elasticapm/base.py:265: in queue
    self._transport.queue(event_type, data, flush)
tests/fixtures.py:234: in queue
    data = self._process_event(event_type, data)
elasticapm/transport/base.py:175: in _process_event
    data = processor(self, data)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

client = <tests.fixtures.DummyTransport object at 0x7fb6acd1e190>
event = {'context': {'custom': {}, 'request': {'env': {'SERVER_NAME': 'localhost', 'SERVER_PORT': '80'}, 'headers': {'host': '...e': 'tests/middleware/tests.py', 'function': 'example_app', ...}], ...}, 'id': '8c188231f39569a3e2b84a3c057a90ac', ...}

    @for_events(ERROR, TRANSACTION)
    def sanitize_http_headers(client, event):
        """
        Sanitizes http request/response headers
        :param client: an ElasticAPM client
        :param event: a transaction or error event
        :return: The modified event
        """
        # request headers
        try:
            headers = event["context"]["request"]["headers"]
            event["context"]["request"]["headers"] = varmap(
>               _sanitize, headers, sanitize_field_names=client.config.sanitize_field_names
            )
E           AttributeError: 'DummyTransport' object has no attribute 'config'

elasticapm/processors.py:154: AttributeError
```
I don't really have enough context into the overall testing infrastructure right now to understand why this test might be failing -- any thoughts?

AumitLeon:
Also, I didn't pass `client.config.sanitize_field_names` to `varmap` for this function yet:
https://github.com/elastic/apm-agent-python/blob/93faa68b57744f7ae3e38001ec41029a5e5c6a65/elasticapm/processors.py#L90-L104

When I do, a bunch of tests fail.

basepi:
Hmm. I'm going to have to look at this closer tomorrow. It almost looks like `client` is a Transport object, rather than a `Client` object. I'll pull your fork tomorrow and see if I can figure out what's going on.

basepi:
Looks like there was a bug. The Transport was sending `self`, instead of `self.client` to the processors. https://github.com/elastic/apm-agent-python/pull/860/commits/1f75ce36fd3b518dfd4327cb6410dcd599c3a31d fixes that, and I took the liberty of pushing to your branch. All of the unit tests now pass locally for me, we'll see what they do on jenkins.

I bet you can now make your `sanitize_stacktrace_locals` change now.

AumitLeon:
Awesome -- looks like that change helped! Thanks so much for that help! I just pushed 50131d134159eca31365925ba4269194293013f9 that has the full changes to all the processors.

If things look good to you, I can update some more of the processor tests and the docs to tidy up this change.

elasticapm/processors.py
```diff
@@ -267,7 +262,12 @@ def mark_in_app_frames(client, event):
     return event
 
 
-def _sanitize(key, value):
+def _sanitize(key, value, **kwargs):
+    if kwargs:
```
basepi, line 266:
```suggestion
    if "sanitize_field_names" in kwargs:
```

elasticapm/processors.py
```diff
@@ -91,15 +84,14 @@ def remove_stacktrace_locals(client, event):
 def sanitize_stacktrace_locals(client, event):
     """
     Sanitizes local variables in all frames
-
-    :param client: an ElasticAPM client
+    :param client: an ElasticAPM clientclear
```
basepi, line 87:
```suggestion
    :param client: an ElasticAPM client
```

basepi:
jenkins run the tests please

AumitLeon:
Happy to help with this. Thank you for all of the awesome feedback and help with debugging -- it was super helpful and definitely helped make this change more approachable! I pushed some updates to the tests and docs as well as some cleanup -- let me know if you have any other comments!

basepi:
jenkins run the tests please

PR #867, opened 2020-06-18
Skip new elasticsearch_async tests for <py3.6

basepi:

The old elasticsearch-async tests were skipped on py3.5 due to [this requirements line](https://github.com/elastic/apm-agent-python/blob/e9dcba55185d90601b4a048483c054ea36d108cd/tests/requirements/reqs-elasticsearch-7.txt#L2). However, as of 7.8.0, the main elasticsearch-py library contains async support, so we need an explicit `skipif` for the new tests on py3.5.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,9 +8,9 @@ flake8:
 	flake8
 
 test:
-	if [[ "$$PYTHON_VERSION" =~ ^(3.6|3.7|3.8|nightly|pypy3)$$ ]] ; then \
+	if [[ "$$PYTHON_VERSION" =~ ^(3.7|3.8|nightly|pypy3)$$ ]] ; then \
 		py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT); \
-	elif [[ "$$PYTHON_VERSION" =~ ^3\.5$$ ]] ; then \
+	elif [[ "$$PYTHON_VERSION" =~ ^(3.5|3.6)$$ ]] ; then \
 		py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT) --ignore-glob='*/asyncio/*'; \
 	else \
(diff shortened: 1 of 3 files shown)


beniwohli:
`skipif` doesn't work here because the `SyntaxError` is raised already at parse time AFAIK.

However, we're excluding tests in `asyncio` folders, so I'm not sure why this is getting executed on Python 3.5... 

https://github.com/elastic/apm-agent-python/blob/272fe24242d055224047dd9b6fe21c035d747c9d/Makefile#L13-L14

beniwohli:
ah, I think this is the issue:

https://github.com/elastic/apm-agent-python/blob/5b54ed195c3257c115bac13c58793eb292bac62c/elasticapm/instrumentation/register.py#L66-L72

We still end up importing the 3.6+-only code via instrumentation. I can't quite remember why we didn't use 3.7 as the lower limit instead of 3.5 for asyncio-related instrumentation, since due to the lack of `contextvars`, it won't work anyway with 3.5 or 3.6.

We could either change the `if` to check for >= 3.7 or add a new version check and add the elasticsearch instrumentation there. The second approach would guarantee backwards compatibility, but I'm not sure that is required here.

basepi:
Good call. I did look at the register but had missed the `=` part of that `>=`. I'll do >=3.7

Makefile
```diff
@@ -8,9 +8,9 @@ flake8:
 	flake8
 
 test:
-	if [[ "$$PYTHON_VERSION" =~ ^(3.6|3.7|3.8|nightly|pypy3)$$ ]] ; then \
+	if [[ "$$PYTHON_VERSION" =~ ^(3.7|3.8|nightly)$$ ]] ; then \
 		py.test -v $(PYTEST_ARGS) $(PYTEST_MARKER) $(PYTEST_JUNIT); \
-	elif [[ "$$PYTHON_VERSION" =~ ^3\.5$$ ]] ; then \
+	elif [[ "$$PYTHON_VERSION" =~ ^(3.5|3.6|pypy3)$$ ]] ; then \
```
beniwohli, line 13: it might make sense to add a comment here regarding pypy3 that it should be removed once it is on par with Python 3.7 (currently it's 3.6, not sure what their timeline is for 3.7+)

PR #871, opened 2020-06-26
Enable flakey test reporting

cachedout:

This turns on the flakey test reporter for the apm-agent-python project.


cachedout:
> @cachedout should I wait with merging #870, as that would probably mask a bunch of flaky tests from this seeing it?

I'd say go ahead and merge it.

> Side note, "flaky tests" has about 10x google results vs "flakey tests". Not sure what's grammatically correct though

Yeah, I think this is right. Luckily, I've been pretty inconsistent in my spelling of it thus far so I'm covered either way. ;)

cachedout:
This will fail until we release a new version of the apm pipeline library.

Issue #755, opened 2020-03-12
[Feature] Add instrumentation for asyncpg


basepi:
Ref #740 

https://github.com/MagicStack/asyncpg

yrik:
We need this feature as well. Do you have any quick guidance how to add it on our own?

basepi:
Your closest parallel is probably the `aiopg` instrumentation that @beniwohli added awhile back: https://github.com/elastic/apm-agent-python/pull/668

Keep us posted if you decide to try to tackle that, we can answer questions.

basepi:
Merged in #889

ajay1mg:
how do I enable this for my application that uses asyncpg based Tortoise ORM?

PR #889, opened 2020-07-17
Add asyncpg instrumentation in Elastic APM

odimko:
Adds support for the `asyncpg` database interface library for Python and PostgreSQL by implementing corresponding instrumentation in the APM agent.

## Related issues
https://github.com/elastic/apm-agent-python/issues/755

## Closes issues
https://github.com/elastic/apm-agent-python/issues/755


basepi:
Heya @odimko! Thanks for this!

Please sign [the CLA](https://www.elastic.co/contributor-agreement) when you get a chance. Do you think you'll be able to add some tests here as well?

odimko:
hey there @basepi! Thanks for checking out my PR. it's not ready yet, so more commit will come (together with tests). 
As for the CLA, I [did sign it](https://github.com/elastic/apm-agent-python/pull/889#issuecomment-660097555). Shall I do it once more?

basepi:
Nope, shows signed. Sometimes it just lags a bit. Thanks, this looks awesome!

odimko:
hey @basepi! Could you let me know how I can run CI tests as I am currently getting this error:
`githubPrCheckApproved: The PR is not allowed to run in the CI yet. (Only users with write permissions can do so.)`, thus, I can't check my tests.

odimko:
hi @basepi :) I was wondering if you have had some time to have a look at my comment above :)

sethmlarson:
@odimko FYI @basepi has been on vacation for past few weeks :)

odimko:
thanks for letting me know @sethmlarson :1st_place_medal:

elasticapm/instrumentation/register.py
```diff
@@ -37,6 +37,7 @@
     "elasticapm.instrumentation.packages.jinja2.Jinja2Instrumentation",
     "elasticapm.instrumentation.packages.psycopg2.Psycopg2Instrumentation",
     "elasticapm.instrumentation.packages.psycopg2.Psycopg2ExtensionsInstrumentation",
+    "elasticapm.instrumentation.packages.asyncio.asyncpg.AsyncPGInstrumentation",
```
beniwohli, line 40: Please move this into the "asyncio" list starting at line 67.

Unfortunately, we haven't come up with a smarter way to avoid importing Python 3.5+ only code on 2.7. I guess a naive way would be to skip any instrumentations that have `.asyncio.` in the path. I'll investigate that.

elasticapm/instrumentation/packages/asyncio/asyncpg.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class AsyncPGInstrumentation(AsyncAbstractInstrumentedModule):
+
+    name = "asyncpg"
+
+    instrument_list = [("asyncpg.connection", "Connection._do_execute")]
+
+    async def call(self, module, method, wrapped, instance, args, kwargs):
+        if method == "Connection._do_execute":
```
beniwohli, line 43: This check shouldn't be necessary

.ci/.jenkins_framework.yml
```diff
@@ -36,6 +36,7 @@ FRAMEWORK:
   - mysqlclient-newest
   - aiohttp-newest
   - aiopg-newest
+  - asyncpg-newest
```
beniwohli, line 39: In addition to this, you'll have to add entries into `.ci/jenkins_exclude.yml` for Python 2.7, 3.5 and 3.6, as well as pypy-2 and pypy-3, analogue to the aiopg entries:

https://github.com/elastic/apm-agent-python/blob/b3b08c282dbd4949129736195df38d2258ee7503/.ci/.jenkins_exclude.yml#L259-L269

odimko:
> @odimko this is great, thanks! And sorry for the long turnaround on the review, both @basepi and me were on PTO

hey @beniwohli and @basepi! I have implemented all the requested changes. Please have a look at the PR if it looks good now and let me know if I should improve anything else :)

beniwohli:
@odimko great work on the updates. The tests currently [fails](https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/PR-889/17/pipeline/#step-6723-log-283) because asyncpg isn't able to parse the DSN. Turns out that for asyncpg, the DSN has a different format than psycopg2, using a URL format:

https://magicstack.github.io/asyncpg/current/api/index.html#asyncpg.connection.connect

You should be able to adapt the `dsn()` function in your asyncpg tests easily.

tests/instrumentation/asyncio/asyncpg_tests.py
```diff
@@ -0,0 +1,116 @@
...
+import pytest  # isort:skip
+
+asyncpg = pytest.importorskip("asyncpg")  # isort:skip
+
+import os
+
+from elasticapm.conf import constants
+
+try:
+    from psycopg2 import sql
```
rafalstapinski, line 40: there's no need for this import block, asyncpg does rely on psycopg2

tests/instrumentation/asyncio/asyncpg_tests.py
```diff
@@ -0,0 +1,116 @@
...
+    cur = await conn.cursor()
+
+    # we use a finalizer instead of yield, because Python 3.5 throws a syntax error, even if the test doesn't run
+    def rollback():
+        cur.raw.execute("ROLLBACK")
+        cur.close()
+        conn.close()
+
+    request.addfinalizer(rollback)
+    await cur.execute(
```
rafalstapinski, line 77: asyncpg cursor does not have a raw property or execute method

elasticapm/instrumentation/packages/asyncio/asyncpg.py
```diff
@@ -0,0 +1,50 @@
...
+from elasticapm.contrib.asyncio.traces import async_capture_span
+from elasticapm.instrumentation.packages.asyncio.base import AsyncAbstractInstrumentedModule
+from elasticapm.instrumentation.packages.dbapi2 import extract_signature
+
+
+class AsyncPGInstrumentation(AsyncAbstractInstrumentedModule):
+
+    name = "asyncpg"
+
+    instrument_list = [("asyncpg.connection", "Connection._do_execute")]
```
rafalstapinski, line 40: `Connection.execute` optionally calls the Cython `BaseProtocol.query` method if there are no arguments. I would suggest changing the instrumentation to the higher level `Connection.execute` and `Connection.executemany`

I believe this will cause the `test_Select_Sleep` test to fail as that query will not call `Connection._do_execute`

tests/instrumentation/asyncio/asyncpg_tests.py
```diff
@@ -0,0 +1,116 @@
...
+
+    has_sql_module = True
+except ImportError:
+    # as of Jan 2018, psycopg2cffi doesn't have this module
+    has_sql_module = False
+
+pytestmark = [pytest.mark.asyncpg, pytest.mark.asyncio]
+
+if "POSTGRES_DB" not in os.environ:
+    pytestmark.append(pytest.mark.skip("Skipping asyncpg tests, no POSTGRES_DB environment variable set"))
```
rafalstapinski, line 50: This seems dangerous? We should not skip these tests and instead assert the environment variables are set or just fail with a verbose message

tests/instrumentation/asyncio/asyncpg_tests.py
```diff
@@ -0,0 +1,116 @@
...
+    )
+
+
+@pytest.fixture()
+async def cursor(request):
+    conn = await asyncpg.connect(dsn())
+    cur = await conn.cursor()
+
+    # we use a finalizer instead of yield, because Python 3.5 throws a syntax error, even if the test doesn't run
+    def rollback():
```
rafalstapinski, line 71: This test is excluded from  from Python 3.5 run, would suggest just adding this after `yield cursor` and awaiting the ROLLBACK and close

tests/instrumentation/asyncio/asyncpg_tests.py
beniwohli: @rafalstapinski this is a pattern we use across all our integration tests with databases and other services. Our CI runs them in a matrix test and injects the necessary environment variables "on demand".

tests/instrumentation/asyncio/asyncpg_tests.py
beniwohli: Unfortunately, the code is still parsed in Python 3.5 runs (during pytest's discovery phase), leading to the `SyntaxError` mentioned in the comment.

Once we deprecate Python 3.5, we can change it to the `yield` approach.

odimko:
hey @rafalstapinski! Thanks for your comments :) I will go through them and implement any necessary changes if needed.

rafalstapinski:
@odimko feel free to look through my implementation for any ideas: https://github.com/elastic/apm-agent-python/pull/895

tests/instrumentation/asyncio/asyncpg_tests.py
odimko: could you please elaborate on this one? how exactly does it rely on `psycopg2`? From what I see there are no dependencies on the mentioned package, e.g. [here](https://github.com/MagicStack/asyncpg/search?q=psycopg2&unscoped_q=psycopg2). 
Also I've used [pipgrip](https://github.com/ddelange/pipgrip) to check the full dependency tree and only found that `asyncpg` does not depend on other packages.

rafalstapinski: hah 🤦 type, 
> there's no need for this import block, asyncpg does `not` rely on psycopg2

but yeah my point is: no need to import psycopg2 in this test. it's used in aiopg 1) because of the dependency and 2) because its used to prepare a statement, however asyncpg doesn't have such a concept

it's not a big deal either way, was just pointing out its not necessary here like it is in aiopg instrumentation tests

tests/instrumentation/asyncio/asyncpg_tests.py
odimko: thanks for pointing out!

tests/instrumentation/asyncio/asyncpg_tests.py
odimko: oh, i see. that's a great explanation. thanks for that! I'll remove it.

elasticapm/instrumentation/packages/asyncio/asyncpg.py
odimko: Could you please explain why do you think it would be more beneficial to go to a higher level? How having calls to the Cython `BaseProtocol.query` method, if there are no arguments, would be helpful?

The idea behind my implementation was to go as low as possible to be able to register any queries to the database.

rafalstapinski: `_do_execute` does not catch everything.

take a look at the implementation of `Connection.execute` for example: https://github.com/MagicStack/asyncpg/blob/master/asyncpg/connection.py#L294-L297

if we pass in a fully prepared query, for example `await connection.execute("select * from pg_sleep(0.2);")` we don't end up running `_do_execute` but rather skip immediately to the Cython implementation `BaseProtocol.query` which is not instrumentable by the apm agent

odimko: thanks for explaining!

odimko:
hey @rafalstapinski, thanks a lot for your explanations! it was super useful.

@beniwohli, I've made some changes to the PR, however, I can't figure out why the tests are failing and if they are related to my PR. May I kindly ask you to have a look at it?

odimko:
hey @beniwohli and @basepi! I'm sorry for disturbing you once again, however, I was wondering if you could help me to understand why the tests are failing after my changes. So far the PR seems to be rather completed.

tests/instrumentation/asyncio/asyncpg_tests.py
basepi: Looks like this is a problem with `asyncpg` compared to the other postgres drivers. We need to await both the rollback and the `conn.close()`. Which is a problem, as @beniwohli mentioned before, because python3.5 throws a syntax error.

Python 3.5 is EOL one month from today. We may just wait to finish this PR until that deprecation takes place. Alternatively, because it's so soon, we may jump the gun a little and start that deprecation sooner rather than later so we can get this merged.

basepi:
@odimko Please see my comment in the newly-unresolved discussion above.

tests/instrumentation/asyncio/asyncpg_tests.py
```diff
@@ -0,0 +1,114 @@
...
+    transaction = elasticapm_client.events[constants.TRANSACTION][0]
+    spans = elasticapm_client.spans_for_transaction(transaction)
+
+    assert len(spans) == 1
+    span = spans[0]
+    assert transaction["id"] == span["transaction_id"]
+    assert span["subtype"] == "postgres"
+    assert span["action"] == "query"
+    assert span["sync"] == False
+    assert span["name"] == "SELECT FROM test"
```
rafalstapinski, line 114: This does not seem like desired behavior. We've executed an insert but the report says `SELECT`

odimko: great catch. thanks!

tests/instrumentation/asyncio/asyncpg_tests.py
odimko: thank you for a thorough explanation @basepi and @beniwohli! would it be possible to go ahead and merge this PR given that the implementation itself is correct and the tests pass/would be expected to pass, as the alternative way mentioned by @basepi? if yes, what would I have to change/add/remove? as the features in this PR are rather highly expected on our side.

basepi: I think if @beniwohli is alright with it, I'm going to remove python 3.5 from our matrix tests later this week when I have some time. In the meantime, can you change this test to properly `await` both the rollback and the close as @rafalstapinski recommended? Then when I remove python3.5 the tests should start passing and we can get this merged.

odimko:
@beniwohli and @basepi, hey there :) 
please have a look at the latest changes to the PR, I've switched to use the `yield` approach as suggested (since we are moving towards Python 3.5 deprecation). Is there anything else left to do?

tests/instrumentation/asyncio/asyncpg_tests.py
odimko: done

beniwohli (approved):
This looks great! Thanks a lot for your contribution and sorry about the sometimes slow answer times!

docs/supported-technologies.asciidoc
```diff
@@ -212,6 +212,22 @@ Instrumented methods:
...
+[float]
+[[automatic-instrumentation-db-asyncg]]
+==== asyncpg
+
+Library: `asyncpg` (`>=0.20`)
+
+Instrumented methods:
+
+ * `async.connection.Connection.execute`
+ * `async.connection.Connection.executemany`
```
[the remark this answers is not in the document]
odimko: my bad :) thanks

odimko:
> This looks great! Thanks a lot for your contribution and sorry about the sometimes slow answer times!

No problem at all! Thanks for supporting with it :1st_place_medal:

basepi:
We don't currently have a release date planned, but I expect it will be Soon (tm) since we have a couple of useful features that are currently unreleased (and since we no longer need to include the removal of py3.5 in the release).

odimko:
> We don't currently have a release date planned, but I expect it will be Soon (tm) since we have a couple of useful features that are currently unreleased (and since we no longer need to include the removal of py3.5 in the release).

got it, thanks for explaining!

odimko:
Good morning @basepi and @beniwohli! just checking in to check what is your plan regarding releasing the next version of the APM package. could you please let me know?

basepi:
Very soon. We'd really like to get #927 in, but it needs some test fixes.

odimko:
perfect, thanks for the update!

basepi:
It's out! https://github.com/elastic/apm-agent-python/releases/tag/v5.9.0

odimko:
> It's out! https://github.com/elastic/apm-agent-python/releases/tag/v5.9.0

awesome! thanks a lot!

PR #890, opened 2020-07-20
Fixing eventlet tests

basepi:
TFW you have no idea why the tests are failing...


basepi:
Threw some sleeps in here because I cannot figure out what might have changed in our nightly builds. But even here, the sleeps are just making it fail in a flakey manner across a bunch of different matrix builds. It appears to be limited to py3.6+ and pypy-3.

Anyway, I'm probably going to ditch this problem for a couple of weeks until @beniwohli is around again.

PR #891, opened 2020-07-21
Add sample_rate to transactions and tracestate

### LINKED ISSUE #888: Add sample_rate to transactions
See https://github.com/elastic/apm/pull/307 for details. Need to add the sampling rate to all transactions and spans, and propagate it through via `tracestate`.

### LINKED ISSUE #917: Implement sample_rate precision spec
You can view the spec changes [here](https://github.com/elastic/apm/pull/335).

basepi:

Adds the sample_rate to transactions, and propagates the sample_rate through tracestate

## Related issues
Closes #888 
Closes #917


basepi:
This is code complete. Just need to add a few more tests (this is mostly a note to self since I'll be PTO for awhile):

* Weight empty if incoming traceparent doesn't define it
* Weight 0 for unsampled transaction
* Weight present in transaction payload
* Weight present in span payload

basepi:
This will have to be rewritten for the new spec in https://github.com/elastic/apm/pull/307 once it is finalized.

basepi:
This has been rewritten for the updated (finalized) [spec](https://github.com/elastic/apm/pull/307) and tests have been written as well. It's ready for review.

elasticapm/utils/disttracing.py
```diff
@@ -119,13 +122,86 @@ def merge_duplicate_headers(cls, headers, key):
...
+            otherstate = re.sub(r"es=([^,]*),?", "", self.tracestate)
+            # No validation of `otherstate` required, since we're downstream. We only need to check `es=`
+            # since we introduced it, and that validation has already been done at this point.
+            if otherstate:
+                if otherstate[-1] == ",":
+                    return "{}{}".format(otherstate, elastic_state)
+                else:
+                    return "{},{}".format(otherstate, elastic_state)
+            else:
+                return elastic_state
```
beniwohli, line 172: Should do the same, I believe

```suggestion
            return otherstate.rstrip(",") + elastic_state
```

elasticapm/utils/disttracing.py
```diff
@@ -119,13 +122,86 @@ def merge_duplicate_headers(cls, headers, key):
...
+        """
+        Add key/value pair to the tracestate.
+
+        We do most of the validation for valid characters here. We have to make
+        sure none of the reserved separators for tracestate are used in our
+        key/value pairs, and we also need to check that all characters are
+        within the valid range. Checking here means we never have to re-check
+        a pair once set, which saves time in the _set_tracestate() function.
+        """
+        key = str(key)
```
beniwohli, line 184: do we need to use `compat.string_type` here for 2.7 compatibility? (sorry)

elasticapm/utils/disttracing.py
```diff
@@ -119,13 +122,86 @@ def merge_duplicate_headers(cls, headers, key):
...
+        val = str(val)
+        for bad in (":", ";", ",", "="):
+            if bad in key or bad in val:
+                logger.debug("New tracestate key/val pair contains invalid character '{}', ignoring.".format(bad))
+                return
+        for c in key + val:
```
beniwohli, line 190: depending on how large `val` can become, it might be worth avoiding creating a new string just to throw it away again. Something like

```suggestion
        for c in itertools.chain(key, val)
```

(requires an `import itertools` up top of course)

elasticapm/utils/disttracing.py
```diff
@@ -119,13 +122,86 @@ def merge_duplicate_headers(cls, headers, key):
...
+            if ord(c) < 0x20 or ord(c) > 0x7E:
```
beniwohli, line 191: can you add a comment on what those hex values signify?

beniwohli (changes requested):
Great job!

elasticapm/traces.py
```diff
@@ -325,6 +328,8 @@ def to_dict(self):
             "sampled": self.is_sampled,
             "span_count": {"started": self._span_counter - self.dropped_spans, "dropped": self.dropped_spans},
         }
+        if self.sample_rate is not None:
+            result["sample_rate"] = float(self.sample_rate) if self.sample_rate != "0" else 0
```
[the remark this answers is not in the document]
basepi: In the [spec](https://github.com/elastic/apm/blob/master/specs/agents/tracing-sampling.md#non-sampled-transactions), it explicitly states `0` as opposed to `0.0` (which is what it would be if it were formatted as a float).

In retrospect, I highly doubt it matters. I'm going to omit the special handling and let it be sent as a `0.0` float. Do you anticipate any problems with that @axw ?

elasticapm/utils/disttracing.py
```diff
@@ -119,13 +122,86 @@ def merge_duplicate_headers(cls, headers, key):
...
+
+            tracestate: es=key:value;key:value...,othervendor=<opaque>
+        """
+        if not tracestate:
+            return {}
+        if "es=" not in tracestate:
+            return {}
+
+        ret = {}
+        state = re.match(r"es=([^,]*)", tracestate).group(1).split(";")
```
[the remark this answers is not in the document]
basepi: Good catch, thanks.

elasticapm/utils/disttracing.py
basepi: Almost. Still need to add a comma if `otherstate` is non-empty.

elasticapm/utils/disttracing.py
basepi: `val` should always be <256 characters, but since we're not checking it at this point, but rather at the tracestate string generation, I'll use your suggestion. Thanks.

elasticapm/utils/disttracing.py
basepi: I don't think so. As far as I know `compat.string_types` is only required for `isinstance` checks against strings. For casting to string, `str()` is valid in both py2 and py3. (Please correct me if I'm wrong.)

elasticapm/traces.py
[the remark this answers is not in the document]
axw: Nope, that's fine, I think it's reasonable to expect agents to parse `0.0` the same as `0`.

It would be nice to send as few bytes as necessary, but 2 bytes isn't going to matter in the grand scheme of things.

beniwohli (approved):
Awesome work!

elasticapm/utils/disttracing.py
basepi: Revisited on chat with Beni. Should be `compat.text_type`, because under certain situations py2 can cast to a byte string. Most of the time it wouldn't be an issue, until it was suddenly an issue. :)

Fixed.

basepi:
A change to the spec is in review and this should wait to be merged until it's finalized. (Though I've already implemented the minimum value requirement, which is really the only change to my implementation so far.)

https://github.com/elastic/apm/pull/335

PR #895, opened 2020-08-04
Add asyncpg instrumentation

rafalstapinski:

This PR adds proper instrumentation for `asyncpg`

I've noticed there's another PR open to also add asyncpg instrumentation: https://github.com/elastic/apm-agent-python/pull/889, however it has some implementation issues. Feel free to close this one out, I'd be happy to merge my changes in to the existing PR.

## Related issues
closes https://github.com/elastic/apm-agent-python/issues/755

### DIFF
--- a/.ci/.jenkins_exclude.yml
+++ b/.ci/.jenkins_exclude.yml
@@ -267,6 +267,17 @@ exclude:
     FRAMEWORK: aiopg-newest
   - PYTHON_VERSION: python-3.6
     FRAMEWORK: aiopg-newest
+  # asyncpg
+  - PYTHON_VERSION: python-2.7
+    FRAMEWORK: asyncpg-newest
+  - PYTHON_VERSION: pypy-2
+    FRAMEWORK: asyncpg-newest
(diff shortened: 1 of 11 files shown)


rafalstapinski:
@beniwohli sorry for the @ mention but what would it take to run this PR through to CI? I imagine the `yield` inside the fixture will cause issues with 3.5 discovery, however adding a finalizer for the fixture request is causing tests to stall. Would ideally want to run this through CI (though I don't believe this is isolated locally) and if not will just prefer to spin up a new connection at each test and avoid the fixture.

beniwohli:
Hi @rafalstapinski! No worries about the ping, that's what it is here for :)

If you have Docker locally and run either Linux or MacOS, you should be able to run the same environment as our CI by running something like

    ./tests/scripts/docker/run_tests.sh python-3.5 asyncpg-newest

I also just started a CI build of your branch on our public CI, the results should be available in about half an hour or so.

rafalstapinski:
@beniwohli looks like everything's ✅ ? Anything blocking from getting this merged in?

PR #844, opened 2020-06-01
Add instrumentation for httpx

HenrikOssipoff:
## What does this pull request do?
Adds instrumentation for `httpx.Client` and `httpx.AsyncClient`.

Inspiration taken from the `requests` instrumentation, as the interface is more or less the same.

**TODO**
- [ ] Add tests
- [x] Add PR link in Changelog

## Related issues
closes #739

PR #898, opened 2020-08-06
Add instrumentation for httpx and httpcore

### LINKED ISSUE #739: [Feature] encode/httpx support
A friend and I were chatting recently about async libraries in python. He had nothing but good things to say about the tools in https://github.com/encode

`httpx` seems like a no-brainer, in particular: https://github.com/encode/httpx

basepi:
Thank you to @HenrikOssipoff for doing most of the work on this instrumentation in #844. I opened this PR to finish up some tests and get this merged in.

Adds instrumentation for `httpx.Client` and `httpx.AsyncClient`.

Inspiration taken from the `requests` instrumentation, as the interface is more or less the same.

**TODO**

`httpx`:
- [x] Sync instrumentation
- [x] Async instrumentation
- [x] Sync tests
- [x] Async tests

`httpcore`:
- [x] Sync instrumentation
- [x] Async instrumentation
- [x] ~Sync tests~
- [x] ~Async tests~

## Related issues
Closes #739


basepi:
Looks like I'll need to instrument `httpcore`, which is the default transport option for httpx. I *think* the sync tests will go green at that point.

basepi:
@beniwohli I banged my head against `httpcore` for a day trying to write useful tests there. It's fairly user-hostile and under-documented.

After more research and chatting with @sethmlarson, I've decided not to write `httpcore`-specific tests. As far as either of us can tell, `httpcore` exists only to be the default low level interface for `httpx`, and is not used by any other library. Thus, the only important behavior in our instrumentation is setting the distributed tracing headers, which is covered in our `httpx` tests. I considered trying to test `httpcore` via the `httpx` tests, but because of the leaf span handling I don't get any `httpcore` spans in those tests.

Any thoughts? Are you ok with the lack of `httpcore` tests?

basepi:
Will need to add the outcome if #899 is merged before this one.

elasticapm/instrumentation/packages/asyncio/httpcore.py
```diff
@@ -0,0 +1,120 @@
...
+            extra={"http": {"url": url}, "destination": destination},
+            leaf=True,
+        ) as span:
+            # if httpcore has been called in a leaf span, this span might be a DroppedSpan.
+            leaf_span = span
+            while isinstance(leaf_span, DroppedSpan):
+                leaf_span = leaf_span.parent
+
+            if headers is not None:
+                # It's possible that there are only dropped spans, e.g. if we started dropping spans.
```
beniwohli, line 92: This comment seems confusing, maybe "e.g. if we reached the transaction_max_spans limit"?

(I realize the irony that you copied this comment verbatim from the urllib3 instrumentation and that I wrote it originally :D )

tests/instrumentation/asyncio/httpx_tests.py
```diff
@@ -0,0 +1,136 @@
...
+    assert constants.TRACEPARENT_HEADER_NAME in waiting_httpserver.requests[0].headers
+    trace_parent = TraceParent.from_string(waiting_httpserver.requests[0].headers[constants.TRACEPARENT_HEADER_NAME])
+    assert trace_parent.trace_id == transactions[0]["trace_id"]
+
+    # this should be the span id of `httpx`, not of urllib3
+    assert trace_parent.span_id == spans[0]["id"]
+    assert trace_parent.trace_options.recorded
+
+
+async def test_httpx_instrumentation_via_client(instrument, elasticapm_client, waiting_httpserver):
```
beniwohli, line 78: This test seems to be virtually identical to the test above, except for the fact that it uses a string as url, not a parsed url. The test name indicates that the differentiating factor is that it uses a client, but the first test also uses a client

beniwohli (approved):
LGTM. It would be great if we had a comment somewhere in the code that explains why we need to instrument both the core and the client, though.

tests/instrumentation/asyncio/httpx_tests.py
basepi: Thanks for the catch. Yay copy pasta?

basepi: Ah, yes, it's copy pasta from the sync tests. Async doesn't have an `httpx.get()` equivalent. I was very confused when I went to fix this, accidentally looking at the sync tests.....because the tests are actually different there. :)

basepi:
Alright, review comments implemented, and `outcome` implemented as well. I'll merge this once it goes green.

HenrikOssipoff:
Just wanted to let you know that I'm absolutely thrilled to see this merged - it'll make instrumentation in all of our apps so much better, once we see the next release of the agent. Thanks a bunch!

PR #899, opened 2020-08-11
implement "outcome" property for transactions and spans

beniwohli:
This implements https://github.com/elastic/apm/pull/299.

Additionally, the "status_code" attribute has been added to HTTP spans.

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -285,7 +285,7 @@ class Transaction(BaseSpan):
             start=start,
         )
 
-    def end_span(self, skip_frames=0, duration=None, outcome=None):
+    def end_span(self, skip_frames=0, duration=None, outcome="unknown"):
         """
         End the currently active span
         :param skip_frames: numbers of frames to skip in the stack trace
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -729,6 +729,13 @@ def tag(**tags):
 
 
 def set_transaction_name(name, override=True):
+    """
+    Sets the name of the transaction
+
+    :param name: the name of the transaction
+    :param override: if set to False, the name is only set if no name has been set before
+    :return: None
+    """
     transaction = execution_context.get_transaction()
     if not transaction:
         return
@@ -737,11 +744,50 @@ def set_transaction_name(name, override=True):
 
 
 def set_transaction_result(result, override=True):
+    """
+    Sets the result of the transaction. The result could be e.g. the HTTP status class (e.g "HTTP 5xx") for
+    HTTP requests, or "success"/"fail" for background tasks.
+
+    :param name: the name of the transaction
+    :param override: if set to False, the name is only set if no name has been set before
+    :return: None
+    """
+
     transaction = execution_context.get_transaction()
     if not transaction:
         return
     if transaction.result is None or override:
         transaction.result = result
+
+
+def set_transaction_success():
+    """
+    Marks this transaction as successful. This should only be done at the end of a transaction
+    when the outcome is determined.
+
+    This value is used for error rate calculations.
+
+    :return: None
+    """
+    transaction = execution_context.get_transaction()
+    if not transaction:
+        return
+    transaction.set_success()
+
+
+def set_transaction_failure():
+    """
+    Marks this transaction as failed. This should only be done at the end of a transaction
+    when the outcome is determined.
+
+    This value is used for error rate calculations.
+
+    :return: None
+    """
+    transaction = execution_context.get_transaction()
+    if not transaction:
+        return
+    transaction.set_failure()
 
 
 def get_transaction_id():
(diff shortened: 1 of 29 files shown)


elasticapm/traces.py
```diff
@@ -274,17 +285,22 @@ def begin_span(
...
         :param duration: override duration, mostly useful for testing
+        :param outcome: outcome of the span, either success, failure or unknown
         :return: the ended span
         """
         span = execution_context.get_span()
         if span is None:
             raise LookupError()
 
+        # only overwrite span outcome if it is still unknown
+        if span.outcome == "unknown":
```
basepi, line 301:
```suggestion
        if outcome and span.outcome == "unknown":
```

I assume that the outcome shouldn't be `None` -- if I'm incorrect then you can ignore this suggestion.

beniwohli: I changed the default to "unknown" in 37ea77d, that should take care of this, right?

basepi: Yep!

elasticapm/traces.py
```diff
@@ -709,13 +744,52 @@ def set_transaction_name(name, override=True):
...
+    when the outcome is determined.
+
+    This value is used for error rate calculations.
+
+    :return: None
+    """
+    transaction = execution_context.get_transaction()
+    if not transaction:
+        return
+    transaction.set_success()
```
felixbarny, line 775: It looks like this the outcome may get overridden by the agent after on transaction end.

The spec states
> This value must always take precedence over the automatically determined value.

elasticapm/traces.py
```diff
@@ -709,13 +744,52 @@ def set_transaction_name(name, override=True):
...
+    """
+
     transaction = execution_context.get_transaction()
     if not transaction:
         return
     if transaction.result is None or override:
         transaction.result = result
 
 
+def set_transaction_success():
```
felixbarny, line 763: Are you planning to add an API for spans too?

elasticapm/traces.py
```diff
@@ -709,13 +744,52 @@ def set_transaction_name(name, override=True):
...
+
+    :return: None
+    """
+    transaction = execution_context.get_transaction()
+    if not transaction:
+        return
+    transaction.set_success()
+
+
+def set_transaction_failure():
```
felixbarny, line 778: I'm wondering whether we should offer an API to set the status to unknown.
Is there a particular reason you favored having two different methods instead of one that takes an enum? Is it because we support Python prior to 3.4 that don't have enums?

elasticapm/traces.py
beniwohli: You're right, I remembered something wrong. I thought we check first if `transaction.outcome != "unknown"`, but that happens at a different place. I'll add that check to the `set_success` and `set_failure` setters.

elasticapm/traces.py
beniwohli: Not in this form, no. Spans that are created by auto-instrumentation aren't accessible to the user (at least not in a documented way). Manually created spans have the `set_success()` and `set_failure()` methods.

elasticapm/traces.py
```diff
@@ -709,13 +740,55 @@ def set_transaction_name(name, override=True):
...
+    """
+
     transaction = execution_context.get_transaction()
     if not transaction:
         return
     if transaction.result is None or override:
         transaction.result = result
 
 
+def set_transaction_outcome(outcome=None, http_status_code=None, override=True):
```
felixbarny, line 759: I'm not sure whether we need the override parameter in the public API.
What would be a use case where you wouldn't want to override?
Can there ever be a situation where the agent already set the outcome when the user calls this method?

felixbarny (approved):
LGTM

The only thing I'd add is more docs around the semantics of the individual status values and how they affect error rate calculation. The spec PR has some details on that but having that on the API docs will make that more discoverable. Either add on the `set_transaction_outcome` docs or on the `OUTCOME` constants, plus on the API docs.

elasticapm/traces.py
beniwohli: It's always possible for people to provide framework integrations as external modules, and for those cases it would be useful to have this I think.

Generally, we don't really have a different internal API. What makes an API public (in the sense that we commit to API stability) is the documentation. What we do not document, we consider private, and if enterprising users find the APIs by reading code, they will have to expect breaking changes at any time.

beniwohli:
@felixbarny does https://github.com/elastic/apm-agent-python/pull/899/commits/0124d6873 address your comment on documentation?

felixbarny:
It does!

beniwohli:
@basepi I saw you added a new "unreleased" changelog section in #898. I'll skip that here to avoid merge conflicts. I'll add the entry when you merged the httpx PR

PR #903, opened 2020-08-18
move most of the setup.py metadata and requirements to setup.cfg

beniwohli:
This should make this information easier to discover for tools like Snyk.

Also, it's the new and fresh way to do things :)


beniwohli:
I pushed an example release to https://test.pypi.org/project/elastic-apm/5.8.1.post2/

One pretty nice thing is that we have direct links to documentation, issue tracker, release notes etc.

PR #910, opened 2020-08-28
Rename variable elasticapm to apm

jimit105:

* Rename variable `elasticapm` to `apm` so that the code works for the examples `apm.client.capture_exception()` and `apm.client.capture_message('hello, world!')` in the [usage section](https://www.elastic.co/guide/en/apm/agent/python/master/starlette-support.html#starlette-usage)
* Also, avoids conflicts when `import elasticapm` is also used


beniwohli:
Hey @jimit105. Great catch, thanks for taking the time to open a PR!

PR #911, opened 2020-08-31
added testing utility to capture messages from given logger

beniwohli:

So far, the usage of caplog.at_level in our code base was erroneous. It assumed that when providing a logger, only messages of that logger would appear in caplog.records. Instead, it would only set the level on the logger, but capture all logs with the given level. This could lead to flaky tests if other code would log at that level during the test.

### DIFF
--- a/tests/client/client_tests.py
+++ b/tests/client/client_tests.py
@@ -32,6 +32,7 @@
 
 from __future__ import absolute_import
 
+import logging
 import os
 import platform
 import socket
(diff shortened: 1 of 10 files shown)


tests/client/client_tests.py
```diff
@@ -246,11 +247,11 @@ def test_send_remote_failover_sync(should_try, sending_elasticapm_client, caplog
     should_try.return_value = True
 
     # test error
-    with caplog.at_level("ERROR", "elasticapm.transport"):
+    with capture_from_logger(caplog, logging.ERROR, "elasticapm.transport") as records:
         sending_elasticapm_client.capture_message("foo", handled=False)
     sending_elasticapm_client._transport.flush()
```
basepi, line 252: Perhaps sticking this flush in the `with` block would solve the issues with caplog? It would let the thread catch up, right? If that works, it would probably work for both failing tests.

beniwohli:
closing in favor of #1073

PR #912, opened 2020-09-01
Add support for callbacks on config change

### LINKED ISSUE #908: Remote Config: Add callbacks on config change
The upcoming remote config option for [`log_level`](https://github.com/elastic/apm/pull/332) will require us to be able to execute code when it changes, so we can change the log level in the `elasticapm` logger.

This should be generalized to allow for functions to be called when a remote config option changes.

basepi:

Adds the ability to specify one or more "callback" functions which will be called when a `_ConfigValue` is changed.

## Related issues
closes #908


elasticapm/conf/__init__.py
```diff
@@ -409,9 +440,10 @@ def reset(self):
         """
         Reset state to the original configuration
         """
-        with self._lock:
-            self._version = self._first_version
-            self._config = self._first_config
+        # We use an update rather than just overwriting the _config and _version
+        # to make sure any config values with callbacks are appropriately
+        # processed.
+        self.update(version=self._first_version, **self._first_config._values)
```
basepi, line 446: This is definitely wrong. `_config._values` doesn't map properly to inline config (I think it would with a `.lower()` but I don't want to introduce that requirements) and in any case it won't pass the validators. Need to figured out a better way to process callbacks on a `reset()`.

beniwohli (commented):
Looks like a clean solution!

I was wondering if it would be possible to use validators for the callback functionality, but this is much better.

elasticapm/conf/__init__.py
```diff
@@ -209,6 +276,11 @@ class _ConfigBase(object):
     def __init__(self, config_dict=None, env_dict=None, inline_dict=None):
         self._values = {}
         self._errors = {}
+        self._dict_key_lookup = {}
+        for field, config_value in self.__class__.__dict__.items():
```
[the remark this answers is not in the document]
basepi: Will fix. @beniwohli, why is this not `compat.iteritems()`? (It's copied from the loop a few lines later in `update()`)

basepi:
@beniwohli Something I realized while trying to figure out how to test this today -- callbacks will also be called on initial Config creation, for any non-default values. I think this is probably ok and expected behavior? But it's something to note, and I wondered if you saw any issues with that. I feel like any operation you want to perform on config change probably needs to be performed on initialization as well, if the config value is different than the default value, but I'm not certain.

elasticapm/conf/__init__.py
```diff
@@ -53,10 +53,51 @@ def __init__(self, msg, field_name):
...
+        The default for this config value if not user-configured.
+    required
+        Whether this config value is required. If a default is specified,
+        this is a redundant option (except to ensure that this config value
+        is specified if a default were ever to be removed).
+
+    Note that _ConfigValues and any inheriting classes must implement __set__
+    and __get__. The calling instance will always be a _ConfigBase descendant
+    and the __set__ and __get__ calls will access `instance._values[self.dict_key]`
+    to get and set values.
```
beniwohli, line 84: 😍

elasticapm/conf/__init__.py
```diff
@@ -418,11 +499,23 @@ def update(self, version, **config):
     def reset(self):
         """
         Reset state to the original configuration
+
+        Note that because ConfigurationValues can have callbacks, we need to
+        note any differences becween the original configuration and the most
```
beniwohli, line 504:
```suggestion
        note any differences between the original configuration and the most

File: tests/config/tests.py [hunk 5]
```diff
@@ -315,3 +315,20 @@ class MyConfig(_ConfigBase):
     c = MyConfig({"this_one_is_required": 1})
     c.update({"this_one_isnt": 0})
     assert not c.errors
+
```
beniwohli, line 318: Could you add a test for the `reset()` behavior?

beniwohli (approved):
Awesome work! Thanks for taking the time to write those docs!

PR #927, opened 2020-09-21
fix handling of required keys in configuration

### LINKED ISSUE #921: Verify that `service.name` is required for data to be sent
We got reports of data from the Python agent making it to the APM Server without `service.name` being set. `service_name` is a required configuration, and we set `disable_send` to `True` if it is missing, so there might be a bug somewhere there.

https://github.com/elastic/apm-agent-python/blob/1d8f1835e6c0ea207b8321ec63ddd0a34784b296/elasticapm/conf/__init__.py#L255

https://github.com/elastic/apm-agent-python/blob/1d8f1835e6c0ea207b8321ec63ddd0a34784b296/elasticapm/base.py#L106-L110

cc @simitt

beniwohli:
If a key wasn't provided, we didn't check if it was required or not.
This led to a problem that if no `service_name` was configured,
we'd send invalid data to APM Server.

fixes #921

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,8 @@ endif::[]
 [float]
 ===== Bug fixes
 
+* Fix validation of config to properly require `required` config items. {pull}927[#927]
+
 
 [[release-notes-5.x]]
 === Python Agent version 5.x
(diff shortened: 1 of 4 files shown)


basepi (approved):

basepi:
This looks great, though as I expected it appears to have caused a bunch of test failures that will need to be fixed.

beniwohli:
Yes, I'm working on fixing the tests. One in particular is a tough nut. It relied on the buggy behavior we had, so I'm looking for ways to test the same code without relying on that bug...

PR #930, opened 2020-09-22
Add missing technologies to support matrix

basepi:
While compiling a list of supported technologies, I realized that some were missing from our [supported technologies doc](https://www.elastic.co/guide/en/apm/agent/python/current/supported-technologies.html).

This PR adds those missing technologies.

### DIFF
--- a/docs/supported-technologies.asciidoc
+++ b/docs/supported-technologies.asciidoc
@@ -90,6 +90,21 @@ be supported.
 The Python APM agent comes with automatic instrumentation of various 3rd party modules and standard library modules.
 
 [float]
+[[automatic-instrumentation-scheduling]]
+=== Scheduling
+
+[float]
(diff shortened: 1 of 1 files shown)


docs/supported-technologies.asciidoc
```diff
@@ -89,6 +89,21 @@ be supported.
...
+[float]
+[[automatic-instrumentation-scheduling-celery]]
+===== Celery
+
+We support these Celery versions:
+
+* 3.x
+* 4.x
+
+Celery tasks will be recorded with Django and Flask only.
```
beniwohli, line 105:
```suggestion
Celery tasks will be recorded automatically with Django and Flask only.
```

Also, it might make sense to document how the instrumentation can be hooked up manually, in a separate PR of course.

beniwohli (approved):
Awesome!

PR #938, opened 2020-10-01
Upgrade to isort v5.5.4

basepi:
While I was fixing `isort` behavior in my IDE for the umpteenth time, I noticed that we're running a fairly old version of isort. After reading the [guide to upgrading to 5.x](https://pycqa.github.io/isort/docs/upgrade_guides/5.0.0/) I realized we were missing out on some real goodness in terms of `isort`'s ability to automatically categorize imports and the like.

You'll note that the changes in actual behavior were pretty minimal, despite removing most of our isort config.

The biggest change to moving to a largely default config was that django is now not its own section. You'll see from the diff that that's actually not a very big deal, but I'm open to reintroducing that as a section if we want to.

@beniwohli Thoughts?

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,6 @@
 repos:
-# seed-isort-config disabled for now, see https://github.com/asottile/seed-isort-config/issues/10
-#-   repo: https://github.com/asottile/seed-isort-config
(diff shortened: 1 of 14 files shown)


beniwohli (approved):
LGTM!

PR #939, opened 2020-10-05
Extend asyncpg instrumentation in Elastic APM

stj:

Issue #889 added instrumentation for asyncpg, though only `execute` and
`executemany` were instrumented. This extends the instrumentation to
include `fetch`, `fetchval` and `fetchrow` methods.

### DIFF
--- a/elasticapm/instrumentation/packages/asyncio/asyncpg.py
+++ b/elasticapm/instrumentation/packages/asyncio/asyncpg.py
@@ -46,6 +46,9 @@ class AsyncPGInstrumentation(AsyncAbstractInstrumentedModule):
     instrument_list = [
         ("asyncpg.connection", "Connection.execute"),
         ("asyncpg.connection", "Connection.executemany"),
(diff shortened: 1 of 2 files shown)


basepi (approved):

basepi:
This looks excellent, thanks!

beniwohli:
@stj just a quick follow up on this: some time ago, I experimented with instrumenting [`fetchone()`](https://www.python.org/dev/peps/pep-0249/#fetchone), [`fetchmany()`](https://www.python.org/dev/peps/pep-0249/#fetchmany) and [`fetchall()`](https://www.python.org/dev/peps/pep-0249/#fetchall) in the DB-API2 based database instrumentations. The result was that it would create way too many spans (depending on the access pattern, up to one span per row). Seeing as every span comes with some overhead (both runtime overhead in the app and storage overhead in ES), I decided against implementing it.

I'm just wondering if we might have a similar issue here. Did you already have the chance to gather some experience with running these instrumentations in production?

PR #940, opened 2020-10-05
Include whole body for elasticsearch-py queries

basepi:
See https://discuss.elastic.co/t/apm-not-showing-aggregation/250762
for context.

Basically, by only including the query, we risk missing very useful
pieces of the query for diagnosing issues -- `aggs` is the big problem
here but I don't see why the user wouldn't want the entire query every
time.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -43,6 +43,8 @@ endif::[]
 [float]
 ===== Bug fixes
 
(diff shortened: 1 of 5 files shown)


sethmlarson (approved):
LGTM. Since the JSON structure is not relevant to machines, more relevant to users this isn't going to break anything I'm assuming :)

PR #943, opened 2020-10-08
WIP - Elastic except hook 681

### LINKED ISSUE #681: Django 2.2 uncaught exceptions not being sent
**Describe the bug**: ...
I am not sure if it is a bug, because reading the documentation I was not able to find any explicit information talking about uncaught exceptions with Django (only with Flask), but all non handled exceptions are not being sent to APM server. However, running the test command (`python manage.py elasticapm test`) the event is sent successfully 

- OS: Linux (Docker image python:3.7.4)
- Python version: 3.7.4
- Framework and version [e.g. Django 2.1]: Django 2.2
- APM Server version: v7.5.0
- Agent version: 5.3.2

**How to reproduce**
In order to tests the uncaught exceptions I did the following tests:
* Manually inserted an `raise ZeroDivisionError('APM TEST')` in one of my functions outside any `try/catch/`
* Executed `raise ZeroDivisionError('APM TEST')` using python shell (`python manage.py shell`)

**Additional context**

`settings.py`

```
INSTALLED_APPS = [
    'django.contrib.admin',
    'django.contrib.auth',
    'django.contrib.sessions',
    'django.contrib.messages',
    'django.contrib.contenttypes',
    'django.contrib.staticfiles',
    'elasticapm.contrib.django',
    ...
]


MIDDLEWARE = [
    'elasticapm.contrib.django.middleware.TracingMiddleware',
    'elasticapm.contrib.django.middleware.Catch404Middleware',
    ...
]


ELASTIC_APM = {
    'SERVICE_NAME': 'MYSERVICE',
    'SECRET_TOKEN': 'SECRETTOKEN,
    'SERVER_URL': 'SERVERURL',  # noqa
    'DEBUG': True
}

```

*Running check command*
```
root@26cb7a8c0615:/usr/src/app# python manage.py elasticapm check
Service name is set, good job!

SERVER_URL https://APM_ADDR:443 looks fine

DEBUG mode is disabled! Looking good!

Tracing middleware is configured! Awesome!

Looks like everything should be ready!
```

romulorosa:

This PR introduces a new config options which allows users to use elastic as the exception hook for it's project.  The idea behind is basically replace the default Python exception hook for elastics exception hook, which means for all unhandled exceptions captured during the execution will be handle/sent through elastic's handler.

**Behavior**

- The configuration is off by default
- If the configuration is set to True, then
    - If the exception hook used is the Python's default one, so the Elastic's exception hook will replace it
    - If the exception hook used is a custom one (written by user) then the Elastic's exception hook will not replace it but at the and will also be called. So, at the end of the execution both elastic exception hook and the custom one should be called

## Related issues
closes #681


basepi:
I want to get @beniwohli's eyes on this as well, so we'll review it next week when he's back from PTO.

elasticapm/base.py
```diff
@@ -559,6 +565,13 @@ def _get_stack_info_for_trace(
             locals_processor_func=locals_processor_func,
         )
 
+    def _excepthook(self):
+        exec_info = sys.exc_info()
+        if self.was_default_excepthook_overridden_by_user:
+            self.original_excepthook(*exec_info)
```
beniwohli, line 571: Could it make sense to call this in a `try/finally` block to ensure that our `capture_exception` method is called even if the original exception hook raises an exception for some reason?

beniwohli (commented):
This is great, thanks a lot! We considered doing this for a long time but have never gotten around to do it. The implementation looks great!

Your tests also look to be comprehensive, but they introduce a couple of new patterns, like using fixtures for mocking, and avoiding to call the actual `capture_exception()` method. So far, we chose to exercise the agent code as much as possible, and only mocking the transport part (the code that serializes and sends the data). That has the benefit of allowing to inspect the collected data, e.g. you could assert that the `handled` flag is indeed set to `False`. But it's not really what one would call unit tests.

I guess your tests are philosophically closer to actual unit tests, which definitely is a plus. I'm not sure which version I prefer. @basepi, what do you think?

elasticapm/base.py
basepi: Seems reasonable to me. 👍

elasticapm/base.py
```diff
@@ -559,6 +565,13 @@ def _get_stack_info_for_trace(
             locals_processor_func=locals_processor_func,
         )
 
+    def _excepthook(self):
```
basepi, line 568: Should this function also be calling `sys.__excepthook__` if it wasn't overridden? We probably want to maintain the default behavior of the exception system, after capturing the data for ourselves.

basepi (commented):
I think I'm ok with the unit testing approach in this case, @beniwohli, especially since it's additive. We're not replacing integration tests with unit tests, but rather just adding some more tests with a limited testing surface area.

Will this create a bunch of duplicate exception captures? Replacing `sys.excepthook` means we will see every exception that is raised, including ones we capture in other places in the code, unless I'm missing something...

elasticapm/base.py
romulorosa: That makes completely sense. Good catch :+1:

romulorosa (commented):
@basepi Regarding your question, the exception may be captured by more the one exception handler. The agent handler and also by the default or custom exception handler. When I first thought about this feature I was looking something similar to what Sentry does, which captures the unhandled exceptions as well. Back then I just had the elastic stack available and in my opinion this would be a good addition.

beniwohli:
@basepi 
> Will this create a bunch of duplicate exception captures? Replacing `sys.excepthook` means we will see every exception that is raised, including ones we capture in other places in the code, unless I'm missing something...

No, `sys.excepthook` is only called for uncaught exceptions that bubble all the way up and will cause the interpreter to exit. From the [docs](https://docs.python.org/3/library/sys.html#sys.excepthook):

> When an exception is raised and uncaught, the interpreter calls sys.excepthook with three arguments, the exception class, exception instance, and a traceback object. In an interactive session this happens just before control is returned to the prompt; in a Python program this happens just before the program exits.

So for frameworks like Django and Flask, this should almost never be invoked, as they bring their own exception handling and don't let exceptions bubble up. It is however super useful for things like scripts that don't bring their own exception handling.

elasticapm/base.py
romulorosa: I was thinking about this modification and ended up with a question. What If the exception hook defined by the user raises an exception, what should be done in such case? Shall the code capture it and handle using `self.capture_exception` or shall let it bubble up?

Example:
```
import sys

sys.excepthook = lambda x, y, z: 1/0

raise Exception()
```

This piece of code results in:
```
>>> import sys
>>> 
>>> sys.excepthook = lambda x, y, z: 1/0
>>> 
>>> 
>>> raise Exception()
Error in sys.excepthook:
  File "<stdin>", line 1, in <lambda>
ZeroDivisionError: division by zero

Original exception was:
  File "<stdin>", line 1, in <module>
Exception
```

So, in this case it is also possible to catch this exception and call elastic handler, similar to the following snippet 
```
    def _excepthook(self):
        exec_info = sys.exc_info()
        try:
            self.original_excepthook(*exec_info)
        except Exception:
            self.capture_exception(sys.exc_info(), handled=False)
        finally:
            self.capture_exception(exec_info, handled=False)
```

Any thoughts here?

romulorosa (commented):
@basepi @beniwohli I pushed the changes suggested by you. There is only one question that came to me about what should be done in case of having an exception raised by the original exception hook. I talked more about this case in the `Conversation`, you can check it there,

elasticapm/base.py
basepi: Hmm. I guess we should probably collect both? Seems unlikely to be an edge case we ever see (I don't think the average user is replacing excepthook) but probably worth covering?

basepi: @beniwohli do you have thoughts here?

beniwohli: IMO the current implementation is fine, as it will capture both exceptions. But I agree that it is probably an edge case.

basepi:
I took the liberty of pushing some lint fixes so that `black` would pass. Note that if you install [`pre-commit`](https://pre-commit.com/) it should help catch these things for you.

romulorosa:
Is there anything else which I can help here?

basepi:
I don't know why it keeps aborting, since I'm triggering them manually. I'm working on getting the tests passing and we'll get this merged.

PR #944, opened 2020-10-08
[docs] Flesh out documentation about creating transactions

### LINKED ISSUE #633: [DOCS] Document tracing of non-web transactions
Ref: https://discuss.elastic.co/t/how-to-trace-non-web-transactions/207052/5

basepi:
Also moved the documentation to a (hopefully) more intuitive location
and added a link from the set-up docs.

## Related issues
Closes #633


beniwohli (approved):
This is great!

One thing that caught my eye was the capitalization of "Client", "Span" and "Transaction". Some of it is from content you moved, so it's not new :) just wondering if that is something we should do. We could also wrap them with backticks to emphasize that they are objects. @bmorelli25, any input?

docs/custom-instrumentation.asciidoc
```diff
@@ -1,12 +1,21 @@
...
 Elastic APM instruments a variety of libraries out of the box, but sometimes you
 need to know how long a specific function took or how often it gets
 called.
 
-You can apply the `@elasticapm.capture_span()` decorator to achieve exactly that.
-It can also be used as a context manager in the following way:
+Assuming you're using one of our <<set-up, supported frameworks>>, you can
+apply the `@elasticapm.capture_span()` decorator to achieve exactly that. (If
+you're not using a supported framework, see
+<<instrumenting-custom-code-transactions, Creating New Transactions>>)
```
bmorelli25, line 15:
```suggestion
apply the `@elasticapm.capture_span()` decorator to achieve exactly that. If
you're not using a supported framework, see
<<instrumenting-custom-code-transactions, Creating New Transactions>>.
```

docs/custom-instrumentation.asciidoc
```diff
@@ -1,12 +1,21 @@
...
+
+`elasticapm.capture_span` can be used either as a decorator, or as a context
```
bmorelli25, line 17:
```suggestion
`elasticapm.capture_span` can be used either as a decorator or context
```

docs/custom-instrumentation.asciidoc
```diff
@@ -48,4 +57,50 @@ async def coffee_maker(strength):
...
+[[instrumenting-custom-code-transactions]]
+==== Creating New Transactions
+
+It's important to note that `elasticapm.capture_span` only works if there is
+an existing transaction. If you're not using one of our <<set-up, supported
+frameworks>>, you need to create a `Client` object and begin and end the
+transactions yourself. You can even utilize the agent's
+<<automatic-instrumentation,automatic instrumentation>>!
+
+In order to collect the Spans generated by the supported libraries, you need
```
bmorelli25, line 72:
```suggestion
To collect the Spans generated by the supported libraries, you need
```

docs/custom-instrumentation.asciidoc
```diff
@@ -48,4 +57,50 @@ async def coffee_maker(strength):
...
+to invoke `elasticapm.instrument()` (just once, at the initalization stage of
```
bmorelli25, line 73:
```suggestion
to invoke `elasticapm.instrument()` (just once, at the initialization stage of
```

docs/custom-instrumentation.asciidoc
```diff
@@ -48,4 +57,50 @@ async def coffee_maker(strength):
...
+    client = elasticapm.Client(service_name="foo", server_url="https://example.com:8200")
+    elasticapm.instrument()  # Only call this once, as early as possible.
+    client.begin_transaction(transaction_type="script")
+    main()
+    client.end_transaction(name=__name__, result="success")
+----
+
+Note that you don't need to do anything to send the data -- the `Client` object
+will handle that before the script exits. Additionally, the Client object should
+be treated as a singleton -- you should only create one instance, and store/pass
```
bmorelli25, line 105: "store/pass around that.." is a dependent clause, so there's no comma needed here.

```suggestion
be treated as a singleton -- you should only create one instance and store/pass
```

bmorelli25 (approved):
Thanks! A few comments.

And yes! If you can update span and transaction to be lowercase here, that'd be great. We're trying to be more consistent with these, but there is a lot of old documentation that still hasn't needs to be updated.

docs/custom-instrumentation.asciidoc
bmorelli25: On second thought, I think I like the `as a` here. Regardless, with correlative conjunctions (in this case, "either" ... "or") there's no need to use a comma before the second conjunction.

```suggestion
`elasticapm.capture_span` can be used either as a decorator or as a context
```

PR #952, opened 2020-10-27
add Python 3.10 beta/RC builds to the matrix

beniwohli:
Also, promote Python 3.9 from RC to release, it has been released a couple weeks ago


beniwohli:
3.10 is currently blocked on https://github.com/python/cpython/pull/20333, as `bdist_wheel` fails

basepi (approved):

Issue #954, opened 2020-10-27
Add a new function to the public api returning the current trace parent header.


diegosperes:
**Is your feature request related to a problem? Please describe.**

Currently I am instrumenting a kafka consumer that receive a message containing the trace parent header to after begin a new transaction passing the `TraceParent` instance on `begin_transaction` method of `Client` class. The issue remains in the ability of easily get the current trace parent header in the service that is publishing a kafka message, for that is necessary to dig into this project, understanding how trace parent header is generated.

**Describe the solution you'd like**
Create a public function that retrieve the trace parent header of the current transaction.

**Describe alternatives you've considered**
Improve the documentation to explicitly show how to generate a trace parent header of the current transaction.

```
from elasticapm.traces import execution_context

transaction = execution_context.get_transaction().
trace_parent_header = transaction .trace_parent.to_string() if transaction and transaction.trace_parent else None
```

basepi:
Implemented in #956

PR #956, opened 2020-10-27
Add public API to retrieve trace parent header.

diegosperes:

Add a new public API to easily return the trace parent header of the current transaction, currently it's necessary to dig into the project and understand how it works.

Code necessary to get the current trace parent header without the new function.

```
from elasticapm.traces import execution_context

transaction = execution_context.get_transaction().
trace_parent_header = transaction .trace_parent.to_string() if transaction and transaction.trace_parent else None
```

## Related issues

https://github.com/elastic/apm-agent-python/issues/954

### DIFF
--- a/docs/api.asciidoc
+++ b/docs/api.asciidoc
@@ -174,6 +174,16 @@ client.begin_transaction('processors', trace_parent=parent)
 
  * `headers`: (*required*) HTTP headers formed as a dictionary.
 
+[float]
+[[api-traceparent-from-headers]]
+===== `elasticapm.get_trace_parent_header()`
+
+Return the string representation of the current transaction TraceParent object:
+
+[source,python]
+----
+elasticapm.get_trace_parent_header()
+----
 
 [float]
 [[api-other]]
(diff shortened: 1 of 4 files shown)


diegosperes:
Couldn't find which place should I add the unit test, does someone has any suggestion? I think there isn't a test case for public API.

beniwohli:
Hey @diegosperes. This looks like a nice addition to the public API! You're right, the testing for public APIs isn't very extensive, but there are a few here that might serve as inspiration: 

https://github.com/elastic/apm-agent-python/blob/65e8b7c7becd85476dda5870781042eab2e2e2ac/tests/instrumentation/transactions_store_tests.py#L363-L442

It would also be great if you could add an entry in the traceparent docs here: https://github.com/elastic/apm-agent-python/blob/master/docs/api.asciidoc#traceparent.

diegosperes:
@beniwohli thanks for guide me, I added a few missing test for the public API also a new entry for this function in the traceparent docs.

beniwohli (approved):
This is great work, thanks a lot!

@basepi ok with merging this?

basepi (approved):

PR #957, opened 2020-10-28
Starlette: use route name as transaction name if available

### LINKED ISSUE #833: Consider using route information instead of URL for Starlette transactions
We're slowly started using Starlette for some smaller applications due to its simplicity and async capabilities. We also happen to use Elastic APM for our monitoring, but we've found that the agent works quite different from other agents, such as Django or Flask.

The default for Django, is "HTTP Method + path.to.view". This can be changed for Django 2.2+  using the `django-transaction-name-from-route` setting, to instead use the route that was matched against. For example `GET /users/<int:user_id>/`.

For Flask, the default seems to be the that of method + route, e.g.: `GET /users/<int:user_id/`.

For Starlette however, information about the matched route is not actually available on the request, and for that reason (I assume), the full URL is instead used as the transaction name.
In the example above, it would mean that every time a user with a different URL was hit, it would generate two separate transaction entries. For some of our views, that would mean 1+ milllion transaction names for the same view.

However, it is somewhat possible to extract the route by mimicking what Starlette itself does. On the request, the Application itself is available, and from that route router + routes. This means it's possible to loop over them, and match their regular expression against the current request.

1) I realise this isn't pretty in any way, shape or form - but it would make instrumentation of Starlette applications a lot more useable
2) I also realise it's backwards incompatible - but perhaps this could be solved in the same way that it's done with Django?

What would your thought be on such a change?

For us at our workplace, we've actually have one major of our main applications ported to Starlette, but once we saw what was happening to the transaction names, we had to revert the deployment. We'd love to have it running, so we'd probably be interested in providing a patch for it as well.

**EDIT**
Just wanted to add this here, since it's a discussion of making this type of information available in the middleware system: https://github.com/encode/starlette/issues/685
It also links another repo for timing of ASGI that basically does what I describe here, by looping over all routes and comparing them to the current URL.

beniwohli:
Unfortunately, we need to do some matching here that Starlette does
too, but currently isn't exposed. If/when https://github.com/encode/starlette/pull/804
is merged, we can re-use the route from there.

closes #833

### DIFF
--- a/elasticapm/contrib/starlette/__init__.py
+++ b/elasticapm/contrib/starlette/__init__.py
@@ -35,6 +35,7 @@ import starlette
 from starlette.middleware.base import BaseHTTPMiddleware, RequestResponseEndpoint
 from starlette.requests import Request
 from starlette.responses import Response
+from starlette.routing import Match
 from starlette.types import ASGIApp
 
 import elasticapm
@@ -178,7 +179,8 @@ class ElasticAPM(BaseHTTPMiddleware):
         self.client.begin_transaction("request", trace_parent=trace_parent)
 
         await set_context(lambda: get_data_from_request(request, self.client.config, constants.TRANSACTION), "request")
-        elasticapm.set_transaction_name("{} {}".format(request.method, request.url.path), override=False)
+        transaction_name = self.get_route_name(request) or request.url.path
+        elasticapm.set_transaction_name("{} {}".format(request.method, transaction_name), override=False)
 
     async def _request_finished(self, response: Response):
         """Captures the end of the request processing to APM.
@@ -192,3 +194,15 @@ class ElasticAPM(BaseHTTPMiddleware):
 
         result = "HTTP {}xx".format(response.status_code // 100)
         elasticapm.set_transaction_result(result, override=False)
+
+    def get_route_name(self, request: Request) -> str:
+        path = None
+        routes = request.scope["app"].routes
+        for route in routes:
+            match, _ = route.matches(request.scope)
+            if match == Match.FULL:
+                path = route.path
+                break
+            elif match == Match.PARTIAL and path is None:
+                path = route.path
+        return path
--- a/elasticapm/contrib/starlette/__init__.py
+++ b/elasticapm/contrib/starlette/__init__.py
@@ -196,13 +196,27 @@ class ElasticAPM(BaseHTTPMiddleware):
         elasticapm.set_transaction_result(result, override=False)
 
     def get_route_name(self, request: Request) -> str:
-        path = None
-        routes = request.scope["app"].routes
+        route_name = None
+        app = request.app
+        scope = request.scope
+        routes = app.routes
+
         for route in routes:
-            match, _ = route.matches(request.scope)
+            match, _ = route.matches(scope)
             if match == Match.FULL:
-                path = route.path
+                route_name = route.path
                 break
-            elif match == Match.PARTIAL and path is None:
-                path = route.path
-        return path
+            elif match == Match.PARTIAL and route_name is None:
+                route_name = route.path
+        # Starlette magically redirects requests if the path matches a route name with a trailing slash
+        # appended or removed. To not spam the transaction names list, we do the same here and put these
+        # redirects all in the same "redirect trailing slashes" transaction name
+        if not route_name and app.router.redirect_slashes and scope["path"] != "/":
+            redirect_scope = dict(scope)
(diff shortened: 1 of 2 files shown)


HenrikOssipoff:
@beniwohli As promised I've taken this for a spin!

It works nicely, and is miles better than the previous implementation! Thank you :)

There's a small thing that I guess is likely working as intended, but from a user perspective isn't the best experience.

I made these two simple routes as a test:

```python
app = Starlette(routes=[
    Route('/users/', list_users),
    Route('/users/{pk:int}/', get_user),
])
```

That is, I forced trailing slashes to my URLs. Attempting to make a request to the URL without the slash causes a redirect in Starlette, but the new code to get the route name couldn't match the route (obviously), causing this mismatch (the ID is listed instead of `{pk:int}`):

Skærmbillede 2020-10-28 kl  19 44 13

This means if you had a big site, your transaction list would still be really cluttered with redirects containing the actual value for the URL path param.

Not sure how easily that's fixable, but it would greatly improve the user experience :)

elasticapm/contrib/starlette/__init__.py
```diff
@@ -192,3 +194,15 @@ async def _request_finished(self, response: Response):
...
+        path = None
+        routes = request.scope["app"].routes
+        for route in routes:
+            match, _ = route.matches(request.scope)
+            if match == Match.FULL:
+                path = route.path
+                break
+            elif match == Match.PARTIAL and path is None:
+                path = route.path
+        return path
```
basepi, line 208: Based on @HenrikOssipoff's comment, I wonder if `path` is `None` when we go to return, should we try appending a `"/"` to `path` to attempt to get a match that way? That *should* solve the matching issues he's seeing, right? Slightly more overhead, but I still think it's worth it to avoid all of those redirect one-off transaction names.

basepi (approved):
This looks like a huge improvement as is, one potential solution to the problems that @HenrikOssipoff is seeing below.

beniwohli:
@HenrikOssipoff @basepi I've implemented a fix for the issue with trailing slashes. In the end I decided to create a separate "bucket" for these transactions and not simply try to find a match with/without trailing slash. The latter would mix "real" requests with redirect requests, and that would give a false impression on the real requests' performance (as redirects will be extremely fast).

WDYT?

elasticapm/contrib/starlette/__init__.py
```diff
@@ -192,3 +194,29 @@ async def _request_finished(self, response: Response):
...
+        # Starlette magically redirects requests if the path matches a route name with a trailing slash
+        # appended or removed. To not spam the transaction names list, we do the same here and put these
+        # redirects all in the same "redirect trailing slashes" transaction name
+        if not route_name and app.router.redirect_slashes and scope["path"] != "/":
+            redirect_scope = dict(scope)
+            redirect_scope["path"] = scope["path"][:-1] if scope["path"].endswith("/") else scope["path"] + "/"
+            for route in routes:
+                match, _ = route.matches(redirect_scope)
+                if match != Match.NONE:
+                    route_name = "redirect trailing slashes"
```
basepi, line 220:
```suggestion
                    route_name = "<internal redirect>"
```

What do you think about this? I'm not sure the "trailing slashes" will mean anything to the average user.

I do like collecting all of these transactions in a separate bucket so that they don't affect the stats for the routes to which they redirect, I'm just not sure of the name.

basepi (approved):

elasticapm/contrib/starlette/__init__.py
beniwohli: I struggled with naming, and you're probably right that my choice isn't super understandable by users not familiar with the internals of Starlette routing (especially because the `Router.redirect_slashes` flag seems to be undocumented and not configurable when instantiating a `Starlette` application). `<internal redirect>` seems to be very broad though, as we won't put any other kinds of redirects into that name bucket...

basepi: Eh, I don't feel strongly. We can keep it as-is.

HenrikOssipoff: Love the changes! Can't wait to see this in master.

As a user, I would probably expect the following behaviour:

```suggestion
                    route_name = route.path[:-1]
```

I _think_ that would be somewhat in line with what we see with other frameworks, that expose the route information on the request objects (I could be wrong though).

beniwohli: @HenrikOssipoff I implemented your suggestion now. I'm a bit concerned that it could spam the transaction list with lots of these transaction names that are only redirects to real transactions, but I guess that's only a problem if the wrong URLs are used consistently.

PR #963, opened 2020-10-30
kafka instrumentation native support for apm-python agent.

### LINKED ISSUE #853: Add instrumentation for Kafka
The apm-agent-java is capable of tracing transactions through Kafka from v7.6, however, Kafka related messages cannot be tracked through the python agent.

apm-agent-python should be able to support Kafka as part of the transaction details captured.

NA

**Additional context**
NA

Rohit-dot-KumarAM:
Add instrumentation for KafkaProducer.send method and KafkaConsumer.next method, while running inside framework.

## How it works:
-> Considering send is being called from a transaction say Django transaction, it will capture a span out of it.
-> Then in elasticapm/instrumentation/packages/kafka.py for send 1 extra header called trace is being added .
-> this "trace" header will contain the tp-string with the span-id.
-> at the receiver's side while instrumenting "next" the "trace" header will be fetched.
-> If there is a running transaction which is not of type "messaging" then will create span for "next".
-> else if there is a transaction with type messaging we will close the transaction considering that the current transaction was started from previous "next" call.
-> else there is no current transaction running then we will create one and make it child of the send-span, making use of "trace" header which contains our tpstring.

## Supported base Technologies:
================
-> Django 1.8.19
-> flask 1.1.2.
-> celery 3.1.26.post2 (considering only kafka-send inside a celery task. Not consume.)

## Related issues
closes #853

### DIFF
--- a/elasticapm/contrib/flask/__init__.py
+++ b/elasticapm/contrib/flask/__init__.py
@@ -49,6 +49,8 @@ from elasticapm.utils.logging import get_logger
 
 logger = get_logger("elasticapm.errors.client")
 
+_client = None
+
 
 def make_client(client_cls, app, **defaults):
     config = app.config.get("ELASTIC_APM", {})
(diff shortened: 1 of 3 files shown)


basepi:
Closing as a duplicate of #964

PR #971, opened 2020-11-16
remove Python 2.7 testing on Windows

beniwohli:
we experience a lot of failed test runs on Windows/Python2.7 due to
a race condition happening in the coverage library.

We will drop support for Python 2.7 soon anyway, so it's not worth
investigating this further.

### DIFF
--- a/.ci/.jenkins_windows.yml
+++ b/.ci/.jenkins_windows.yml
@@ -6,9 +6,6 @@
 #
 
 windows:
-  - VERSION: "2.7"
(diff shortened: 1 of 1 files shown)


beniwohli:
Should have known that adding Python 3.9 in the same PR is a bad idea. Turns out there are issues with building psutils on Windows/Python3.9 right now https://github.com/giampaolo/psutil/issues/1850

basepi (approved):

PR #981, opened 2020-12-02
Avoid the slow pkg_resources package on modern python versions

boxed:
pkg_resources is very slow as can be seen in this simple test:

```
> time python -c "print( __import__('pkg_resources').get_distribution('elastic-apm').version)"
4.2.2
        0.33 real         0.24 user         0.08 sys
        
> time python -c "from importlib.metadata import version; print(version('elastic-apm'))"
4.2.2
        0.08 real         0.05 user         0.02 sys

> time python -c ""                   Wed Dec  2 09:36:26 2020
        0.03 real         0.02 user         0.01 sys
```

### DIFF
--- a/elasticapm/__init__.py
+++ b/elasticapm/__init__.py
@@ -52,7 +52,10 @@ from elasticapm.utils.disttracing import trace_parent_from_headers, trace_parent
 __all__ = ("VERSION", "Client")
 
 try:
-    VERSION = __import__("pkg_resources").get_distribution("elastic-apm").version
+    try:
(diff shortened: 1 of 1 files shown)


boxed:
I have signed the agreement now.

Btw, that signing page doesn't work in Safari. Presumably it does some shady tracking thing that Safari blocks.

beniwohli:
@boxed this is great, thanks!

Regarding the CLA problem with Safari, I'll forward it to the correct team, thanks for the heads up!

PR #982, opened 2020-12-03
[META 319] Align sanitize_field_names with cross-agent spec

### LINKED ISSUE #904: [META 319] Add sanitize_field_names to remote config
Spec: https://github.com/elastic/apm/pull/334
Meta: https://github.com/elastic/apm/issues/319

Default should be:

```
password, passwd, pwd, secret, *key, *token*, *session*, *credit*, *card*, authorization, set-cookie
```

Note that the implementation of this also includes adding the option to [Kibana](https://github.com/elastic/kibana/blob/master/x-pack/plugins/apm/common/agent_configuration/setting_definitions/general_settings.ts).

- [x] Implemented
- [x] Added to Kibana
- [ ] Released (breaking, 6.0)

### LINKED ISSUE #920: Remove query string sanitization in 6.0
To align with other agents, we should remove query string sanitization in the next major version. 

See https://github.com/elastic/apm/pull/334

basepi:

Aligns the python agent's behavior and default for the `sanitize_field_names` config with that defined in the [cross-agent spec](https://github.com/elastic/apm/blob/3fa78e2a1eeea81c73c2e16e96dbf6b2e79f3c64/specs/agents/sanitization.md)

Note that this is a breaking change. Previously, entries in `sanitize_field_names` would be checked with an `if pattern in key` check. This meant that `"secret"` would match (and mask) `"the_secret"`. With the change to wildcard glob matching, this will no longer be the case. Users should add stars to each side of their matches (i.e. `"*secret*"`) in order to retain the previous behavior.

The value sanitization of credit card numbers has also been removed. The credit card validation was of dubious value and removing it will improve performance.

Additionally, the sanitization of request querystrings has been removed.

Finally, the sanitization mask has been changed to `"[REDACTED]"`, to match the recommendation from https://github.com/elastic/apm/pull/378.

This PR should be merged for version 6.0, which will coincide with the 7.11 stack release.

## Related issues

Closes #904 
Closes #920


elasticapm/conf/constants.py
```diff
@@ -60,14 +60,31 @@
 HARDCODED_PROCESSORS = ["elasticapm.processors.add_context_lines_to_frames"]
 
 BASE_SANITIZE_FIELD_NAMES = [
-    "authorization",
```
beniwohli, line 63: couldn't we use `starmatch_to_regex` to convert `BASE_SANITIZE_FIELD_NAMES_UNPROCESSED` into `BASE_SANITIZE_FIELD_NAMES`? That would avoid having to define the list twice, and risking discrepancies with additions/deletions in the future

beniwohli (approved):
just one nitpick, other than that LGTM!

elasticapm/conf/constants.py
basepi: I considered that but it would introduce a circular dependency. I suppose I could copy `starmatch_to_regex` into this file since it doesn't ever change so the duplication probably wouldn't cause a problem.

Issue #983, opened 2020-12-04
docs: Add "added in version x" to the API reference


bmorelli25:
Like other agents, we should document which version each API call was added to the Python agent. This is particularly important because we only publish documentation per major.

I should be able to add this to my 7.11 to-do list.

PR #984, opened 2020-12-09
docs: Add "added in version" to API reference

bmorelli25:

This PR adds `Added in vx.x.x` text to each API method. Here's what it looks like:

<kbd>
Screen Shot 2020-12-09 at 1 50 13 PM
</kbd>

## Related issues

Closes https://github.com/elastic/apm-agent-python/issues/983.


beniwohli (approved):
This is awesome, thank you @bmorelli25!!!

PR #995, opened 2021-01-04
remove Python 3.5 on Windows testing

beniwohli:
psutil doesn't provide binary wheels anymore, and we intend to
remove Python 3.5 support soon anyway.

As an added bonus, let's activate Python 3.9 testing on Windows,
since the latest psutil release has binary wheels for 3.9


basepi (approved):

beniwohli:
Not sure what happened with the commit message of that last commit :/ anyway, it's been squashed...

PR #998, opened 2021-01-05
Fix for json-incompatible elasticsearch bodies

### LINKED ISSUE #997: 5.10.0 "kind of" regression for elasticsearch instrumentation due to logging of whole body
**Describe the bug**:

The elasticsearch library makes it possible to define your own custom serializer as part of the library. However, the Python agent doesn't take this into account, which can lead to wrong request bodies being collected, and it is possible to downright crash the agent (which unfortunately was what happened to us).

The problem seems to have happened with #940.

**To Reproduce - wrong body**

https://gist.github.com/HenrikOssipoff/a231b45ce504691cd71a5de23a771132

This gist shows a simple Starlette application with a custom serializer.

Using the code, the following document is correctly indexed:

```json
{"_index":"test-index","_type":"_doc","_id":"zrmN0nYBBrB60DkP7BVZ","_score":1.0,"_source":{"numbers":2}}
```

However, the wrong body is recorded with the agent:

Skærmbillede 2021-01-05 kl  13 39 05

**To Reproduce - crash**

The following gist shows an application that works perfectly fine with the elasticsearch client, but crashes the apm agent:

https://gist.github.com/HenrikOssipoff/2219d059d3156ab8b11879e1191cefe4

Causes exception `TypeError: keys must be str, int, float, bool or None, not NumberObj`.

I'll admit this is some weird code; this is a simplified version of what we currently use.
The version in our app uses Django's translation module as keys for certain parts of the query, which corresponds to `django.utils.functional.Promise` objects. We're basically looping over these to force-evaluate them, as they're lazy objects.

It works perfectly fine, but crashes the agent.

- OS: [e.g. Linux] Linux/Mac
- Python version: 3.9
- Framework and version [e.g. Django 2.1]: Starlette, Django (any version, probably)
- APM Server version: 7.8.0
- Agent version: >= 5.10.0

**Additional context**

I'm not actually sure how best to solve this. The elasticsearch library already supports a wide array of custom serialization, even without using your own custom one: https://github.com/elastic/elasticsearch-py/blob/master/elasticsearch/serializer.py

The best case (from a user perspective), would be for the agent to somehow find the correct serializer to use, when dealing with the body. I'm not sure if this is even possible.

basepi:

Uses the Elasticsearch serializer for better compatibility

## Related issues
Closes #997

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,22 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 5.x" heading
+[float]
+===== Features
+
+
+[float]
+===== Bug fixes
+
+* Fix for custom serializers in elasticsearch-py {pull}998[#998]
+* Fix large query truncation in psycopg2 {pull}994[#994]
+* Fix memory metrics reporting when `memory.usage_in_bytes` is unavailable {pull}987[#987]
+
+
 [[release-notes-5.x]]
 === Python Agent version 5.x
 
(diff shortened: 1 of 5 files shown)


beniwohli:
@basepi AFAICT, the serialized `body` is an argument to `Connection.perform_request()` already:

https://github.com/elastic/elasticsearch-py/blob/4441da0cbc4ecd02c81b93830b1318647b4018f9/elasticsearch/connection/base.py#L236-L245

So there shouldn't be a need for us to serialize it again and passing it around in the `params` "back channel"

Issue #993, opened 2020-12-29
Unhandled exception with FastAPI and client disabled blocks the response


ca-simone-chiorazzo:
**Describe the bug**: ...
Using FastAPI and the ElasticAPM middleware, we noticed that for an **api view** with APM **disabled** (via config) the HTTP call goes stuck in the method "capture_exception" that never returns.

**To Reproduce**

1. ... Install fast API and APM client
2. ...Install ElasticAPM middleware and configure client with *enabled=false*
3. ...Raise an unhandled exception from an api
4. ...You can see that the HTTP response is never returned to the client and the call remains in a limbo
5. remove elastic middleware
6. Fire again an unhandled exception
7. You can see that the exception is then managed and a 500 is returned to the client

**Environment (please complete the following information)**
- Python version: 3.8.6
- Framework and version: Fastapi 0.63.0
- APM Server version: 
- Agent version: 5.10.0

beniwohli:
Hi @ca-simone-chiorazzo 

I tried reproducing this first in a test (albeit just with Starlette, which is the layer we hook into), and then by adding an `assert False` [here](https://github.com/nsidnev/fastapi-realworld-example-app/blob/master/app/api/routes/tags.py#L14) in [fastapi-realworld-example-app](https://github.com/nsidnev/fastapi-realworld-example-app). I didn't notice anything blocking with `enabled=false`.

If you run your fastapi in the console and hit `Ctrl-C` when it blocks, do you get any traceback that would indicate where the code is blocking exactly?

ca-simone-chiorazzo:
Hi @beniwohli

thanks to having given a look to this issue.

I've tried what you suggest, but once I hit `Ctrl-C` uvicorn tries to wait for pending connections forever..

I report here the log:
```
INFO:     Uvicorn running on http://0.0.0.0:8666 (Press CTRL+C to quit)
INFO:uvicorn.error:Uvicorn running on http://0.0.0.0:8666 (Press CTRL+C to quit)
INFO:     Started reloader process [26394] using statreload
INFO:uvicorn.error:Started reloader process [26394] using statreload
Could not register elasticapm.metrics.sets.cpu.CPUMetricSet metricset: psutil not found. Install it to get system and process metrics
WARNING:elasticapm.metrics:Could not register elasticapm.metrics.sets.cpu.CPUMetricSet metricset: psutil not found. Install it to get system and process metrics
WARNING:rollbar:Rollbar already initialized. Ignoring re-init.
[...]

PR #1000, opened 2021-01-07
Fix starlette to not instrument when enabled=False

basepi:

Noticed that we were gating instrumentation for starlette only on `config.instrument` -- we need to also gate on `config.enabled` to make sure we don't instrument when we shouldn't.

## Related issues
Ref #993

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,15 @@
 repos:
 -   repo: https://github.com/pycqa/isort
-    rev: 5.5.4
+    rev: 5.7.0
     hooks:
     -   id: isort
 -   repo: https://github.com/ambv/black
-    rev: stable
+    rev: 20.8b1
     hooks:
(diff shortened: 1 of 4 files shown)


basepi:
Test failure appears to be a syntax error in pymysql. I'm going to ignore it until this is reviewed/ready to merge, I'm sure they're on it.

beniwohli (approved):
LGTM! Did you check if it would be feasible to write a test for this?

beniwohli:
pymysql 1.0.1 (and 1.0.2) were released in the meantime, with `python_requires >= 3.6", so tests on 2.7 pass again using a pre 1.0 release.

basepi:
@beniwohli I could probably write a test for this though maybe we should make it a broader issue to write tests for all of our frameworks -- as far as I remember seeing, we're not testing `enabled`/`instrument` in other frameworks either.

PR #1001, opened 2021-01-08
Fix python-memcached package name

angelakis:

This PR makes a small change in python-memacached library name.

I think that the package name should be used instead of the imported library name for clarity (or at least referenced alongside).

Until I looked at the code, I thought that python-memcached was not supported based on the documentation, so hopefully this may save a little time for someone.


basepi (approved):

basepi:
Thanks for this!

PR #1002, opened 2021-01-11
remove certifi dependency, and instead rely on urllib3 to find the OS cert store

### LINKED ISSUE #922: Move certifi to an optional dependency
**Describe the bug**: ...

https://github.com/elastic/apm-agent-python lists the license as `BSD-3-Clause`.  However https://pypi.org/project/liccheck/ reports that it pulls in `certifi` which has an MPL license:
```
1 package.
    certifi (2020.6.20): ['Mozilla Public License 2.0 (MPL 2.0)', 'MPL-2.0']
      dependency:
          certifi << elastic-apm
```

My understanding of the MPL is that it is a copy-left license, which means it should have to appear in the larger work's license notification.  If that's the case, that's also problematic for us since the agent has to be built into the source code, and we can't have copy left licenses like GPL/LGPL in there.

**To Reproduce**

1. create a requirements.txt file with nothing but `elastic-apm[flask]` in it
2. create an authorized-licenses.ini file that allows MIT and BSD licenses:
```
[Licenses]
authorized_licenses:
  mit
  bsd

unauthorized_licenses:

[Authorized Packages]

```
3. create a simple Dockerfile:
```
FROM amazonlinux:2018.03

RUN yum update -y && \
    yum install -y python27 python27-devel.x86_64 && \
    yum install -y wget && \
    wget https://bootstrap.pypa.io/get-pip.py && \
    python27 get-pip.py && \
    pip install liccheck && \
    pip install elastic-apm[flask]

COPY ./ ./

RUN liccheck -s authorized-licenses.ini -r requirements.txt -l Paranoid
```
4. Run Docker build:
```
docker build .
```

Result:
```
gathering licenses...
5 packages and dependencies.
check authorized packages...
4 packages.
check unknown packages...
1 package.
    certifi (2020.6.20): ['Mozilla Public License 2.0 (MPL 2.0)', 'MPL-2.0']
      dependency:
          certifi << elastic-apm
The command '/bin/sh -c liccheck -s authorized-licenses.ini -r requirements.txt -l Paranoid' returned a non-zero code: 255
```

- OS: Linux
- Python version: 2.7
- Framework and version [e.g. Django 2.1]: Flask
- APM Server version:
- Agent version:

**Additional context**

- `requirements.txt`:
  <details>
    <summary>Click to expand</summary>

    ```
    elastic-apm[flask]
    ```
  </details>

beniwohli:
This requires at least urllib 1.25.3.

See https://github.com/urllib3/urllib3/blob/master/CHANGES.rst#1253-2019-05-23

## Related issues
fixes #922

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -836,11 +836,12 @@ This disables most of the tracing functionality, but can be useful to debug poss
 | `ELASTIC_APM_VERIFY_SERVER_CERT`  | `VERIFY_SERVER_CERT` | `True`
 |============
 
-By default, the agent verifies the SSL certificate if you use an HTTPS connection to the APM Server.
+By default, the agent verifies the SSL certificate if you use an HTTPS connection to the APM Server,
+using default certificates obtained by calling 
+https://docs.python.org/3/library/ssl.html#ssl.SSLContext.load_default_certs[`SSLContext.load_default_certs()`].
+
 Verification can be disabled by changing this setting to `False`.
 This setting is ignored when <<config-server-cert,`server_cert`>> is set.
-
-NOTE: SSL certificate verification is only available in Python 2.7.9+ and Python 3.4.3+.
 
 [float]
 [[config-server-cert]]
--- a/elasticapm/transport/http.py
+++ b/elasticapm/transport/http.py
@@ -35,7 +35,6 @@ import json
 import re
 import ssl
 
-import certifi
 import urllib3
 from urllib3.exceptions import MaxRetryError, TimeoutError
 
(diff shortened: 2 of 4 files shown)


beniwohli:
@sethmlarson I added you as a reviewer due to your expertise in this area :D 

I'm especially wondering if there is any issues with common docker base images not having bundled SSL cert stores

sethmlarson:
I wouldn't recommend removing certifi as a default, unfortunately the situation for Ubuntu and Debian is that installing Python via `apt` doesn't install `ca-certificates` so [won't be able to verify TLS out of the box](https://gist.github.com/tiran/2dec9e03c6f901814f6d1e8dad09528e#ssl-module-cannot-verify-connections).

beniwohli:
@sethmlarson ugh :( that's what I feared.

I wonder if it is possible to have some way to default on requiring certifi, but allow a way to install the agent without certifi. Similar to`extras_require`, `elastic-apm[without-certifi]`. Unfortunately, `extras_require` only allows to add dependencies, not remove default dependencies.

Naive solution would be to ship an additional package elastic-apm-without-certifi, but I'm not sure that's worth the hassle.

beniwohli:
replaced by #1016

PR #1013, opened 2021-01-20
align elasticsearch body capturing with other agents

### LINKED ISSUE #419: Capture request body for more Elasticsearch APIs
See https://github.com/elastic/apm/issues/55

beniwohli:
See https://github.com/elastic/apm/issues/55

This also allows us to get rid of a hack we used to pass the API method
and an unserialized version of the body to the instrumentation. Neither
is used anymore, so we can skip that part of the instrumentation.

This does also mean that we stop capturing the body for some methods,
specifically `update` and `delete_by_query`.

closes #419


beniwohli:
@basepi @felixbarny I guess strictly speaking this could constitute a backwards incompatible change, so I vote for including this in 6.0, not 5.10.1. What do you think?

felixbarny:
@trentm has been advocating for having the list of endpoints that we capture bodies for be configurable. 

> This does also mean that we stop capturing the body for some methods,
specifically update and delete_by_query.
> I guess strictly speaking this could constitute a backwards incompatible change

Would the backwards compatibility be remained if you just included `*/delete_by_query` and `*/_update/*` to the URL patterns that have body capturing enabled?
This would enable us to merge the new code in now as opposed to the 6.x branch at the expense of being slightly out-of-sync with what the other agents are doing. But in 6.0, we could just remove the patterns again or decide to add them to the default lists of other agents.

For these things, I sometimes struggle to find a clear border of what's considered backwards incompatible. In general, I think we should be able to fine-tune the amount of data we collect without having to major-bump the agent version.
Maybe in this case it's fine to remove the delete_by_query and update APIs from the default list as long as there's a config option for users to re-enable them.

beniwohli:
@felixbarny 6.0 isn't very far away, probably just a couple of weeks.

As for including those two endpoints, at least for `update`, we previously only captured the `script` part of the body (for reasons that I'm sure made sense at the time). If at all possible, I want to avoid adding that back, since we then need to introduce that ugly hack again (or re-parse the already serialized JSON `body`).

felixbarny:
> 6.0 isn't very far away, probably just a couple of weeks.

ah, nvm then

basepi (approved):

elasticapm/instrumentation/packages/elasticsearch.py
```diff
@@ -30,18 +30,15 @@
...
 from elasticapm.instrumentation.packages.base import AbstractInstrumentedModule
-from elasticapm.utils import compat
 from elasticapm.utils.logging import get_logger
 
 logger = get_logger("elasticapm.instrument")
 
-
-API_METHOD_KEY_NAME = "__elastic_apm_api_method_name"
-BODY_REF_NAME = "__elastic_apm_body_ref"
+should_capture_body_re = re.compile("_((search|msearch)(/template)?|count)$")
```
trentm, line 41: FWIW, this is what the node.js agent is currently using:

```
// URL paths matching the following pattern will have their query params and
// request body captured in the span (as `context.db.statement`). We match
// a complete URL path component to attempt to avoid accidental matches of
// user data, like `GET /my_index_search/...`.
const pathIsAQuery = /\/(_search|_msearch|_count|_async_search|_sql|_eql)(\/|$)/
```

That adds `_async_search` (requested during some Elasticsearch perf work by @ gingerwizard), and a couple others. It also matches full URL segments to avoid accidental matches.

trentm: I realize I haven't yet opened anything on apm.git to spec this out.

beniwohli: @trentm awesome, thanks! I actually copied the regex from the Node.js agent, but from the original PR that introduced the regex, not from current master :facepalm:

felixbarny: A shared spec would be great for this!

PR #1015, opened 2021-01-21
Fix GraphQL instrumentation: span spamming

seedofjoy:

Do not create spans for scalar fields with a required flag, because it causing span spamming.
Before this commit it's worked only for optional fields.

### DIFF
--- a/elasticapm/instrumentation/packages/graphql.py
+++ b/elasticapm/instrumentation/packages/graphql.py
@@ -52,7 +52,7 @@ class GraphQLExecutorInstrumentation(AbstractInstrumentedModule):
         query = args[2]
 
         if "ResolveInfo" == type(query).__name__:
-            if str(query.return_type) in [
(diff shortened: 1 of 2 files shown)


basepi (approved):
This looks great! Can you please sign [the CLA](https://www.elastic.co/contributor-agreement)?

elasticapm/instrumentation/packages/graphql.py
```diff
@@ -52,7 +52,7 @@ def call(self, module, method, wrapped, instance, args, kwargs):
         query = args[2]
 
         if "ResolveInfo" == type(query).__name__:
-            if str(query.return_type) in [
+            if str(query.return_type).rstrip('!') in [
```
basepi, line 55:
```suggestion
            if str(query.return_type).rstrip("!") in [
```

This is why lint is failing

basepi:
Looks like there are some tests you'll need to fix. Keep me posted if you need any help.

PR #1016, opened 2021-01-25
make certifi import optional

beniwohli:
this allows users to uninstall certifi if they have objections
to its MPL license.

replaces #1002


gtback (approved):
This makes sense to me, and is how I would get around making a `certifi` a "default but removable" dependency. 👍🏻

basepi (approved):

PR #1031, opened 2021-01-29
Add support for httplib2 (#1022)

### LINKED ISSUE #1022: Support for google API client (via httplib2)
**Describe the solution you'd like**: I would like http requests done by `httplib2` captured the same as from the other http libraries.

**Additional context**: `httplib2` library is used in the [google API client](https://github.com/googleapis/google-api-python-client) which would be extremely useful to see as spans.

I would be happy to work on a PR if it would be accepted.

Thank you!

HonzaKral:

Implement instrumentation for `httplib2`

## Related issues
closes #1022


elasticapm/instrumentation/packages/httplib2.py
```diff
@@ -0,0 +1,100 @@
...
+    name = "httplib2"
+
+    instrument_list = [("httplib2", "Http.request")]
+
+    # https://httplib2.readthedocs.io/en/latest/libhttplib2.html#httplib2.Http.request
+    args_list = ("url", "method", "body", "headers")
+
+    def call(self, module, method, wrapped, instance, args, kwargs):
+        # grab agrs passed in as positional
+        params = dict(zip(self.args_list, args))
```
beniwohli, line 48: Nifty!

elasticapm/instrumentation/packages/httplib2.py
```diff
@@ -0,0 +1,100 @@
...
+                span_id=parent_id, trace_options=TracingOptions(recorded=True)
+            )
+            self._set_disttracing_headers(headers, trace_parent, transaction)
+
+            response, content = wrapped(*args, **kwargs)
+            if span.context:
+                span.context["http"]["status_code"] = response.status
+            span.set_success() if response.status < 400 else span.set_failure()
+            return response, content
+
```
beniwohli, line 93: We also need to re-implement `mutate_unsampled_call_args`, something like https://github.com/elastic/apm-agent-python/blob/15278433c97d9c81c5ad1dcdf57d27da847be912/elasticapm/instrumentation/packages/urllib3.py#L105-L116

This sets the distributed tracing headers even if this transaction isn't sampled. That way, we can avoid breaking distributed tracing chains if we, for some reason, don't sample the transaction.

beniwohli (commented):
Looks great so far! Let me know if you want to tackle `mutate_unsampled_call_args` yourself or if we should take it from here!

(I already took the liberty to add a couple boilerplate files, to get Jenkins to actually run your tests :D )

elasticapm/instrumentation/packages/httplib2.py
HonzaKral: I saw this code and I was wondering - wouldn't this only set the header if a user is passing in the `headers` kwarg (line 110)? This might work for some libraries where we instrument an "internal" method, but for `httplib2` the request is the public API. Is that correct thought process or am I missing something?

beniwohli: @HonzaKral yes, looks like you're right, if the user creates a `HTTP(S)ConnectionPool` directly, they could call `urlopen` with `headers` as a positional argument.

I think we're saved here by the fact that most people probably use a `PoolManager` as shown in the [user guide](https://urllib3.readthedocs.io/en/latest/user-guide.html).

I'll open a PR with a fix for the urllib3 instrumentation, thanks for the heads up!

HonzaKral: >  they could call urlopen with headers as a positional argument

Yes, they could also not specify headers at all actually, or specify `None` as a positional argument for headers which will be a bit trickier to replace. This is a corner case even in the `httplib2` instrumentation I submitted here...

beniwohli: @HonzaKral I tried to capture all cases in #1033 (also noticed that headers can also be set on the `HTTPConnectionPool` object, which will be used as a fallback if no headers are set in the `urlopen` call)

HonzaKral: added in 62188486

PR #1035, opened 2021-02-02
with Python 2 support gone, we can update a bunch of requirements

beniwohli:
pytest-bdd isn't updated all the way, as it introduced a few
backwards incompatibilities that need to be fixed first

A test of urllib3 had to be modified slightly, as some
requirement update seems to truncate the error message we
are looking for. It's not quite clear where this happens.


basepi (approved):
👏

PR #1037, opened 2021-02-03
Fix transaction names for starlette Mount routes

### LINKED ISSUE #1036: Inaccurate transaction names for FastAPI sub-applications
**Description**
For requests to endpoints defined in FastAPI sub-applications, the mount path is chosen as the transaction name. I expected the full route of the endpoint.

**To Reproduce**

1. Run the following simple FastAPI app:

```python
import uvicorn
from elasticapm.contrib.starlette import ElasticAPM, make_apm_client
from fastapi import FastAPI

app = FastAPI()
sub = FastAPI()
app.mount("/sub", sub)

apm = make_apm_client(
    {
        "SERVICE_NAME": "sub-app-test",
    }
)

@sub.get("/hi")
async def hi():
    return "hi"


@sub.get("/bye")
async def bye():
    return "bye"


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=8888)
```

2. 
- **Observed behavior**
The transactions of `/sub/hi` and `/sub/bye` are both named `/sub` and grouped.
transactions

- **Expected behavior**
The transactions of `/sub/hi` and `/sub/bye` are named according to the full route.

**Environment**
- OS: 
  - Client: Windows
  - Server: Ubuntu
- Python version: 3.7.3
- Framework and version: `fastapi==0.61.2`
- APM Server version: docker image `elasticsearch/elasticsearch:7.10.2` 
- Agent version: `elastic-apm==6.0.0`

pypae:
Fix transaction names for starlette [Mount](https://www.starlette.io/routing/#submounting-routes) routes. 
Recursively build route names for mounted sub-applications.

## Related issues
closes #1036 

*Note: Opened it as a draft pull request because I couldn't write tests yet, but I still wanted to give you an idea of a potential fix.*

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -44,6 +44,7 @@ endif::[]
 * Fix for potential `capture_body: error` hang in Starlette/FastAPI {pull}1038[#1038]
 * Fix a rare error around processing stack frames {pull}1012[#1012]
 * Fix for Starlette/FastAPI to correctly capture request bodies as strings {pull}1042[#1041]
+* Fix transaction names for Starlette Mount routes {pull}1037[#1037]
 
 
 [[release-notes-6.x]]
--- a/elasticapm/contrib/starlette/__init__.py
+++ b/elasticapm/contrib/starlette/__init__.py
@@ -210,24 +210,7 @@ class ElasticAPM(BaseHTTPMiddleware):
         app = request.app
         scope = request.scope
         routes = app.routes
-
-        def _get_route_name(scope, routes, route_name=None):
-            for route in routes:
-                match, child_scope = route.matches(scope)
-                if match == Match.FULL:
-                    route_name = route.path
-                    child_scope = {**scope, **child_scope}
-                    if isinstance(route, Mount):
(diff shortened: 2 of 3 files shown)


beniwohli:
@pypae great work on the fix! I had a look at the issue that you opened yesterday and came to the same conclusion that a recursive approach is probably the cleanest. I don't expect these hierarchies of mounts to become very deep, so the overhead of recursion over an iterative approach should be fairly small.

I took the liberty to push a test I wrote yesterday to confirm the issue to your branch.

beniwohli (approved):
Great work, thanks!

pypae:
I just realized that trailing slash redirects do not work correctly with sub-applications.
On another note: Are you okay with the nested and untyped function `_get_route_name`, or should I move it?

Thank you for this awesome project by the way.

basepi:
> Are you okay with the nested and untyped function _get_route_name, or should I move it?

I don't necessarily have a problem with it, but since it's not a closure it would probably be cleaner to make it a separate method.

basepi (approved):

PR #1039, opened 2021-02-05
updated test matrix and supported technologies docs

beniwohli:


basepi (approved):
Thanks for grabbing these. Can you also add py3.9 to setup.cfg?

Issue #1019, opened 2021-01-27
Wrapper script for no-code-changes onboarding


basepi:
Making code changes optional for agent onboarding would be a win, allowing more flexibility with an ops-led onboarding story.

Should be doable with a pythonpath insertion and a `sitecustomize.py` in the target path.

[Meta](https://github.com/elastic/apm-agent-python/issues/1600)

PR #1042, opened 2021-02-10
Add defaults for all required config values

### LINKED ISSUE #1008: [META 382] Zero-configuration support
Implement zero-configuration support for the agent as described in: https://github.com/elastic/apm/issues/382

SPEC issue: https://github.com/elastic/apm/issues/401

basepi:

Adds a default for all required configuration values. This allows the agent to be used with zero configuration alongside something like the Fleet Agent. We don't do any fancy detection around the `SERVICE_NAME`, it just defaults to `python_service` (and logs a warning if it's not overridden by the user).

"Zero-configuration" is actually pretty worthless by itself, but the hope is that along with #1019 we should be able to allow the user to onboard with no code changes and just an invocation change.

## Related issues
Fixes #1008

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -77,8 +77,8 @@ apm = ElasticAPM(app)
 ----
 
 [float]
-[[required-options]]
-=== Required options
+[[core-options]]
+=== Core options
 
 [float]
 [[config-service-name]]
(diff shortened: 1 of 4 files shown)


beniwohli (approved):
LGTM!

PR #964, opened 2020-10-30
kafka instrumentation native support for apm-python agent.

Rohit-dot-KumarAM:
Add instrumentation for KafkaProducer.send method and KafkaConsumer.next method, while running inside framework.

## How it works:
-> Considering send is being called from a transaction say Django transaction, it will capture a span out of it.
-> Then in elasticapm/instrumentation/packages/kafka.py for send 1 extra header called trace is being added .
-> this "trace" header will contain the tp-string with the span-id.
-> at the receiver's side while instrumenting "next" the "trace" header will be fetched.
-> If there is a running transaction which is not of type "messaging" then will create span for "next".
-> else if there is a transaction with type messaging we will close the transaction considering that the current transaction was started from previous "next" call.
-> else there is no current transaction running then we will create one and make it child of the send-span, making use of "trace" header which contains our tpstring.

## Supported base Technologies:
================
-> Django 1.8.19
-> flask 1.1.2
-> celery 3.1.26.post2 (considering only kafka-send inside a celery task. Not consume.)

## Related issues
closes #853

PR #1043, opened 2021-02-10
[WIP] Make Client object globally available

basepi:

Managing the Client object has always been a bit of a pain. Most of the time passing around the Client via the framework's application objects works ok. But as we move into supporting more instrumentations that require creating transactions (like [Kafka](https://github.com/elastic/apm-agent-python/pull/964)), getting the client object becomes more of a pain.

The idea behind this (unfinished) PR is to painlessly save the Client object when it is created, and provide `elasticapm.get_client()` to make it easy to get that Client object from anywhere.

If we decide this is a good idea, the next step is to remove all of the framework-specific scaffolding around the Client object and use this new `elasticapm.get_client()` everywhere.

@beniwohli thoughts?


beniwohli:
Interesting idea! So far, there is nothing in the agent that would make it impossible to have multiple agent instances running, so we'll have to make sure we don't break any relevant use cases with this.

I think this change would also allow us to tie instrumentation to the agent instance, making it possible to make the list of instrumentations configurable, which would be nice. Not sure how that would interact with the wrapper script that we plan to introduce, though.

Semi-related, looks like we forgot another 6.0 breaking change :( : https://github.com/elastic/apm-agent-python/blob/c1a912642676d3609ef4071b3380bcd1587454ce/elasticapm/handlers/logging.py#L60-L68

basepi:
> So far, there is nothing in the agent that would make it impossible to have multiple agent instances running, so we'll have to make sure we don't break any relevant use cases with this.

Do you have any use cases in mind? In the "multiple agent instances" case we end up with separate client instances anyway before this PR, right?

> Not sure how that would interact with the wrapper script that we plan to introduce, though.

Luckily we get to write the wrapper script so I don't expect it to be a problem. :)

basepi:
I realized that I don't really need to change every instance of `self.client` and `app.client` to `get_client()` -- but I did do a little cleanup in some places to use the new system (especially in Django land where we were already using globals to store the Client object).

I can't think of any other places that will be hit adversely by this change, so I think this is ready for review.

tests/contrib/django/django_tests.py
```diff
@@ -511,17 +511,6 @@ def test_response_error_id_middleware(django_elasticapm_client, client):
...
-    with mock.patch.dict("os.environ", {"ELASTIC_APM_METRICS_INTERVAL": "0ms"}):
-        client2 = get_client("elasticapm.base.Client")
-        try:
-            assert get_client() is get_client()
-            assert client2.__class__ == Client
-        finally:
-            get_client().close()
-            client2.close()
-
-
```
basepi, line 524: Django's `get_client()` is only ever called with an argument here, in this test. Thus, I don't think we need the type-checking functionality that was in the old implementation. I just removed the test.

beniwohli (approved):
Lots of simplifications,I like!

DougDK:
> Interesting idea! So far, there is nothing in the agent that would make it impossible to have multiple agent instances running, so we'll have to make sure we don't break any relevant use cases with this.
> 
> I think this change would also allow us to tie instrumentation to the agent instance, making it possible to make the list of instrumentations configurable, which would be nice. Not sure how that would interact with the wrapper script that we plan to introduce, though.
> 
> Semi-related, looks like we forgot another 6.0 breaking change :( :
> 
> https://github.com/elastic/apm-agent-python/blob/c1a912642676d3609ef4071b3380bcd1587454ce/elasticapm/handlers/logging.py#L60-L68

It looks like there's a use case that this would break. I create an issue for it to investigate a possible bug.
[Issue 1098](https://github.com/elastic/apm-agent-python/issues/1098)

PR #1050, opened 2021-02-22
Fix excepthook to take correct arguments

### LINKED ISSUE #1049: Using use_elastic_excepthook throws TypeError
**Describe the bug**:
When using the option [use_elastic_excepthook](https://www.elastic.co/guide/en/apm/agent/python/master/configuration.html#config-use-elastic-excepthook) every uncaught exception also produces the following `TypeError`:

```bash
Error in sys.excepthook:
TypeError: _excepthook() takes 1 positional argument but 4 were given
```

**To Reproduce**

Run the following Python snippet
```python
import elasticapm


def run():
    raise Exception("hello")

client = elasticapm.Client(
    {"SERVICE_NAME": "test"},
    use_elastic_excepthook=True,
)
run()
```
It will produce this output:

```bash
Error in sys.excepthook:
TypeError: _excepthook() takes 1 positional argument but 4 were given

Original exception was:
  File "t.py", line 13, in <module>
    run()
  File "t.py", line 5, in run
    raise Exception("hello")
Exception: hello
```

- OS: MacOS 10.15.3
- Python version: 3.8.5
- Agent version: 6.0.0

**Additional context**

The problem seems to be with the function signature of the [Client._excepthook](https://github.com/elastic/apm-agent-python/blob/b7e408a7b8c9554c0ee74a5b7e7c6347b4b814f2/elasticapm/base.py#L567) function that is used to override the `sys.excepthook`.

The [sys.excepthook](https://docs.python.org/3/library/sys.html#sys.excepthook) expects three arguments (type, value, traceback), while the `Client._excepthook` only expects the `self` argument.

basepi:

I don't know how we got the original excepthook merged without catching this. Did we not test? 🤦‍♂️ 

In any case, this fixes our excepthook to take the correct arguments.

However, I have no idea how to test this because I can't get an exception from within a pytest test to trigger `sys.excepthook`. Any ideas, @beniwohli?

## Related issues
Fixes #1049


beniwohli:
@basepi `sys.excepthook` is only called at the very end of life of the python process, so I don't see how we could fully test it in-process. We could trigger it in a subprocess, but that has other fun challenges.

I think calling `sys.excepthook` ourselves with the right arguments should exercise our own code enough.

basepi:
@beniwohli test added, feel free to review at your leisure.

beniwohli (approved):
LGTM!

PR #1051, opened 2021-02-24
Fix starlette/fastapi hang on capture_body 

adamwen829:
#1038 hack the request to make everyone access, but there is still a chance of hanging. 

apm-agent-python use the [following code](https://github.com/elastic/apm-agent-python/commit/6cffb0ecbf93f442b4f2bcbf3fb1a95dc6049451#diff-f9e770f7326cda6eb54ca0dd8842e87343851916da33047083fde4810ee438b0R110-R111) to act a function to return starlette Message
```python
async def receive() -> Message:
    return {"type": "http.request", "body": body}
```

And it will be calling and hanging by [starlette Response process](https://github.com/encode/starlette/blob/master/starlette/responses.py#L223-L230)

`receive` what we declared will be passed as parameter `receive` to `__call__` and will hang in `run_until_first_complete`

Cause according [PEP492](https://www.python.org/dev/peps/pep-0492/#await-expression)
> every await is suspended by a yield somewhere down the chain of await calls

`receive` we declared don't have `yield` in the chain,it will not suspends and pass control to the event loop.

The following code is a minimal example to reproduce it
```Python
In [1]: import asyncio
In [2]: loop = asyncio.get_event_loop()
In [3]: async def foo():
   ...:     async def receive():
   ...:         return "haha"
   ...:
   ...:     while True:
   ...:         data = await receive()

In [4]: async def bar():
   ...:     while True:
   ...:         await asyncio.sleep(1)
   ...:         print("Hello")

In [5]: loop.create_task(foo())
In [6]: loop.create_task(bar())
In [7]: loop.run_forever()
```
This code will hang(trap in `while loop` in `foo`0  and never print `Hello`

### DIFF
--- a/elasticapm/contrib/starlette/utils.py
+++ b/elasticapm/contrib/starlette/utils.py
@@ -27,6 +27,7 @@
 #  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
 #  OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+import asyncio
 
 from starlette.requests import Request
 from starlette.responses import Response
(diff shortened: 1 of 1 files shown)


basepi (approved):
This is a *great* catch. Thank you! I guess I need to buff up my asyncio knowledge a little more. :)

PR #1055, opened 2021-02-25
Use the new auto-synced json schema files

basepi:

Fixes the test suite to use the new auto-synced json schema files. Also cleans up a lot of scaffolding around `make update-json-schema`

## Related issues
Ref #1053


beniwohli (approved):
LGTM

My only concern is that we lose the ability to easily test against other versions of the APM Server spec. I'm not sure if I ever needed that, so I guess it's not a huge issue.

PR #1058, opened 2021-03-03
Add `auto_ecs_logging` config

### LINKED ISSUE #1006: [META 373] Support replacement of plain-text formatters with ECS formatters
Evaluate and if applicable implement https://github.com/elastic/apm/issues/373

basepi:

Adds a new config value, called `auto_ecs_logging`, which defaults to False. If the user sets this value to True (and has `ecs_logging` installed), we will do two things:

1. Call `setFormatter(ecs_logging.StdlibFormatter())` on each of the handlers on the root logger
2. Set the default processor for `structlog` (if installed) to `ecs_logging.StructlogFormatter()`.

I also documented the new setting and added a test.

## Related issues
Closes #1006

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,6 +38,7 @@ endif::[]
 ===== Features
 
 * Add global access to Client singleton object at `elasticapm.get_client()` {pull}1043[#1043]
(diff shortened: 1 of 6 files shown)


beniwohli (approved):
LGTM!

PR #1063, opened 2021-03-09
Refactor auto_ecs_logging to log_ecs_formatting

basepi:
This matches the Java implementation which is more flexible (due
to using an enum instead of a bool) and will match the upcoming
ecs log formatting [spec](https://github.com/elastic/apm/pull/423)

Ref #1058 and #1006


beniwohli (approved):
LGTM!

PR #1068, opened 2021-03-11
Fix issue when resetting config after new value has been added

### LINKED ISSUE #1067: Traceback from agent on max spans
**Describe the bug**: ...

We are seeing the following traceback coming out of the ElasticAPM agent via Sentry frequently on many of our environments.

```
KeyError: 'TRANSACTION_MAX_SPANS'
  File "elasticapm/utils/threading.py", line 80, in run
    rval = self._function(*self._args, **self._kwargs)
  File "elasticapm/conf/__init__.py", line 701, in update_config
    self.reset()
  File "elasticapm/conf/__init__.py", line 650, in reset
    if self._config.values[key] != self._first_config.values[key]:
```

**To Reproduce**

Unknown. However, Sentry shows a lot of POST requests going to the APM server before this happens. Our assumption is that we are hitting TRANSACTION_MAX_SPANS, but this is not being handled gracefully, and causing this exception instead.

- OS: Linux
- Python version: 3.8
- Framework and version: Django 3.1.7
- APM Server version: 7.6
- Agent version: 6.0.0

beniwohli:
When a new value has been added to the config, resetting to the original config can raise a `KeyError`

## Related issues
closes #1067

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -38,7 +38,7 @@ endif::[]
 ===== Features
 
 * Add global access to Client singleton object at `elasticapm.get_client()` {pull}1043[#1043]
-* Add `auto_ecs_logging` config option {pull}1058[#1058]
+* Add `log_ecs_formatting` config option {pull}1058[#1058] {pull}1063[#1063]
 
 
 [float]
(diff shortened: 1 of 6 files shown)


basepi (approved):
Good catch!

PR #1078, opened 2021-03-23
more eclusions due to collections.abc moves

beniwohli:


beniwohli:
jenkins run the full tests please

PR #1108, opened 2021-04-21
Add some data about response data at ES instrumentation

### LINKED ISSUE #354: Tracking enhacements for `Elasticsearch.search`
Currently in https://github.com/elastic/apm-agent-python/blob/master/elasticapm/instrumentation/packages/elasticsearch.py#L47 we are discarding everything but the `query` in the body which can cover some very important (from a performance point of view) aspects of the operation for example pagination - large `from`/`size` can have severe impact - and aggregations - if present can severely impact everything.

It would also be great to track the statistics on the response:

`took`, `hits.total`, `_shards.total/failed/successful` information

which can expose badly structured code/queries or performance issues in elasticsearch

Thank you!

Sparkycz:

Adds `http.status_code` and `db.row_affected` from ElasticSearch response in ES instrumentation

## Related issues
closes #354


basepi (approved):
Looks awesome, thanks for grabbing this!

basepi:
I'm a bit mystified as to why the tests are failing. I'll keep digging but the `result_data["hits"]["total"]["value"]` should definitely be working based on my local testing.

basepi:
I spoofed the tests to point to my local elasticsearch instance and they passed. I'm going to investigate the elasticsearch that's used in the integration tests (https://github.com/elastic/apm-integration-testing) -- I'm guessing the issue is there.

basepi:
Oh, I figured it out. It's failing against Elasticsearch 2.x (which is not surprising). Just need to add some conditionals. Stay tuned.

Sparkycz:
Thanks a lot for finishing the change. I’ve been out of Github for a while..

PR #1109, opened 2021-04-22
added item to release checklist for updating docs repo

beniwohli:

We forgot in the past to add the new major version to the docs repo, hopefully this should avoid that from happening in the future.


bmorelli25 (approved):
Good idea–thanks

PR #1113, opened 2021-04-26
exclude pypy-3/psycopg-2.7

beniwohli:
pypy-3 gained support for the `async` keyword, making code fail that uses
`async` as an identifier. Such as old versions of pyscopg2cffi.


beniwohli:
Test failures are unrelated, merging.

Issue #1111, opened 2021-04-24
Python agent should show database as a dependency, shows up in span


ajay1mg:
**Is your feature request related to a problem? Please describe.**
I can see Postgres under **Time spent by span type** graph but it is not shown as a dependency. Rails agent is able to show this as Postgresql

<img width="397" alt="Screenshot 2021-04-24 at 12 03 10 PM" src="https://user-images.githubusercontent.com/36618391/115949791-14edbe80-a4f5-11eb-999a-9b2880460abc.png">

I am using asyncpg==0.21.0

**Describe the solution you'd like**
The database(Postgres) should show up as a dependency just like it shows up for ruby agent. Screenshot attached below.

<img width="979" alt="Screenshot 2021-04-24 at 11 40 09 AM" src="https://user-images.githubusercontent.com/36618391/115949754-d526d700-a4f4-11eb-9d43-bad114aad4eb.png">

PR #1114, opened 2021-04-26
add database as a dependency in asyncpg instrumentation

ajaygupta2790:
Adds database(postgres) as a dependency in python elastic apm for asyncpg instrumentation

I have taken the code from psycopg2 instrumentation which is able to show database as a dependency. Ran these changes locally and I can now see postgres under dependencies.

Screenshot 2021-04-26 at 2 28 07 PM

## Related issues
closes  https://github.com/elastic/apm-agent-python/issues/1111


beniwohli:
@ajaygupta2790 great catch, the fix looks good!

If you don't mind, could you [sign the CLA](https://www.elastic.co/contributor-agreement)? Then we can get this merged ASAP :)

ajaygupta2790:
I have already signed this, twice :P

beniwohli:
@ajaygupta2790 ah, I see, the problem is that your commits are linked to the @ajay1mg account, for which we don't have a CLA on file yet. Would you mind signing the CLA again, using the `ajay1mg` account? That would be the quickest way to resolve the issue.

PR #1118, opened 2021-04-27
Fix httpx 0.18.0

beniwohli:
The newly released httpx 0.18.0 and its dependency httpcore 0.13 changed some internals that affect us.


beniwohli:
test failure is unrelated

PR #1119, opened 2021-04-28
exclude old version of httpx on Python 3.10 as it fails

beniwohli:


beniwohli:
test failures are unrelated

PR #1120, opened 2021-04-28
fix warnings raised for pyodbc

beniwohli:


beniwohli:
unrelated test failure

PR #1137, opened 2021-05-26
Fix Starlette middleware - providing only static files

### LINKED ISSUE #1131: Crash on starlette middleware in mount endpoints (StaticFiles)
**Describe the bug**: Elastic middleware crash on access to static files

- OS: [e.g. Linux]
- Python version: 3.9.4
- Framework and version [e.g. Django 2.1]: FastAPI
- APM Server version: 7.12.1
- Agent version: 7.12.1

```python
from __future__ import annotations

import uvicorn
from elasticapm.contrib.starlette import ElasticAPM, make_apm_client
from fastapi import FastAPI
from fastapi.staticfiles import StaticFiles

app = FastAPI(routes=[])
apm_config = {
    "SERVICE_NAME": "testing",
    "ENABLED": True,
    "SERVER_URL": "http://localhost:8200",
    "CAPTURE_HEADERS": True,
    "CAPTURE_BODY": "all",
}
apm = make_apm_client(apm_config)
app.mount("/static",StaticFiles(directory="/"),name="static")


if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", debug=False)
```

The issue is happens when elastic running the function _get_route_name.
the middleware will try to loop over the routes where starlette will return None in some cases:
```python
  File "/home/syniex/.local/lib/python3.9/site-packages/elasticapm/contrib/starlette/__init__.py", line 233, in _get_route_name
    for route in routes:
                 └ None

TypeError: 'NoneType' object is not iterable

```
Starlette code that makes the issue:
```python
@property
    def routes(self) -> typing.List[BaseRoute]:
        return getattr(self.app, "routes", None)
```

Sparkycz:

Starlette can provide also static files but when a developer defined routes only for static files the middleware crashed.

## Related issues
closes #1131


basepi (approved):
This looks good to me, let me just double check with @beniwohli and we'll get this merged!

PR #1139, opened 2021-05-26
Ensure that metrics are flushed before shutting down

beniwohli:
To ensure that the flushed metrics are actually sent, we need
to make sure that the transport thread is shut down last.

See also https://github.com/elastic/apm/pull/444

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -33,6 +33,7 @@
 import gzip
 import os
 import random
+import sys
 import threading
 import time
 import timeit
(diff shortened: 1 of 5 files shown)


briancurtin (approved):
Looks good to me!

/cc @stj — this will help our use of custom metrics

elasticapm/transport/base.py
```diff
@@ -86,6 +87,7 @@ def __init__(
         self._closed = False
         self._processors = processors if processors is not None else []
         super(Transport, self).__init__()
+        self.priority = sys.maxsize  # ensure that the transport thread is always started/stopped last
```
felixbarny, line 90: Does that correspond to the native thread's priority? If so, could that impact CPU scheduling in a way that takes away precedence from application threads? If not, maybe choose a different name for the property to avoid these associations. Maybe something like `order`.

PR #1171, opened 2021-06-21
Make units optional for duration/size formats in config

basepi:

I've often thought it was a bit odd that we required the `ms` whenever setting a configuration value that was a duration. While writing the spec for the upcoming `span_stack_trace_min_duration` I discovered that some agents (Java, .NET, PHP) don't require the unit (and have configurable default units). This PR is designed to bring the Python agent in line with this behavior in those other agents. A different default unit can be chosen for each configuration option.

Assuming @beniwohli agrees, I still need to:

- [ ] Update/write tests

## Related issues
Ref https://github.com/elastic/apm/pull/454

### DIFF
--- a/docs/configuration.asciidoc
+++ b/docs/configuration.asciidoc
@@ -216,6 +216,9 @@ The size of the log file if <<config-log_file>> is set.
 The agent always keeps one backup file when rotating, so the maximum space that
 the log files will consume is twice the value of this setting.
 
+It has to be provided in *<<config-format-size, size format>>*, with a default
+unit of `mb`.
+
 [float]
 [[config-log_ecs_formatting]]
 ==== `log_ecs_formatting`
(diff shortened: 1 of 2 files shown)


beniwohli:
At least back when we implemented these units, it was agreed that we should be strict in requiring a unit (see discussion in https://github.com/elastic/apm-dev/issues/282, https://github.com/elastic/apm-dev/issues/310). I'm don't know the reasoning behind why some agents are less strict / provide a default unit per setting (maybe just because there isn't a spec for it outside of those issues), but I kinda like the explicitness of requiring a unit. With defaults, there's always the potential for confusion.

basepi:
Closing this. The current desired path is moving toward making all units required. I was confused because the brand new PHP agent introduced default units, so I assumed that was the desired path, but apparently that was just an oversight.

Issue #1172, opened 2021-06-24
Fix for Starlette 0.15.0


basepi:
Tests are failing.

From the Changelog:
> This release includes major changes to the low-level asynchronous parts of Starlette.

PR #1174, opened 2021-06-24
Update test framework for starlette 0.15.0

basepi:
Fixes https://github.com/elastic/apm-agent-python/issues/1172


basepi:
The new version of the `BaseHTTPMiddleware` from Starlette that uses `anyio` now raises a `RuntimeError` if there's no response to stream. This `raise` obfuscates the actual exception that we're trying to handle in our middleware. (Raising an error definitely means there's no response to stream.)

Seems like the best solution is to move to an ASGI middleware directly, instead of using Starlette's helper class `BaseHTTPMiddleware`.

elasticapm/contrib/starlette/__init__.py
```diff
@@ -108,34 +104,69 @@ class ElasticAPM(BaseHTTPMiddleware):
...
+            if not message:
+                break
+            if message["type"] == "http.request":
+                b = message.get("body", b"")
+                if b:
+                    body += b
+                if not message.get("more_body", False):
+                    break
+            if message["type"] == "http.disconnect":
+                break
```
beniwohli, line 159: Won't this assemble the whole response in-memory before returning it? That would negate any benefit the user wanted to have by using streaming responses in the first place (e.g. streaming large responses that are built on-the-fly).

Or I misunderstood something, which is quite possible.

basepi: This is request-side, not response side. And yes, it will construct the whole request body in memory, and then create a wrapped callable to allow the user to also "stream" the request body. This is the only way we can get the request body in middleware -- see https://github.com/encode/starlette/issues/495#issuecomment-513138055

(We were doing this before this PR, it's just uglier now)

beniwohli (approved):
LGTM!

Issue #1184, opened 2021-07-07
Cannot startup fastapi with apm 


alexcc4:
With start apm in docker, I just set up agent in fastapi. Then I run app with uvicorn, it hangs.
**To Reproduce**

1. Start Service with docker compose [Reference](https://www.elastic.co/guide/en/apm/get-started/current/quick-start-overview.html)

![image](https://user-images.githubusercontent.com/5210569/124731645-fff5c880-df44-11eb-8b4f-876ec7b0ea3e.png)

- Test apm status
![image](https://user-images.githubusercontent.com/5210569/124731884-3af7fc00-df45-11eb-9f6b-856fc5ef8ae6.png)

2. Run the following simple FastAPI app:
```
from fastapi import FastAPI
from elasticapm.contrib.starlette import make_apm_client, ElasticAPM
import uvicorn


apm = make_apm_client({'SERVICE_NAME': 'test', 'SERVER_URL': 'http://0.0.0.0:8200'})
app = FastAPI()
app.add_middleware(ElasticAPM, client=apm)


@app.get("/")
def read_root():
    return {"Hello": "World"}


if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=8000)
```

- Result
![image](https://user-images.githubusercontent.com/5210569/124732134-83171e80-df45-11eb-92ff-bbcc0c3598f4.png)

**Environment**
- OS: Mac
- Python version: `3.7.10`
- Framework and version: `fastapi==0.66.0`
- APM Server version: `image: docker.elastic.co/apm/apm-server:7.13.2`
- Agent version: `elastic-apm==6.3.0`

beniwohli:
Hi @alexcc4! Thanks for opening the issue and providing a test case! I opened a PR with a fix in #1185. We'll make a release with this fix ASAP!

beniwohli:
@alexcc4 version 6.3.1 should fix this issue. Thanks again for the report and sorry about the inconvenience!

alexcc4:
Thanks for quick reply!

beniwohli:
@alexcc4 make sure to update to  6.3.2, the fix in 6.3.1 was faulty :(

alexcc4:
@beniwohli  OK~

PR #1187, opened 2021-07-07
Fix Starlette/FastAPI middleware for non-http scopes

### LINKED ISSUE #1186: assert scope["type"] == "http"`on websocket connection
`
    ...lib/python3.9/site-packages/elasticapm/contrib/starlette/__init__.py", line 135, in __call__
    assert scope["type"] == "http"
AssertionError
`

**To Reproduce**

```py
import uvicorn
from elasticapm.contrib.starlette import make_apm_client, ElasticAPM
from fastapi import FastAPI, WebSocket

apm = make_apm_client({'SERVICE_NAME': 'test', 'SERVER_URL': 'http://0.0.0.0:8200'})
app = FastAPI()


@app.websocket('/ws/')
async def websocket_endpoint(
        websocket: WebSocket,
):
    await websocket.accept()


if __name__ == '__main__':
    uvicorn.run(app, host='0.0.0.0', port=18000)

```

run in Chrome console
```js
socket = new WebSocket('ws://127.0.0.1:18000/ws/');
```

- OS: MacOs
- Python version: 3.9
- Framework and version: FastAPI
- APM Server version: 3.6.1

basepi:

Raising an error as we do in #1185 is not the correct behavior for non-http scopes in our middleware. Just because we don't handle non-http scopes doesn't mean Starlette/FastAPI can't. Instead of raising an exception, we just need to skip all of our processing and await the app and immediately return.

I verified that this works as a fix for both #1184 and #1186.

## Related issues
Closes #1186

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -30,6 +30,20 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 6.x" heading
+[float]
+===== Features
+
+
+[float]
+===== Bug fixes
+
+* Fix handling of non-http scopes in Starlette/FastAPI middleware {pull}1187[#1187]
+
+
 [[release-notes-6.x]]
 === Python Agent version 6.x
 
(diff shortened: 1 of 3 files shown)


beniwohli (approved):
LGTM!

PR #1192, opened 2021-07-12
Add docs for manual distributed tracing

basepi:

The obvious next step when looking at manual instrumentation is propagating TraceParent for distributed tracing. Although we have some public API functions that help this, I figured it would be good to distill it down with some examples in our [custom code](https://www.elastic.co/guide/en/apm/agent/python/current/instrumenting-custom-code.html) documentation.


beniwohli (approved):
Great stuff, LGTM!

PR #1193, opened 2021-07-13
AWS Lambda support

basepi:
## Related issues
Ref #428

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -147,7 +147,7 @@ class Client(object):
             constants.EVENTS_API_PATH,
         )
         transport_class = import_string(self.config.transport_class)
-        self._transport = transport_class(self._api_endpoint_url, self, **transport_kwargs)
+        self._transport = transport_class(url=self._api_endpoint_url, client=self, **transport_kwargs)
         self.config.transport = self._transport
         self._thread_managers["transport"] = self._transport
 
@@ -633,3 +633,23 @@ def set_client(client):
         logger = get_logger("elasticapm")
         logger.warning("Client object is being set more than once", stack_info=True)
     CLIENT_SINGLETON = client
+
+
+class ServerlessClient(Client):
+    """
+    Custom client for serverless applications, where we don't want any
+    background threads and need to dump messages to logs rather than to the
+    APM server directly.
+    """
+
+    def __init__(self, config=None, **inline):
+        inline["transport_class"] = "elasticapm.transport.serverless.ServerlessTransport"
+        if isinstance(config, dict) and "transport_class" in config:
+            config.pop("transport_class")
+        super(ServerlessClient, self).__init__(config, **inline)
+
+    def start_threads(self):
+        """
+        No background threads for serverless
+        """
+        pass
--- a/elasticapm/contrib/serverless/__init__.py
+++ b/elasticapm/contrib/serverless/__init__.py
@@ -1,0 +1,41 @@
+#  BSD 3-Clause License
+#
+#  Copyright (c) 2019, Elasticsearch BV
+#  All rights reserved.
+#
+#  Redistribution and use in source and binary forms, with or without
+#  modification, are permitted provided that the following conditions are met:
+#
+#  * Redistributions of source code must retain the above copyright notice, this
+#    list of conditions and the following disclaimer.
+#
+#  * Redistributions in binary form must reproduce the above copyright notice,
+#    this list of conditions and the following disclaimer in the documentation
+#    and/or other materials provided with the distribution.
+#
+#  * Neither the name of the copyright holder nor the names of its
+#    contributors may be used to endorse or promote products derived from
+#    this software without specific prior written permission.
+#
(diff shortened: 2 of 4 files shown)


elasticapm/conf/__init__.py
```diff
@@ -563,6 +563,7 @@ class Config(_ConfigBase):
     prometheus_metrics = _BoolConfigValue("PROMETHEUS_METRICS", default=False)
     prometheus_metrics_prefix = _ConfigValue("PROMETHEUS_METRICS_PREFIX", default="prometheus.metrics.")
     disable_metrics = _ListConfigValue("DISABLE_METRICS", type=starmatch_to_regex, default=[])
+    disable_metrics_thread = _BoolConfigValue("DISABLE_METRICS_THREAD", default=False)
```
beniwohli, line 566: In other parts of the code, we take a metrics interval of 0 as an indication that the metrics thread should not be started, e.g.

https://github.com/elastic/apm-agent-python/blob/b33f4494d5120612c43b8ee76aa26833543ad65a/elasticapm/metrics/base_metrics.py#L94

We should document that fact though, currently it isn't: https://github.com/elastic/apm-agent-python/blob/fdc0bf03a5bd46449c5ab579accb6514b2c8a2b6/docs/configuration.asciidoc#metrics_interval

basepi: Good callout, thanks.

elasticapm/utils/time.py
```diff
@@ -0,0 +1,81 @@
...
+    `time.time()`) to the "equivalent" result from `time.perf_counter()`.
+
+    Note that because these functions vary in their resolution and tick rate,
+    this is only a close approximation.
+
+    Note also that because `time.time()` is *usually* monotonic (but not
+    guaranteed), if a system clock is changed, this function could become
+    very inaccurate.
+    """
+    global CLOCK_DIFF
```
beniwohli, line 49: Is this `global` necessary? You're not changing the value in this function

elasticapm/utils/time.py
```diff
@@ -0,0 +1,81 @@
...
+
+        tolerance = (time2 - time1) / 2
+        timetime = time1 + tolerance
+
+        if tolerance < best_tolerance:
+            best_tolerance = tolerance
+            CLOCK_DIFF = perftime - timetime
+
+        if tolerance < 0.00001:  # try to get the two time.time() calls within 20 microseconds
+            break
```
beniwohli, line 81: Would it make sense to log something if we do not manage to achieve the desired tolerance in 10 tries?

elasticapm/utils/time.py
```diff
@@ -0,0 +1,81 @@
...
+def time_to_perf_counter(timestamp):
+    """
+    This function converts a given epoch timestamp in seconds (typically from
+    `time.time()`) to the "equivalent" result from `time.perf_counter()`.
+
+    Note that because these functions vary in their resolution and tick rate,
+    this is only a close approximation.
+
+    Note also that because `time.time()` is *usually* monotonic (but not
+    guaranteed), if a system clock is changed, this function could become
```
beniwohli, line 46: Would it make sense to periodically update `CLOCK_DIFF` (e.g. on every nth call of `time_to_perf_counter`)?

beniwohli (commented):
Regarding `time_to_perf_counter`, it generally LGTM. I worry however that especially in an environment like Lambda, where the runtime is put to sleep and woken up on a regular basis, clock skew could be a phenomenon that is more regular than usual. But that's just a hunch...

elasticapm/utils/time.py
basepi: Maybe so. For lambda it's not necessary, because the runtime is not long-lived. But if we're going to use something like this for the otel PR that's open, we'll probably want to be able to invalidate previous findings.

basepi:
> Regarding time_to_perf_counter, it generally LGTM. I worry however that especially in an environment like Lambda, where the runtime is put to sleep and woken up on a regular basis, clock skew could be a phenomenon that is more regular than usual. But that's just a hunch...

I don't think that's actually an issue in this case. `perf_counter` is designed to be monotonic and take sleep into account. `time.time()` is designed to be "real-world" (take sleep into account) the same as `perf_counter`, and will be monotonic absent a system clock change. I can't imagine we will have to deal with system clock changes in AWS but if we do, we might have a weird outlier but there's not much we can do about that.

basepi:
Did some testing today and the conversion measurement we take at the beginning of the cold start appears to stay accurate throughout multiple invocations, with freezes in between.

I'm still going to implement some re-measurement for long-running processes but I don't think we need to measure every time with Lambda like I was fearing.

elasticapm/utils/time.py
basepi: Decided to make it based on the time elapsed, rather than on the number of calls. I set it to 5 minutes.

Again, it should never fall out of sync unless someone changes the system time which I don't see happening on most modern systems. But it's worth handling that case since this is pretty lightweight. In this case only 5 minutes of data will ever be out of sync.

PR #1194, opened 2021-07-16
fetch server version on transport thread start

beniwohli:

this will allow putting new features behind version guards, e.g.
only sending histogram metrics to server versions that understand
it

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -102,7 +102,8 @@ class Client(object):
         self.processors = []
         self.filter_exception_types_dict = {}
         self._service_info = None
-        self.server_version = None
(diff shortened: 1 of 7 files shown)


basepi (approved):

beniwohli:
@basepi are you :+1: with how I implemented an option to disable version fetching in tests? See e83866ab4

Without the change, a bunch of tests would need to be updated, as they have assertions on number of requests or similar that would change with this initial http request.

basepi:
> @basepi are you 👍 with how I implemented an option to disable version fetching in tests? See e83866a

Yep! LGTM.

PR #1196, opened 2021-07-20
limit the transport exception message to 10000 characters

beniwohli:
In cases where the response from APM Server is very large,
the log is spammed with extremely long log messages.


beniwohli:
Alternatively, we could inspect the response JSON instead of simply dumping it, then we could also print some information on accepted/error counts....

basepi (approved):

basepi:
> Alternatively, we could inspect the response JSON instead of simply dumping it, then we could also print some information on accepted/error counts....

Probably better overall but this is a good first step.

PR #1301, opened 2021-08-18
Fix cookie sanitization if Cookie is capitalized

### LINKED ISSUE #1297: Cookie value from request header is not getting sanitized
Hi, 

The APM python agent is silently throwing error when trying to sanitize the cookie value from request header. On debugging, I found that we are reading the cookie value incorrectly from context. 
 
Instead of  reading the cookie value using `event["context"]["request"]["headers"]["Cookie"]`, we read it as `event["context"]["request"]["headers"]["cookie"]`. 

https://github.com/elastic/apm-agent-python/blob/31c858595db709c766226e5cddb826d27e377d9d/elasticapm/processors.py#L116

basepi:

Sometimes, the `cookie` header is actually `Cookie`, so check for both if there's an error.

## Related issues
Closes #1297

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,11 +3,13 @@ repos:
     rev: 5.7.0
     hooks:
     -   id: isort
--   repo: https://github.com/ambv/black
-    rev: 20.8b1
+-   repo: local
     hooks:
     -   id: black
-        language_version: python3
+        name: black
+        entry: black --check
+        language: system
+        types: [python]
 -   repo: https://gitlab.com/pycqa/flake8
     rev: 3.8.4
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,13 +3,11 @@ repos:
     rev: 5.7.0
     hooks:
     -   id: isort
--   repo: local
+-   repo: https://github.com/ambv/black
+    rev: 20.8b1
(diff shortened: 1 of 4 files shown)


beniwohli:
I wonder if we should normalize the key for the `Cookie` header at the places where we capture the headers. HTTP headers are case-insensitive, so a client could send them as `cOokie` and sanitization would break again. Same with `Set-Cookie` in the response headers (but less likely, as response headers are not client settable).

The drawback is that we may lose a tiny bit of information ("with which upper/lowercase variant did the client write `Cookies`?"), but I think that's acceptable.

basepi:
I disagree. We capture headers in a bunch of places, and only sanitize them in one place. Unless this type of code pattern ends up in more than one place, I don't think adding a whole bunch of normalization code is worth the churn.

Unless I'm missing an easy place to do it centrally.

beniwohli:
@basepi I agree that `cookie` and `Cookie` will cover virtually all cases in the wild, and normalizing it to one casing is tedious as we'd have to do it for all framework integrations.

An option might be to introduce a `dict` subclass that normalizes string keys to a standard casing and use that when capturing headers. But probably not worth the effort.

beniwohli (approved):

basepi:
I do like the idea of having a class which does this for us, we'll keep that in mind if this comes up again in the wild.

PR #1316, opened 2021-08-31
[WIP] Azure SDK Instrumentation

basepi:
## Related issues
#1027
#1028 
#1029 
#1030

### DIFF
--- a/elasticapm/instrumentation/packages/azure.py
+++ b/elasticapm/instrumentation/packages/azure.py
@@ -1,0 +1,122 @@
+#  BSD 3-Clause License
+#
+#  Copyright (c) 2019, Elasticsearch BV
+#  All rights reserved.
+#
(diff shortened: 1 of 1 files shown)


basepi:
Note that the tests included are not actually being run (yet) -- see #1348 for that technical debt. But they all pass if you set up your own `"AZURE_STORAGE_CONNECTION_STRING"` env variable with the connection string for an Azure storage account.

beniwohli (approved):
Those are some humongous if-statements :D good work on assembling those! Would it make sense to write unit tests (as in, they can run without an azure service) for those? Maybe even driven by a JSON file that can be shared amongst agents.

basepi:
Excellent idea. I decided to put the issue in the apm repo since it's a cross-agent thing: https://github.com/elastic/apm/issues/519

PR #1321, opened 2021-09-09
Implemented span compression algorithm

### LINKED ISSUE #1152: [META 432] Implement compressed spans algorithm
Implement the compressed spans algorithm in the agent as described in https://github.com/elastic/apm/issues/432.

beniwohli:

See https://github.com/elastic/apm/issues/432

## Related issues
closes #1152

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -35,12 +35,13 @@ import threading
 import time
 import timeit
 from collections import defaultdict
+from typing import Callable, Optional, Union
 
 from elasticapm.conf import constants
 from elasticapm.conf.constants import LABEL_RE, SPAN, TRANSACTION
 from elasticapm.context import init_execution_context
 from elasticapm.metrics.base_metrics import Timer
-from elasticapm.utils import compat, encoding, get_name_from_func
+from elasticapm.utils import compat, encoding, get_name_from_func, nested_key
 from elasticapm.utils.disttracing import TraceParent, TracingOptions
 from elasticapm.utils.logging import get_logger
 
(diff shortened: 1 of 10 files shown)


elasticapm/traces.py
```diff
@@ -481,13 +557,63 @@ def end(self, skip_frames=0, duration=None):
...
+            self._child_durations.stop(self.start_time + self.duration)
             self.transaction.track_span_duration(
                 self.type, self.subtype, self.duration - self._child_durations.duration
             )
+        p.child_ended(self)
+
+    def report(self) -> None:
+        self.tracer.queue_func(SPAN, self.to_dict())
+
+    def try_to_compress(self, sibling: SpanType) -> bool:
```
basepi, line 572: Maybe some docstrings through here so we can remember at a glance what composite vs regular means?

basepi (commented):
Did a brief review, everything looks good so far! I'll do a deeper review once this is ready for review.

I noticed that you haven't really refactored `leaf` to `exit` -- is that still something we want to do? I made a note [here](https://github.com/elastic/apm-agent-python/issues/1152#issuecomment-892862867) after one of our previous discussions.

beniwohli:
> I noticed that you haven't really refactored `leaf` to `exit` -- is that still something we want to do? I made a note [here](https://github.com/elastic/apm-agent-python/issues/1152#issuecomment-892862867) after one of our previous discussions.

I started it implementing in this branch, but soon realized that it would overload it (something that also became apparent with introducing type hints with this PR, hence I only committed the ones in `tracing.py`). The plan is to implement it in #1159, which is a smaller change in general, but also depends on an `exit` flag.

basepi (approved):
LGTM! :+1:

PR #1327, opened 2021-09-15
Dropped span statistics

### LINKED ISSUE #1310: [META 497] Collecting statistics about dropped spans
As described in [this spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-dropped-stats.md), agents should collect statistics on dropped spans.

Meta issue: [issue details](https://github.com/elastic/apm/issues/497)

beniwohli:
## Related issues
closes #1310

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -421,13 +421,6 @@ class Span(BaseSpan):
         self.parent_span_id = parent_span_id
         self.frames = None
         self.sync = sync
-        if span_subtype is None and "." in span_type:
-            # old style dottet type, let's split it up
-            type_bits = span_type.split(".")
-            if len(type_bits) == 2:
-                span_type, span_subtype = type_bits[:2]
-            else:
-                span_type, span_subtype, span_action = type_bits[:3]
         self.type = span_type
         self.subtype = span_subtype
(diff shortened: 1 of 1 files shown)


elasticapm/traces.py
```diff
@@ -664,6 +657,13 @@ def __init__(
         sync=None,
     ):
         self.name = name
+        if span_subtype is None and "." in span_type:
+            # old style dottet type, let's split it up
```
basepi, line 661: nit

```suggestion
            # old style dotted type, let's split it up
```

basepi (approved):
Looks great! One nit.

Definitely not going to play well with my lambda changes to allow for easier backdating of spans/transactions but life will go on. Maybe I'll drop the backdating stuff since we're not actually doing that in lambda right anymore, at least for now.

beniwohli:
@basepi just for my understanding, how would this PR make backdating of spans more difficult in Lambda?

basepi:
Oh, it won't make it more difficult, it will just merge conflict and I'll have to move some stuff around. I'm torn on whether I should even include backdating since we're not actually using it anymore. Some day it may be useful but I don't know?

PR #1328, opened 2021-09-16
Fixed AsyncPG Integration who missed a lot of spans

zikphil:
Hey guys,

So I've noticed that the ElasticAPM Python Agent was not binding itself to the right methods to gather AsyncPG data. I've taken a look at how the NewRelic Agent did it and came up with this modification. So instead of using the Connection object, which is sometimes not used (i.e: sqlalchemy) Im binding at the protocol layer to ensure no spans are missed.

I have been running this on production for a while now and it's great :)

Enjoy

### DIFF
--- a/elasticapm/instrumentation/packages/asyncio/asyncpg.py
+++ b/elasticapm/instrumentation/packages/asyncio/asyncpg.py
@@ -49,18 +49,13 @@ class AsyncPGInstrumentation(AsyncAbstractInstrumentedModule):
         ("asyncpg.protocol.protocol", "Protocol.bind_execute_many"),
         ("asyncpg.protocol.protocol", "Protocol.bind"),
         ("asyncpg.protocol.protocol", "Protocol.execute"),
-        ("asyncpg.protocol.protocol", "Protocol.prepare"),
         ("asyncpg.protocol.protocol", "Protocol.query"),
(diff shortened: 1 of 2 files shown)


basepi (approved):
This looks awesome! Thank you!

basepi:
Looks like I can't push changes to your branch, so you'll have to merge `upstream/master` here so it's up to date with the base branch. Then I'll tell the tests to run.

basepi:
Looks like we do have some failures. https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/PR-1328/7/tests

It may be that we're capturing spans that weren't being captured before, and that the tests need to be updated. I'm not certain.

zikphil:
I have fixed the tests and re-merged master. One thing to note is that for a `SELECT X FROM Y` database operation before we used to capture 1 span. Now since we are integrated at a lower level, we will have 2 spans, 1 for the PREPARE statement and 1 for the query. Just wanted to outline that for people who have the integration already, it will double the amount of spans.

basepi:
Can we actually remove the instrumentation around `PREPARE`? We don't create a span for `PREPARE` statements in any of our other database instrumentations and I want to keep things consistent. You'll also want to revert your test changes. Sorry for the confusion!

PR #1345, opened 2021-09-28
Pass custom client to aiohttp tracing middleware

stj:

`elasticapm.contrib.aiohttp.ElasticAPM` allows an optional client
instance as argument, though that client is not used in the tracing
middleware. Instead, it always uses the `base` modules client singleton.

This change passes the client object to the tracing middleware allowing
for a consistent use of a custom client object.


beniwohli (approved):
Good catch, thanks for the PR!

PR #1349, opened 2021-10-01
Handle TypeError for missing server version

basepi:
If an HTTP response comes back from the "server" but isn't formed correctly, we can get the following error:

```
version = data["version"]
TypeError: 'int' object is not subscriptable
```

This came up when testing against the new (upcoming) lambda extension. Apparently it responded, but not with useful data. This PR fixes that error.

## Related issues
Ref #1194


beniwohli (approved):
Would you mind adding a test for this case? :)

PR #1368, opened 2021-10-18
[WIP] More lambda improvements + docs

basepi:
## Related issues
Ref #1193

### DIFF
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -63,9 +63,6 @@ class capture_serverless(object):
         @capture_serverless()
         def handler(event, context):
             return {"statusCode": r.status_code, "body": "Success!"}
-
-    Note: This is an experimental feature, and we may introduce breaking
-    changes in the future.
     """
 
     def __init__(self, name=None, **kwargs):
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -121,10 +121,13 @@ class capture_serverless(object):
         transaction_type = "request"
         transaction_name = os.environ.get("AWS_LAMBDA_FUNCTION_NAME", self.name)
 
-        if "httpMethod" in self.event:  # API Gateway
+        self.httpmethod = self.event.get("requestContext", {}).get(
+            "httpMethod", self.event.get("requestContext", {}).get("http", {}).get("method")
+        )
+        if self.httpmethod:  # API Gateway
             self.source = "api"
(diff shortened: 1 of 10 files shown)


elasticapm/contrib/serverless/aws.py
```diff
@@ -121,10 +121,13 @@ def __enter__(self):
         transaction_type = "request"
         transaction_name = os.environ.get("AWS_LAMBDA_FUNCTION_NAME", self.name)
 
-        if "httpMethod" in self.event:  # API Gateway
+        self.httpmethod = self.event.get("requestContext", {}).get(
```
beniwohli, line 124: I added a utility function to get deeply nested dict keys with one of my recent PRs:

https://github.com/elastic/apm-agent-python/blob/a06f74f6da75554bed0ea609f3f327721108674e/elasticapm/utils/__init__.py#L200-L207

I should also add a docstring that explains the usage, but it should be something like

    self.httpmethod = nested_key(self.event, "requestContext", "httpMethod") or nested_key(self.event, "requestContext", "http", "method")

elasticapm/contrib/serverless/aws.py
```diff
@@ -367,21 +377,23 @@ def get_url_dict(event):
     Reconstruct URL from API Gateway
     """
     headers = event.get("headers", {})
-    proto = headers.get("X-Forwarded-Proto", "https")
-    host = headers.get("Host", "")
-    path = event.get("path", "")
-    port = headers.get("X-Forwarded-Port")
+    protocol = headers.get("X-Forwarded-Proto", headers.get("x-forwarded-proto", "https"))
```
beniwohli, line 380: you could wrap `headers` in a [`wsgiref.headers.Headers`](https://docs.python.org/3/library/wsgiref.html#wsgiref.headers.Headers) object, which would handle the case-insensitiveness aspect of HTTP headers. It's a somewhat dumb implementation, that loops over all headers on a `.get()`. If you know that the two variants you handle are the only two possible, it's probably not worth the hassle.

beniwohli (approved):

elasticapm/contrib/serverless/aws.py
basepi: Super good to know, but in this case there are only a couple of instances (which only have two case variations) so I probably won't bother.

elasticapm/contrib/serverless/aws.py
basepi: Oh right! Thank you I had forgotten that.

PR #1376, opened 2021-11-03
count spans only when they are reported

beniwohli:
this should avoid counting spans that are compressed


beniwohli:
Turns out this is not a great way to achieve this, as it favors inner spans in the count. E.g. if max_spans is 2 and we have nested spans like

    span1:
        span2:
            span3:

`span1` would be the one that is turned into a dropped span.

The alternative is to subtract from `transaction._span_counter` when a span is compressed. This is implemented in #1377

PR #1377, opened 2021-11-03
decrement span_counter when span is compressed

beniwohli:
this ensures that compressed spans are not counted against
transaction_max_spans

alternative for #1376


basepi (approved):
Seems like a reasonable solution!

PR #1382, opened 2021-11-04
More AWS Lambda improvements and spec changes

basepi:
Better error handling for the `Xxx` conversion, and remove case-insensitive `statusCode` handling, as anything besides `statusCode` is invalid and will throw a server error.

Edit: Also added the field changes from spec PR here: https://github.com/elastic/apm/pull/540

### DIFF
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -245,9 +245,9 @@ class capture_serverless(object):
             cloud_context["origin"]["region"] = record["awsRegion"]
             cloud_context["origin"]["account"] = {"id": record["eventSourceARN"].split(":")[4]}
             cloud_context["origin"]["provider"] = "aws"
-            message_context["queue"] = record["eventSourceARN"]
+            message_context["queue"] = {"name": record["eventSourceARN"]}
             if "SentTimestamp" in record["attributes"]:
-                message_context["age"] = int((time.time() * 1000) - int(record["attributes"]["SentTimestamp"]))
+                message_context["age"] = {"ms": int((time.time() * 1000) - int(record["attributes"]["SentTimestamp"]))}
             if self.client.config.capture_body in ("transactions", "all") and "body" in record:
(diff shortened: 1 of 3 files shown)


elasticapm/contrib/serverless/aws.py
```diff
@@ -171,9 +167,18 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
+                "response",
+            )
             if status_code:
-                result = "HTTP {}xx".format(int(status_code) // 100)
-                elasticapm.set_transaction_result(result, override=False)
+                try:
+                    result = "HTTP {}xx".format(int(status_code) // 100)
+                    elasticapm.set_transaction_result(result, override=False)
+                except ValueError:
+                    elasticapm.set_transaction_result("HTTP 2xx", override=False)
```
beniwohli, line 181: could you add a comment here lining out what causes the `ValueError`?

basepi: Done. Casting to an `int` throws a `ValueError` if it's unsuccessful.

basepi:
Thanks to @trentm for pointing out my bad assumption about case-sensitivity of `statusCode`. This is now ready for review again.

elasticapm/contrib/serverless/aws.py
```diff
@@ -65,7 +65,7 @@ def handler(event, context):
             return {"statusCode": r.status_code, "body": "Success!"}
     """
 
-    def __init__(self, name=None, **kwargs):
+    def __init__(self, name: str = None, **kwargs) -> None:
```
beniwohli, line 68: this is type `Optional[str]` if its default value is `None`

beniwohli (approved):
just a tiny typing nit, otherwise LGTM

elasticapm/contrib/serverless/aws.py
basepi: 🤦‍♂️ Honestly don't know how I missed that.

PR #1385, opened 2021-11-08
add github action that runs black and creates suggestions using the diff

beniwohli:
## Related issues
closes #ISSUE


beniwohli:
ugh, it's still impossible to get black output as review suggestions :/

PR #1386, opened 2021-11-08
fix issue with multiple hosts/ports in psycopg2 connect span

### LINKED ISSUE #1383: APM crashes when postgres is configured with multiple hosts / ports in Django
Hello.

I'm using multiple postgresql hosts (https://www.postgresql.org/docs/current/libpq-connect.html#LIBPQ-CONNSTRING section 34.1.1.3).

The APM crashes trying to convert comma separated list of ports in a int:

```
  File "/usr/local/lib/python3.7/site-packages/elasticapm/instrumentation/packages/psycopg2.py", line 86, in call
    if int(port) != default_ports.get("postgresql"):
ValueError: invalid literal for int() with base 10: '5432,5433'
```

**To Reproduce**

1. Configure multiple hosts in postgres config (more than 1 host and port)
2. Start the app
3. crashes with the exception:
```
  File "/usr/local/lib/python3.7/site-packages/elasticapm/instrumentation/packages/psycopg2.py", line 86, in call
    if int(port) != default_ports.get("postgresql"):
ValueError: invalid literal for int() with base 10: '5432,5433'
```

- OS: Linux containerd 
- Python version: 3.7
- Framework and version [e.g. Django 2.1]: Django 2.1
- APM Server version: 7.12.1
- Agent version: 6.6.1

**Additional context**

https://github.com/elastic/apm-agent-python/blob/master/elasticapm/instrumentation/packages/psycopg2.py#L86

beniwohli:
fixes #1383

### DIFF
--- a/elasticapm/instrumentation/packages/psycopg2.py
+++ b/elasticapm/instrumentation/packages/psycopg2.py
@@ -29,6 +29,8 @@
 #  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 from __future__ import absolute_import
 
+from typing import Optional, Union
+
 from elasticapm.instrumentation.packages.dbapi2 import (
     ConnectionProxy,
     CursorProxy,
@@ -76,22 +78,11 @@ class Psycopg2Instrumentation(DbApi2Instrumentation):
     def call(self, module, method, wrapped, instance, args, kwargs):
         signature = "psycopg2.connect"
 
-        host = kwargs.get("host")
-        if host:
-            signature += " " + compat.text_type(host)
-
-            port = kwargs.get("port")
-            if port:
-                port = str(port)
-                if int(port) != default_ports.get("postgresql"):
-                    host += ":" + port
-            signature += " " + compat.text_type(host)
-        else:
-            # Parse connection string and extract host/port
-            pass
+        host, port = get_destination_info(kwargs.get("host"), kwargs.get("port"))
+        signature = f"{signature} {host}:{port}"
         destination_info = {
-            "address": kwargs.get("host", "localhost"),
-            "port": int(kwargs.get("port", default_ports.get("postgresql"))),
+            "address": host,
+            "port": port,
         }
         with capture_span(
             signature,
@@ -140,3 +131,18 @@ class Psycopg2ExtensionsInstrumentation(DbApi2Instrumentation):
                 kwargs["scope"] = kwargs["scope"].__wrapped__
 
         return wrapped(*args, **kwargs)
+
+
+def get_destination_info(host: Optional[str], port: Union[None, str, int]) -> tuple:
+    if host:
+        if "," in host:  # multiple hosts defined, take first
+            host = host.split(",")[0]
+    else:
+        host = "localhost"
+    if port:
+        port = str(port)
+        if "," in port:  # multiple ports defined, take first
+            port = int(port.split(",")[0])
+    else:
+        port = default_ports.get("postgresql")
+    return host, port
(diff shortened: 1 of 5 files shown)


elasticapm/instrumentation/packages/psycopg2.py
```diff
@@ -140,3 +131,18 @@ def call(self, module, method, wrapped, instance, args, kwargs):
...
+def get_destination_info(host: Optional[str], port: Union[None, str, int]) -> tuple:
+    if host:
+        if "," in host:  # multiple hosts defined, take first
+            host = host.split(",")[0]
+    else:
+        host = "localhost"
+    if port:
+        port = str(port)
+        if "," in port:  # multiple ports defined, take first
+            port = int(port.split(",")[0])
```
basepi, line 145: nit: Any particular reason you cast back to `int` here? The result is that a single port will be a `str`, but multiple ports will result in an `int`.

tests/fixtures.py
```diff
@@ -201,6 +201,12 @@ def elasticapm_client(request):
     assert not client._transport.validation_errors
 
 
+@pytest.fixture()
+def elasticapm_transaction(elasticapm_client):
+    transaction = elasticapm_client.begin_transaction("test")
+    yield transaction
```
basepi, line 207: Do we need this fixture? It's only saving you one line in the test you wrote, since you end the transaction and inspect the spans in the end anyway. Actually, since you have to do `get_client()`, it's saving you no lines...

tests/instrumentation/psycopg2_tests.py
```diff
@@ -507,3 +510,18 @@ def test_psycopg2_execute_values(instrument, postgres_connection, elasticapm_cli
         spans = elasticapm_client.spans_for_transaction(transactions[0])
         assert spans[0]["name"] == "INSERT INTO test"
         assert len(spans[0]["context"]["db"]["statement"]) == 10000, spans[0]["context"]["db"]["statement"]
+
+
+@pytest.mark.integrationtest
+def test_psycopg2_connection(instrument, elasticapm_transaction, postgres_connection):
+    elasticapm_client = cast(TempStoreClient, get_client())
```
basepi, line 517: What is the purpose of this line? From the docs:

> Cast a value to a type.
>
> This returns the value unchanged. To the type checker this signals that the return value has the designated type, but at runtime we intentionally don’t check anything (we want this to be as fast as possible).

Seems like it doesn't do anything, and I can't see how the type checker cares here.

basepi (approved):
Couple of questions below.

tests/fixtures.py
beniwohli: the idea is that this way the transaction starts _before_ the `postgres_connection` fixture runs, which enables us to capture the `connect` span.

I can add a comment to explain why it is useful.

basepi: Ohhhh, thank you. Yes, please add a comment.

tests/instrumentation/psycopg2_tests.py
beniwohli: a line down, we access `elasticapm_client.events`, which is only available on the `TempStoreClient`. Without it, PyCharm (and mypy, once we get around to activate that) will complain.

basepi: Thanks! Can you add a comment here as well? Something like

```
# elastciapm_client.events is only available on `TempStoreClient`, this keeps the type checkers happy
```

elasticapm/instrumentation/packages/psycopg2.py
beniwohli: good catch! It should be cast to int no matter what

PR #1389, opened 2021-11-09
Cast transaction name to str

### LINKED ISSUE #1167: Force transaction name in `set_transaction_name` to string
**Describe the bug**: 

When providing a non-string value to `elasticapm.set_transaction_name()`, APM Server discards the transaction.

To avoid this, we should try and force the value to a string, raising a warning if it isn't possible.

**To Reproduce**

1. start a transaction
2. call `elasticapm.set_transaction_name(["a list"])`
3. transaction fails schema validation

- OS: [e.g. Linux] Linux
- Python version: 3.9
- APM Server version: 7.13
- Agent version: 6.2.2

basepi:
I decided not to do any error handling around casting to `str()` as suggested in #1167 -- while theoretically it can throw errors, the user would have to be doing something exceptionally odd to make that happen and I don't think it's worth muddying the code. We cast to `str()` regularly without error handling.

## Related issues
Closes #1167

### DIFF
--- a/tests/instrumentation/transactions_store_tests.py
+++ b/tests/instrumentation/transactions_store_tests.py
@@ -352,13 +352,13 @@ def test_set_transaction_name(elasticapm_client):
 
     elasticapm_client.begin_transaction("test")
 
-    elasticapm.set_transaction_name("another_name")
+    elasticapm.set_transaction_name(["another_name"])
 
     elasticapm_client.end_transaction("test_name", 200)
 
     transactions = elasticapm_client.events[TRANSACTION]
     assert transactions[0]["name"] == "test_name"
-    assert transactions[1]["name"] == "another_name"
+    assert transactions[1]["name"] == "['another_name']"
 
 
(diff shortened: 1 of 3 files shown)


tests/instrumentation/transactions_store_tests.py
```diff
@@ -352,13 +352,13 @@ def test_set_transaction_name(elasticapm_client):
 
     elasticapm_client.begin_transaction("test")
 
-    elasticapm.set_transaction_name("another_name")
+    elasticapm.set_transaction_name(["another_name"])
```
beniwohli, line 355: I'd prefer if string coercing would be its own test (or use `pytest.mark.parametrize`)

beniwohli (approved):
just a small nit, otherwise LGTM

tests/instrumentation/transactions_store_tests.py
basepi: Done.

PR #1411, opened 2021-11-19
opentelemetry bridge

### LINKED ISSUE #1362: [META 524] OpenTelemetry Bridge
Implement https://github.com/elastic/apm/issues/524

basepi:
## Related issues
Closes #1362


elasticapm/context/contextvars.py
```diff
@@ -38,9 +38,15 @@
 
 class ContextVarsContext(BaseContext):
     elasticapm_transaction_var = contextvars.ContextVar("elasticapm_transaction_var")
-    elasticapm_span_var = contextvars.ContextVar("elasticapm_span_var")
+    elasticapm_spans_var = contextvars.ContextVar("elasticapm_spans_var", default=[])
```
basepi, line 41: I have a suspicion that this is not OK -- I'm guessing I end up mutating this default. Will need to move to tuples or at least verify that this is OK.

elasticapm/context/base.py
```diff
@@ -31,13 +31,48 @@
...
+
+        The previously-activated span will be saved to be re-activated later.
+
+        Optionally, `extra` data can be provided and will be saved alongside
+        the span.
+        """
         raise NotImplementedError
 
-    def set_span(self, span):
+    def get_span(self, extra=False):
```
beniwohli, line 59: I think it would help with legibility if the `extra` argument had a different name here compared to `set_span`, as they have a different meaning. Something like `with_extra`. Same in `unset_span`.

elasticapm/context/contextvars.py
```diff
@@ -50,16 +56,60 @@ def get_transaction(self, clear=False):
...
+    def set_span(self, span, extra=None):
+        """
+        Set the active span for the current execution context.
+
+        The previously-activated span will be saved to be re-activated later.
+
+        Optionally, `extra` data can be provided and will be saved alongside
+        the span.
+        """
+        self.elasticapm_spans_var.set(self.elasticapm_span_var.get() + (span, extra))
```
beniwohli, line 89: Shouldn't this be

```suggestion
        self.elasticapm_spans_var.set(self.elasticapm_span_var.get() + ((span, extra),))
```
Adding two tuples creates a new tuple that is the concatenation of the two tuples.

beniwohli (commented):
Looks great! I'd be interesting to know if this has any negative effects on performance. And if it does, I wonder if using a list of tuples instead of a tuple of tuples would help, as we wouldn't be constantly recreating the outer container.

elasticapm/context/contextvars.py
basepi: You're totally right, good catch.

elasticapm/contrib/opentelemetry/span.py
```diff
@@ -0,0 +1,291 @@
...
+        self._set_types()
+        if end_time:
+            if is_transaction:
+                self.client.end_transaction(
+                    name=self._name,
+                    result=self.elastic_span.outcome or "OK",
+                    duration=time.time() - self.elastic_span.timestamp,
+                )
+            else:
+                self.elastic_span.end(duration=time.time() - self.elastic_span.timestamp)
```
beniwohli, line 90: Same `end_time` comment :)

elasticapm/traces.py
```diff
@@ -543,6 +562,22 @@ def to_dict(self) -> dict:
             self.context["tags"] = self.labels
         if self.context:
             self.autofill_resource_context()
+            # otel attributes and spankind need to be top-level
+            if "otel_spankind" in self.context:
+                result["otel"] = {"span_kind": self.context.pop("otel_spankind")}
+            if elasticapm.get_client().check_server_version(gte=(7, 16)):
```
beniwohli, line 568: On [line 401](https://github.com/elastic/apm-agent-python/pull/1411/files#diff-42a05cd7587872ceaa81c35f1bd163e5420820d2cac99c70a32750230df9c508R401), you're checking first if `get_client()` returns something. Is this not necessary here?

tests/context/test_context.py
```diff
@@ -35,17 +35,6 @@
 from elasticapm.context.threadlocal import ThreadLocalContext
 
 
-def test_execution_context_backing():
```
beniwohli, line 38: do we still have to remove this test after you moved opentelemetry to its own matrix entry?

beniwohli (approved):
Some minor comments, other than that LGTM!

elasticapm/contrib/opentelemetry/context.py
```diff
@@ -0,0 +1,144 @@
...
+    otel_span = getattr(span, "otel_wrapper", Span(span.name, span))
+    context = otel_span.otel_context
+
+    return context
+
+
+def attach(context: Context) -> object:
+    """Associates a Context with the caller's current execution unit.
+
+    Due to limitations in the Elastic APM context management, a token is not
```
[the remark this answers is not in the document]
basepi: Should be in both places. `attach` is supposed to return a token.

elasticapm/contrib/opentelemetry/span.py
basepi: Neither spans nor transactions support an `end_time` unless I'm missing something: https://github.com/elastic/apm-agent-python/blob/97188c362497910762e5b1ef8fc53e4011ec75b8/elasticapm/traces.py#L123

elasticapm/contrib/opentelemetry/span.py
```diff
@@ -0,0 +1,291 @@
...
+        if self.elastic_span.ended_time:
+            # Already ended
+            return
+        self._set_types()
+        if end_time:
+            if is_transaction:
+                self.client.end_transaction(
+                    name=self._name,
+                    result=self.elastic_span.outcome or "OK",
+                    duration=time.time() - self.elastic_span.timestamp,
```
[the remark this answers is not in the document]
basepi: Hmmm. As far as I can tell, `end_transaction` doesn't support `end_time`: https://github.com/elastic/apm-agent-python/blob/97188c362497910762e5b1ef8fc53e4011ec75b8/elasticapm/base.py#L298

elasticapm/traces.py
basepi: It shouldn't be needed in either place in normal operation. I had to add it on 401 to get a couple of transaction_store_tests to pass, like: https://github.com/elastic/apm-agent-python/blob/97188c362497910762e5b1ef8fc53e4011ec75b8/tests/instrumentation/transactions_store_tests.py#L300

It doesn't actually instantiate a Client, just uses the Tracer object directly, so the test was failing. I'll add a little comment on 401.

elasticapm/contrib/opentelemetry/span.py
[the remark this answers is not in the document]
beniwohli: I meant the `end_time` that is an argument to this method:

https://github.com/basepi/apm-agent-python/blob/b9ff8d6dadc42b0a88433390bfa3f3729111c8d8/elasticapm/contrib/opentelemetry/span.py#L71-L82

You check if `end_time` is set, but then don't use it. So I'm wondering if 
```
duration=time.time() - self.elastic_span.timestamp,
```

shouldn't be 
```
duration=end_time - self.elastic_span.timestamp,
```

(assuming that `end_time` and `self.elastic_span.timestamp` are of the same time domain)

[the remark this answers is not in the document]
basepi: Oh, yep! Good call. I thought I did that but it slipped past me.

PR #1417, opened 2021-11-29
Unify async_capture_span/capture_span common code

beniwohli:

Some recent changes have only been implemented in `capture_span`, while `async_capture_span` was forgotten. This change should avoid such issues going forward.


basepi (approved):
Good catch!

PR #1418, opened 2021-11-29
implement dropping of fast exit spans

### LINKED ISSUE #1309: [META 496] Dropping fast exit spans
As described in the [Dropping fast exit spans](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-drop-fast-exit.md) spec, agents should introduce the config option `exit_span_min_duration` that defines the minimal duration for exit spans to not being dropped.

The implementation should be aligned with the overall approach described for [Handling huge traces](https://github.com/elastic/apm/tree/master/specs/agents/handling-huge-traces).

Meta issue: [issue details](https://github.com/elastic/apm/issues/496)

beniwohli:
closes #1309


tests/instrumentation/transactions_store_tests.py
```diff
@@ -177,7 +177,7 @@ def tracer():
     def queue(event_type, event, flush=False):
         events[event_type].append(event)
 
-    store = Tracer(lambda: frames, lambda frames: frames, queue, Config(), None)
+    store = Tracer(lambda: frames, lambda frames: frames, queue, elasticapm_client.config, elasticapm_client)
```
beniwohli, line 180: this change was needed so that the tests that use this fixture also respect the default settings we set up in the `elasticapm_client` fixture

elasticapm/conf/__init__.py
```diff
@@ -590,6 +590,12 @@ class Config(_ConfigBase):
         validators=[duration_validator],
         type=int,
     )
+    exit_span_min_duration = _ConfigValue(
+        "exit_span_min_duration",
+        default=1,
+        validators=[duration_validator],
```
basepi, line 596: From the [spec](https://github.com/elastic/apm/blob/master/specs/agents/handling-huge-traces/tracing-spans-drop-fast-exit.md):

> The minimum allowed duration for this setting is `1us` (microsecond). Agents may need to add support the `us` unit.

We still need to add support for microseconds, though I don't think it's critical to have in this PR.

basepi (approved):
I had a hunch this was where the conversation about adding microseconds happened, so I checked the spec and added one note below. Otherwise looks great!

basepi (approved):

PR #1419, opened 2021-11-30
chore(ci): use filebeat step to archive docker container logs

mdelapenya:
It uses the filebeat step (https://github.com/elastic/apm-pipeline-library/blob/master/vars/filebeat.groovy) which grabs all Docker container logs that ran within the Closure block, archiving them into Jenkins UI

## Follow-up concerns
Should we store all logs or limit archiving them only on failures using the `archiveOnlyOnFail: true` attribute for the filebeat step?

## Related issues
closes #ISSUE


v1v (approved):

beniwohli (approved):
Thanks for the house keeping! I think it would be ok to only archive logs on failures. I guess there might be scenarios where it would be helpful to compare logs of a failed build with logs from a build that passed, but I can't remember such an occurrence since we started using Jenkins.

mdelapenya:
I confirm the logs are stored under Jenkins Artifacts tab, so it's possible to look them up on failures in an easier manner, i.e. downloading them

mdelapenya:
Test errors (MongoDB-related) seem unrelated and present in master, if you agree, let's merge this one

beniwohli:
Merge is 👍 

I don't necessarily mean an automatic comparison, but it might e.g. be interesting to see if the passed build installed other dependencies from pypi compared with the failed build. But as I said, that's mostly a theory, I don't think it ever came in handy in real life :D

PR #1403, opened 2021-11-16
refactor httpx instrumentation

beniwohli:
httpx 0.21 and httpcore 0.14 once again changed how it works,
necessitating more special handling. To make this a tad less
cumbersome, most of this code has moved into a utils file.
To make the structure somewhat sane, httpx got its own
package in instrumentation.

closes #1399
closes #1400
closes #1401

PR #1423, opened 2021-12-01
Fix AttributeError in sync httpx instrumentation

### LINKED ISSUE #1422: Breaking change in sync httpx refactor from #1403
#1403 refactored a lot of the httpx code. However it seems that the sync version of the instrumentation still calls `self._set_disttracing_headers` which was removed in the same PR, causing any form of sync httpx instrumentation to crash.

- Agent version: 6.7.1

HenrikOssipoff:

#1403 introduced a breaking change due to a refactor. This should make the sync code in line with the async version and fix the AttributeError.

## Related issues
closes #1422

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -28,7 +28,7 @@ endif::[]
 //
 [float]
 ===== Bug fixes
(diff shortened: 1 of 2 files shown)


basepi:
Thank you for spotting and fixing this!

PR #1429, opened 2021-12-06
add span_compression_enabled setting

### LINKED ISSUE #1428: Include option of turning off compressed spans for python agent
We have endpoints that hit the database a lot, but with different queries. Previously before compressed spans went in we could see each of these queries in our trace breakdown. Now they are rolled up into a single "X Calls to postgres" which I can see being useful to some to avoid bloat in their span index but for us is near useless if we cant get the underlying queries being ran. Conversely if we had the option to add another layer to this logic 

```
Two spans are considered exact matches if the following attributes are identical:
 * span name
 * span type
 * span subtype
 * destination resource (e.g. the Database name)
```

for inclusion of the query being ran (ideally before parameters are applied to the query ie `SELECT name FROM person WHERE age > 18` and `SELECT name FROM person WHERE age > 21` could be considered the same would be awesome).  

Either a flag to not compress spans (if it already exists I couldn't find it in documentation), or additional logic to include thequery as part of the attribute consideration. 

None I suppose, tried to figure out what was happening and had to dig through release notes to find the addition to the python agent. 

**Additional context**
It would be nice if new features like this came out with a feature flag that defaulted to current behavior rather then having to dig around for why our spans now look differently.

beniwohli:
closes #1428


beniwohli:
test failures are because f68d35d4bee5 isn't on this branch (`6.x`) yet

PR #1443, opened 2022-01-05
ci: use the apm-its/main branch

v1v:

Leftover from the main renaming for the apm-its


beniwohli (approved):
LGTM, thanks for the quick reaction and fix @v1v !

PR #1444, opened 2022-01-10
Fix span compression for redis et. al.

beniwohli:

@LucaWintergerst pointed out that span compression doesn't work with redis. As it turns out, there were two bugs that caused this:

 1. after we implemented auto-calculation of span destination resource fields, these were no longer available in span compression. Moving the auto-calculation to run before span destination fixed this (e2c9c8f)
 2. `redis` spans, along with a bunch of other services, were not marked as leaf spans

To ensure that we don't regress on this again, I added a bunch of service-specific span compression tests.


basepi (approved):
Excellent! Redis is definitely a big target of span compression, so I'm glad this was caught fairly quickly!

PR #1453, opened 2022-02-01
jjbb: remove periodic trigger

v1v:

Remove the periodic trigger as we don't really need this based on the webhook push based events. 
This will reduce the number of builds which have been stalled for quite some time.

In addition, this will help us to update the job definition and pick up the `main` branch which it's not available at the moment.

### DIFF
--- a/.ci/jobs/apm-agent-python-linting-mbp.yml
+++ b/.ci/jobs/apm-agent-python-linting-mbp.yml
@@ -41,5 +41,4 @@
         timeout: '15'
         use-author: true
         wipe-workspace: 'True'
-    periodic-folder-trigger: 1d
     prune-dead-branches: true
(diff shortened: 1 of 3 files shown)


beniwohli:
@v1v will we still have the nightly build with the full matrix? We thinned out the matrix that is used for PRs quite substantially to speed them up, with the expectation that the full matrix is still run on a nightly basis on the main branch.

v1v:
> @v1v will we still have the nightly build with the full matrix? We thinned out the matrix that is used for PRs quite substantially to speed them up, with the expectation that the full matrix is still run on a nightly basis on the main branch.

this is nothing about the nightly jobs. It's regarding the internals for the Multibranch Pipelines and their reindexing policy. There are options to configure the multibranch pipelines to get triggered automatically when:

- the target branch has changed then all the dependent PRs without a rebase/merge will be built.
- polling every X minutes/hours

This change is only about this. 

Nightly jobs will run as usual, their configuration is defined in a different place -> https://github.com/elastic/apm-agent-python/blob/d8788b2e99a8e590391e976a05017721d0f1b974/.ci/nightly.groovy#L35-L37

beniwohli:
Ah gotcha, thanks for the explanation! Carry on, then :)

PR #1462, opened 2022-02-13
fix aioredis span context handling

qeternity:

Checks span context before attempting mutate values.

A bit surprised this made it into a release...completely breaks code outside of instrumented paths.

### DIFF
--- a/elasticapm/instrumentation/packages/asyncio/aioredis.py
+++ b/elasticapm/instrumentation/packages/asyncio/aioredis.py
@@ -54,7 +54,8 @@ class RedisConnectionPoolInstrumentation(AbstractInstrumentedModule):
         with async_capture_span(
             wrapped_name, span_type="db", span_subtype="redis", span_action="query", leaf=True
         ) as span:
-            span.context["destination"] = _get_destination_info(instance)
+            if span.context:
+                span.context["destination"] = _get_destination_info(instance)
 
             return wrapped(*args, **kwargs)
 
@@ -70,7 +71,8 @@ class RedisPipelineInstrumentation(AbstractInstrumentedModule):
         with async_capture_span(
             wrapped_name, span_type="db", span_subtype="redis", span_action="query", leaf=True
         ) as span:
-            span.context["destination"] = _get_destination_info(instance)
+            if span.context:
+                span.context["destination"] = _get_destination_info(instance)
 
             return wrapped(*args, **kwargs)
 


beniwohli:
Hi @qeternity, good catch! I took the liberty to open a PR against your branch with a test, it's here: https://github.com/zumalabs/apm-agent-python/pull/1

elasticapm/instrumentation/packages/asyncio/aioredis.py
```diff
@@ -54,7 +54,8 @@ def call(self, module, method, wrapped, instance, args, kwargs):
         with async_capture_span(
             wrapped_name, span_type="db", span_subtype="redis", span_action="query", leaf=True
         ) as span:
-            span.context["destination"] = _get_destination_info(instance)
+            if span.context:
+                span.context["destination"] = _get_destination_info(instance)
```
basepi, line 58:
```suggestion
            if span.context:
                span.context["destination"] = _get_destination_info(instance)
            else:
                span.context = {"destination": _get_destination_info(instance)}
```

elasticapm/instrumentation/packages/asyncio/aioredis.py
```diff
@@ -70,7 +71,8 @@ def call(self, module, method, wrapped, instance, args, kwargs):
...
```
basepi, line 75:
```suggestion
            if span.context:
                span.context["destination"] = _get_destination_info(instance)
            else:
                span.context = {"destination": _get_destination_info(instance)}
```

basepi (changes requested):
I think we want to set `span.context` if it doesn't exist.

qeternity:
Hi @beniwohli @basepi - sure thing, just needed to get a working branch to triage some internal deps issues. Merging both of your suggestions.

beniwohli:
@qeternity @basepi actually, I don't think we should initialize `context` if it is `None`. It is only `None` for `DroppedSpan` instances (`Span` instances have `context` initialized to an empty `dict`, see https://github.com/elastic/apm-agent-python/blob/ec9a733d6632e6a2aacafe2008a019724f062aa2/elasticapm/traces.py#L492). For `DroppedSpan` instances, we discard the whole thing in the end, so we'd do unnecessary work here by initializing it.

qeternity:
@beniwohli this was my understanding as well from my initial read of the repo - I'll revert the last commit

basepi:
Yep, my bad! Should have investigated more closely.

basepi (approved):

PR #1492, opened 2022-03-21
Reduce stack_trace_limit default from 500 to 50

### LINKED ISSUE #1128: Reduce default stack_trace_limit from 500 to 50
Looking through the [config docs](https://docs.google.com/spreadsheets/d/1JJjZotapacA3FkHc2sv_0wiChILi3uKnkwLTjtBmxwU/edit#gid=0&range=A63), all other agents set the default limit to 50.

Setting a modest limit for the stack traces is important for APM Server to avoid out of memory issues when the in-memory queue fills up. The APM Server's queue size is measured in the number of events, not bytes. Therefore, it's important to be able to estimate the maximum event size in bytes so that we can provision an appropriate amount of memory for the default queue size.

cc @simitt

basepi:
## Related issues
Closes #1128


beniwohli (approved):
LGTM

PR #1489, opened 2022-03-15
Lambda improvements

basepi:
## What does this pull request do?

Brings the lambda integration in line with the latest changes to the [spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md)

## Related issues
Closes #1486
Closes #1446 
Refs https://github.com/elastic/apm/pull/614
Closes #1480 
Closes #1424

PR #1493, opened 2022-03-22
Fix bug in lambda support with service_name overrides

basepi:

The JIT metadata overrides in our AWS lambda support had the side effect of not allowing for a custom-configured `service_name`. This fixes that issue and adds a test.

## Related issues
Closes #1489


beniwohli (approved):
Makes sense!

PR #1500, opened 2022-03-26
Celery contrib: read custom parent span id from the task headers

namoshizun:

In a complex Celery workflow, tasks might be invoked in a chain like this:

```
<A> ---> <B> ---> <C> ... --> <Z>
```

In such setting, because the current Celery instrument only propagates the first caller's traceparent, all downstream tasks share the same traceparent from A. This might be undesirable if every task should be regarded as a subtask of its caller.

For instance, "**submit_task_for_computation**" initiates "**do_heavy_lifting**" and waits for completion so that it can immediately rollback a database transaction if the job failed. But instead of correlate **do_heavy_lifting** with **submit_task_for_computation**, current instrument produces the following trace:

messy-spans

Whereas we actually hope for this: 

nicely-aligned

This PR is to allow a caller task to set the receiver task's parent span by passing a `parent_span_id` in the Celery headers.

I have prepared this [snippet](https://gist.github.com/namoshizun/27c8bef46ffac2e6015c3e23d1cd0d99) as a demo.


elasticapm/contrib/celery/__init__.py
```diff
@@ -68,24 +68,30 @@ def set_celery_headers(headers=None, **kwargs):
...
+        headers.update(elasticapm=apm_headers)
 
 
 def get_trace_parent(celery_task):
     """
     Return a trace parent contained in the request headers of a Celery Task object or None
     """
-    trace_parent = None
-    with suppress(AttributeError, KeyError, TypeError):
-        if celery_task.request.headers is not None:
```
namoshizun, line 82: This is unsafe because other tracing tools may also set the inner headers to add their tracing ids, whereas we just use the task headers. When using both elasticapm and sentry, elastic's traceparent propagation breaks.

basepi (approved):
Looks great! We're fixing a CI failure caused by upstream changes in another library, and then we should be able to get the tests run here and merge this.

Can you add a note to the CHANGELOG?

PR #1513, opened 2022-03-30
Add doc redirects

bmorelli25:

- [x] Reverts #1512. A good idea, but because the file isn't included in the repo's `index.asciidoc` file, it isn't actually built.
- [x] Adds a hidden redirect page that includes the old `opentracing-bridge` id and sends users to the new `opentelemetry-bridge`.

* This will need to be backported to 6.x.
* After this PR is merged into 6.x, links from the APM Guide can be updated to point to the new `opentelemetry-bridge` id.


basepi (approved):

basepi:
I reset `6.x` to `main` so this fix is there too.

PR #1516, opened 2022-04-04
ci: packer_cache never fail

v1v:

Avoid failing the packer_cache process.

## Why

https://github.com/elastic/apm-agent-python/issues/1505 broke the script, but the packer_cache should work on best-effort, so if possible then cache the images, otherwise do not fail.

Then the packer cache image generation can move on with the next packer_cache project. Otherwise, we won't be using latest cache images


cachedout (approved):

v1v:
Unrelated test failures, this particular change is not tested in the CI directly but in another process. We are safe to merge it

PR #1517, opened 2022-04-04
Update release process to not require admin access

basepi:

Currently our release process requires multiple no-check pushed to protected branches. As we (eventually) would like to stop using admin overrides on the repo, our release process needs an update.

This update will require both @beniwohli and I to be around at the same time for releases, as we'll need to approve the PRs. We'll also have to wait for tests to run on the release PRs. I don't think that's a terrible requirement, but it will hurt our ability to do fast regression releases.

Thoughts welcome.

cc @felixbarny @estolfo @AlexanderWert

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -5,4 +5,4 @@ Here you can explain the changes made on the PR.
 -->
 
 ## Related issues
-closes #ISSUE
+Closes #ISSUE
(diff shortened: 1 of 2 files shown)


basepi:
On further thought, maybe we can put the release tag on `main`, because then it will show up on `6.x` after we merge from `main` to `6.x`? That's probably an improvement, I'll make that change.

felixbarny:
Another route could be automating the steps that require admin permissions and allow the CI user to bypass the PR requirement. The downside is that if someone gets ahold of the CI user, they can do anything.

@Mpdreamz What's your stance on exempting CI users from certain branch protection rules?

> This update will require both @beniwohli and I to be around at the same time for releases, as we'll need to approve the PRs. 

Leads can also help out with the reviews if someone is out.

basepi:
Closing this for now. We can re-address this when we next do more SLSA.

PR #1518, opened 2022-04-05
Fix sanic tests

### LINKED ISSUE #1515: Sanic 22.3.0 has CI failures
The newly released Sanic 22.3.0 changed something in URL routing that seems to affect our ability to get the route. See e.g.

https://apm-ci.elastic.co/job/apm-agent-python/job/apm-agent-python-nightly-mbp/job/main/5/testReport/junit/tests.contrib.sanic/sanic_tests/Test___Python_python_3_10_0___test_get__greet_sanic_GET__greet__name_string__0_custom_context1_/

I couldn't quite figure out what the problem is yet, but a quick debugging session showed that `request.route` is set to None, and that the  `Router.get()` method changed calling arguments, used here: https://github.com/elastic/apm-agent-python/blob/main/elasticapm/contrib/sanic/__init__.py#L228

basepi:

See https://github.com/sanic-org/sanic/issues/2423 for some context.

Our Sanic tests were failing due to not being able to get the route from one of the requests.

It took me awhile to track down the issue, but it turns out the router stopped accepting `<name:string>` and wants `<name:str>` instead. This was not listed as a breaking change in their release notes, so I recommended they add it.

## Related issues
Closes #1515

### DIFF
--- a/tests/contrib/sanic/fixtures.py
+++ b/tests/contrib/sanic/fixtures.py
@@ -125,7 +125,7 @@ def sanic_elastic_app(elasticapm_client):
                 pass
             return json({"response": "ok"})
 
-        @app.get("/greet/<name:string>")
+        @app.get("/greet/<name:str>")
         async def greet_person(request: Request, name: str):
             return json({"response": f"Hello {name}"})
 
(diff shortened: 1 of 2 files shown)


beniwohli (approved):
Great work on tracking this down!

PR #1520, opened 2022-04-05
Add instrumentation for aiobotocore

stj:

This adds instrumentation for aiobotocore. It heavily uses the existing instrumentation from boto3.

I got the tests to run locally for this, but noticed that the instrumentation of `aiohttp.client` already sends spans for `aiobotocore`. Though they are not labelled as AWS specific as the `boto3` instrumenter does.

As is, this implementation currently results in 2 spans for each request. I am looking for some advice on what is the better approach here. Try to remove the wrapper on the `ClientSession` used by `aiobotocore` or change the spans created by `ClientSession`?

## Related issues
Nothing opened (yet)

### DIFF
--- a/.ci/.jenkins_exclude.yml
+++ b/.ci/.jenkins_exclude.yml
@@ -237,3 +237,8 @@ exclude:
     FRAMEWORK: aiomysql-newest
   - PYTHON_VERSION: python-3.10  # getting "loop argument must agree with lock" error
     FRAMEWORK: aiomysql-newest
+  # aiobotocore
+  - PYTHON_VERSION: pypy-3
+    FRAMEWORK: aiobotocore-newest
+  - PYTHON_VERSION: python-3.6
+    FRAMEWORK: aiobotocore-newest
--- a/.ci/.jenkins_framework.yml
+++ b/.ci/.jenkins_framework.yml
@@ -52,3 +52,4 @@ FRAMEWORK:
   - prometheus_client-newest
   - sanic-newest
   - aiomysql-newest
+  - aiobotocore-newest
--- a/.ci/.jenkins_framework_full.yml
+++ b/.ci/.jenkins_framework_full.yml
@@ -84,3 +84,4 @@ FRAMEWORK:
   - httplib2-newest
   - sanic-20.12
   - sanic-newest
+  - aiobotocore-newest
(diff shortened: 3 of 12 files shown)


basepi:
@stj A few tests failing. It appears that with `aiobotocore` it's not correctly detecting the service. The span type/subtype are set in the `handle_*` functions, which are chosen dynamically when we detect the service:

https://github.com/elastic/apm-agent-python/blob/114ee6ca998b4d6a5cb075a289af39cb963cf08a/elasticapm/instrumentation/packages/botocore.py#L61-L82

Can you take a look and see if you can figure out where we're going wrong in the service detection for `aiobotocore`?

stj:
> This is awesome! Very clean way to re-use the botocore instrumentation. +1

Thanks, the way the library instruments sync and async made this a fairly straight forward change!

Though, as I mentioned, this is currently creating two spans for each request to AWS (one created by the instrumentation of `aiohttp.client` and another by this instrumentation). I am not sure what is the cleanest way to handle this and though I put the PR up to discuss this.

My current thinking is to find out if the wrapper for `aiohttp.client` can be removed for the `ClientSession` instance used by `aiobotocore`. WDYT?

basepi:
Oh, right, sorry, I missed your discussion note.

That also probably explains the test failures, we're probably inspecting the http span (which is `external`) instead of the s3 span (which is `storage`).

The correct solution is to mark the spans created by this instrumentation as `leaf=True`, which means that sub-spans will be dropped. This is how we handle spans where we don't want the underlying http span (such as spans from the elasticsearch client, for example).

But...you already are doing that, so there might be a bug.

stj:
> The correct solution is to mark the spans created by this instrumentation as leaf=True, which means that sub-spans will be dropped. This is how we handle spans where we don't want the underlying http span (such as spans from the elasticsearch client, for example).

TIL

stj:
Turns out that not creating an `async def call` method on the async instrumentor closes the span to early. Basically the sync `call` method, as I had it, returned a coro. The return exits the span context manager. When the coro is awaited the span for `aiohttp.client` does not find a leaf span in the execution context and continues as normal.

https://github.com/elastic/apm-agent-python/pull/1520/commits/4e778c92d43f07077894235000d010a93ea11526 makes changes to `botocore` and `aiobotocore` instrumentation classes to avoid code duplication as much as possible.

PR #1535, opened 2022-04-22
ci: use coverageReport for code coverage

v1v:

Use the new step for the code coverage

## Related issues

As soon as https://github.com/elastic/apm-pipeline-library/pull/1673 is merged and released then the code coverage report on a PR basis will be available

## Test

The below tests are the ones I used for supporting the GitHub comments with the Code Coverage delta between the PR and the target branch

If it cannot be compared with the target branch:

image

If it can be compared with the target branch:

image

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 #!/usr/bin/env groovy
-@Library('apm@test/support-coverage-file') _
+@Library('apm@current') _
 
 import co.elastic.matrix.*
 import groovy.transform.Field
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -189,6 +189,7 @@ pipeline {
                 changeRequest()
                 expression { return !params.Run_As_Main_Branch }
               }
+              expression { return false }
```
v1v, line 192:
```suggestion
```

kuisathaverat (approved):

PR #1536, opened 2022-04-22
Conditionally log warning when adding labels

ckoehn:

This change conditionally logs a warning, based on the state of the APM integration (`ELASTIC_APM_ENABLED`), if no transaction is present and adding labels. It still warns developers when data is discarded and at the same time doesn't spam the log with unnecessary log lines in case the integration is disabled.

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -1029,7 +1029,8 @@ def label(**labels):
     """
     transaction = execution_context.get_transaction()
     if not transaction:
-        error_logger.warning("Ignored labels %s. No transaction currently active.", ", ".join(labels.keys()))
+        if elasticapm.get_client().config.enabled:
+            error_logger.warning("Ignored labels %s. No transaction currently active.", ", ".join(labels.keys()))
     else:
         transaction.label(**labels)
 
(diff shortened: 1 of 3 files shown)


elasticapm/traces.py
```diff
@@ -1029,9 +1028,8 @@ def label(**labels):
     """
     transaction = execution_context.get_transaction()
     if not transaction:
-        error_logger.warning("Ignored labels %s. No transaction currently active.", ", ".join(labels.keys()))
-    else:
-        transaction.label(**labels)
+        return
+    transaction.label(**labels)
```
basepi, line 1032: I think removing the error log completely is probably not the right move. Instead, can we check if the agent is enabled, and if so, then we can log the error? I think warnings are useful when we're throwing away data unexpectedly -- but I agree that it's not unexpected if the agent is `enabled=False`.

You can get the config from `transaction._tracer.config.enabled`.

basepi (changes requested):

elasticapm/traces.py
ckoehn: Fair point. I'm gonna change it accordingly.

ckoehn: In case of a disabled agent there is no transaction and thus your suggestion doesn't work. Do you have any other recommendation to retrieve the config? Maybe via `get_client()`?

basepi (approved):
Looks great! 👍

PR #1538, opened 2022-04-25
Profiler poc update

dhx:
Merged with upstream - resolved obvious conflicts and fixed bug induced by the zprofile import ordering - works with apm 8.1.3 afaict

@beniwohli do you see a chance to land this upstream while the visualization in kibana still is an internal experimental feature?

### DIFF
--- a/elasticapm/base.py
+++ b/elasticapm/base.py
@@ -50,6 +50,7 @@ import elasticapm
 from elasticapm.conf import Config, VersionedConfig, constants
 from elasticapm.conf.constants import ERROR
 from elasticapm.metrics.base_metrics import MetricsRegistry
+from elasticapm.profiler import Profiler
 from elasticapm.traces import Tracer, execution_context
 from elasticapm.utils import cgroup, cloud, compat, is_master_process, stacks, varmap
 from elasticapm.utils.encoding import enforce_label_format, keyword_field, shorten, transform
@@ -212,6 +213,8 @@ class Client(object):
         if self.config.use_elastic_excepthook:
             self.original_excepthook = sys.excepthook
             sys.excepthook = self._excepthook
+        if self.config.profiler:
+            self._thread_managers["profiler"] = Profiler(self)
         if config.enabled:
             self.start_threads()
 
--- a/elasticapm/conf/__init__.py
+++ b/elasticapm/conf/__init__.py
@@ -594,11 +594,10 @@ class Config(_ConfigBase):
         default=timedelta(seconds=30),
     )
     profiler = _BoolConfigValue("PROFILER", default=True)
-    profiler_interval = _ConfigValue(
+    profiler_interval = _DurationConfigValue(
         "PROFILER_INTERVAL",
-        type=int,
-        validators=[duration_validator, ExcludeRangeValidator(1, 999, "{range_start} - {range_end} ms")],
-        default=20000,
+        validators=[ExcludeRangeValidator(1, 999, "{range_start} - {range_end} ms")],
+        default=timedelta(seconds=20),
     )
     breakdown_metrics = _BoolConfigValue("BREAKDOWN_METRICS", default=True)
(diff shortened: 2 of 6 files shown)


beniwohli:
Hey @dhx! Thanks for the PR. Unfortunately, we will not move further on this path.

The good news however is that we recently joined forces with Optimyze and will leverage Prodfiler to profile applications. This is a much more mature solution. You can read more about it in this blog post:

https://www.elastic.co/blog/elastic-and-optimyze-join-forces-to-deliver-always-on-continuous-profiling-of-infrastructure-applications-and-services

qeternity:
This is incredibly disappointing, and clearly a decision driven by corporate. Prodfiler is an absolutely terribly product (sorry Optimyze team). I'm actually stunned that this is the replacement. After using Elastic for years, we'll probably switch to Datadog given this has been nuked from the roadmap.

SeanHeelan:
@qeternity Hey Chase! One of the founders of Optimyze here. Do you have any specific feedback on what you disliked about Prodfiler?

SeanHeelan:
@qeternity I had a quick look at your project on the Prodfiler platform and I am _guessing_ the issue you encountered is that you did not see proper symbols in the stack traces for your custom application. In order to assign function names to stack traces from custom binaries, Prodfiler needs you to upload debug symbols to our platform. This can be done after the profiling has been done, and you do *not* need to ship these symbols to your own production environment. 

Details on how to do so can be found here https://documentation.prodfiler.com/faq.html#why-are-my-stack-traces-not-symbolised

Apologies that our documentation is a bit lacking. We should really have a tooltip in the UI that highlights that this is the problem.

qeternity:
Hi Sean - let me caveat by saying that it's terrible _for our purpose_, and to be blunt, I think not well suited to interpreted languages. I'm sure it's great for compiled binaries. We're running Pyston, which is a fantastic improvement over even recently optimized bleeding edge CPython versions. There is already a fairly rich profiler in Python which Pyston supports, and so it's frustrating to have to do and end-run around this support by trying to upload debug symbols.

But the biggest issue is the documentation: if you make claims about supporting languages, but that support only extends to upstream packages, and then there is no good documentation on how to resolve, as I user I simply cannot use the product at that point. Additionally, the added step of shipping debug symbols because we might not be using exact upstream packages is an awkward burden when there is a rich built-in profiler that Elastic have simple decided not to support in favor of their acquisition. I also don't want to have to use another product when I have Kibana well integrated into our ops.

I can guess why this decision was made, but for this specific Elastic-APM-Python context it doesn't work out well for the end user.

SeanHeelan:
Yea, you are correct, we don't support Pyston at the moment (you're the first to request it, and I'll add it to our R&D list). Uploading symbols is unlikely to solve this problem for you. The issue is that we will have to add a new unwinder for however Pyston itself represents stacks internally. It appears to have a custom JIT, so we'd need to look into that, but it's likely a solvable problem (we recently added support for PHP's JIT, for example). 

You're correct on the docs though. They were intended to support an initial beta period, and our plan post-beta (last August) was to continue to iterate on them. After the acquisition our focus has been on landing the integration on top of Elasticsearch itself, as opposed to the prodfiler.com SaaS. As a result, the docs have been neglected. 

On the upside, that means that once the integration is complete you won't have to use another application. Prodfiler will live inside Kibana itself, as a normal part of the Observability solution.

qeternity:
Thanks for the info. I think everyone sympathizes with neglecting docs, but yeah just to frame my rather blunt comment above, hopefully it gives some insight into someone who would actually jump at being your customer if we could (the UI looks pretty solid, more so than Kibana).

Well if you want to chuck this thread in your issue tracker, if you guys get around to supporting Pyston, we'll happily beta.

SeanHeelan:
@qeternity Yea, that's super helpful, thanks! We'll add Pyston to our R&D list and I'll reach out once we get a chance to build an unwinder for it.

PR #1549, opened 2022-05-16
Better "capture_body" flag support

### LINKED ISSUE #1545: Starlette middleware not respecting "capture_body" flag
Hello :)

While investigating a bug regarding **very slow** uploads, we discovered two things about the Starlette middleware.

- It does not honor the "capture_body" flag by consuming the stream regardless the flag is turned on or not. It replaces the original stream with a mocked one, making the Starlette request.stream() feature useless because the request is already entirely in memory (using a lot of memory in case of file uploads). I think the following piece of code could check the flag and behave differently.https://github.com/elastic/apm-agent-python/blob/69beac4a5fc72249b40fc811a9ef34fd33313735/elasticapm/contrib/starlette/__init__.py#L151-L163
- The same block of code above uses byte concatenation instead of a byte array which is fine for very small buffers but **prohibitively slow** with anything bigger because of the quadratic behavior. Gigabyte-sized streams do not work at all.

I would happily send a pull request for that, but I would like some opinions first, especially after seeing this (https://github.com/elastic/apm-agent-python/pull/1528), which seems promising.

Anyway, Elastic APM is awesome. Thanks for the great work!

frnkvieira:
- Fixes "capture_body" flag not being respected causing the request.stream() of Starlette not being able to properly stream (by loading everything into memory first)
- Fixes non-UTF-8 data triggering errors in requests that should be ignored as configured.
- Improves performance of all requests (especially those who send big payloads like files) by not mocking the stream at all in case the capture_body flag is turned off (which is the default). When the flag is turned on requests are also faster because of the faster loop by using a list of bytes instead of bytes concatenation.

I'm still wrapping my head around how this agent works so my apologies in advance if I made any wrong assumptions =)

## Related issues
closes #1545


frnkvieira:
Hi @beniwohli, to be clear, the intent of this pull request is closing #1545, but as a side effect, this fixes non-UTF-8 errors in **ignored paths**, only ignored ones, thanks to the ignore path checkup being done way earlier (which seems a good choice to me but maybe I'm missing something). (https://github.com/frnkvieira/apm-agent-python/blob/main/elasticapm/contrib/starlette/__init__.py#L134)

Btw, I don't think crashing on non-utf-8 data is a good thing, but I didn't feel confident enough to keep changing stuff =P
Ignore the errors may be a solution, although it would probably create some strange corner cases for people processing the APM data... encode it in base64 or similar may help preserve the original content but would create other problems... which brings me to other questions, maybe add a size limit where the content gets truncated ?... Anyway... thanks for the fast reply =)

julianogv:
Hi @beniwohli, any idea on when you might merge this PR and publish it to pypi? 
I'm currently using a forked version of elastic-apm due to this issue (and some other fixes that are already fixed/merged and not published yet).

basepi:
@frnkvieira looks like you've turned off "allow maintainer edits" so I opened a PR instead: https://github.com/frnkvieira/apm-agent-python/pull/1

@julianogv apologies for the delay here!

frnkvieira:
I'm probably missing something because the flag seems to be checked.
Anyway, I just merged the PR on the fork, thanks.

basepi (approved):

PR #1571, opened 2022-06-09
Fix wrong transaction names in Django 4.0+

### LINKED ISSUE #1570: Wrong transaction name in Django 4.0+
**Describe the bug**:
When using Django versions 4.0+, the transaction names are incorrect.

Expected: `django_app.views.ListUsers`
Actual: `django_app.views.view`

- OS: Debian Bullseye
- Python version: 3.9.13
- Framework and version: Django 4.0.5
- APM Server version: 7.12.0
- Agent version: 6.9.1

**Additional context**

This issue seems to be introduced with the following [PR](https://github.com/django/django/pull/14124) to Django.
https://github.com/django/django/blob/e96320c91724830034033a9cb8afd9cf8c11e2fd/django/views/generic/base.py#L108-#L110

I will try to provide a fix shortly.

dhalenok:

- Tweaks `get_name_from_func` to use `view_class` attribute when resolving view name.

## Related issues
closes #1570

### DIFF
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -83,7 +83,9 @@ def get_name_from_func(func: FunctionType) -> str:
 
     module = func.__module__
 
-    if hasattr(func, "__name__"):
+    if hasattr(func, "view_class"):
+        view_name = func.view_class.__name__
+    elif hasattr(func, "__name__"):
         view_name = func.__name__
     else:  # Fall back if there's no __name__
         view_name = func.__class__.__name__


beniwohli:
@dhalenok awesome, thanks for opening the issue and this PR! The fix looks good. I was wondering why this wasn't caught by our tests, and it turns out we don't have any class based views in out [test app](https://github.com/elastic/apm-agent-python/blob/main/tests/contrib/django/testapp/views.py). If you like, you could add one, and then test that we get the correct name by copying and modifying this test: https://github.com/elastic/apm-agent-python/blob/986c86820b6e2ec7a9f05941bbf376706f6d5d26/tests/contrib/django/django_tests.py#L732-L747

beniwohli (approved):
Nice fix and test, thanks!

PR #1579, opened 2022-06-29
fix issues when transaction is unsampled in Kafka instrumentation

### LINKED ISSUE #1578: AttributeError with kafka-python integration in 6.10.0
**Describe the bug**:

After upgrading the agent to 6.10.0, we immediately started seeing AttributeError failues in the ElasticAPM code when calling through to our kafka-python code paths (we've had to rollback to 6.9 because of this).

**To Reproduce**

Try to send a message using the KafkaProducer class in kafka-python.

Traceback:

```python
AttributeError: 'NoneType' object has no attribute 'id'
[...]
  File "elasticapm/instrumentation/packages/kafka.py", line 94, in call_if_sampling
    return self.call(module, method, wrapped, instance, args, kwargs)
  File "elasticapm/instrumentation/packages/kafka.py", line 106, in call
    return self._trace_send(instance, wrapped, destination_info=destination_info, *args, **kwargs)
  File "elasticapm/instrumentation/packages/kafka.py", line 71, in _trace_send
    tp = transaction.trace_parent.copy_from(span_id=span.id)
```

I have had to remove proprietary code from the stack trace. However, the two snipped lines of code before it hits the elasticapm code paths is basically just doing:

```python
from kafka import KafkaProducer
producer = KafkaProducer(**kafka_config)  # where kafka_config is a dict of config options.
producer.send(topic, message).get()
```

- OS: Linux
- Python version: 3.9.6-slim-bullseye (Docker container)
- Framework and version [e.g. Django 2.1]: Django 3.2.13, kafka-python 2.0.2
- Agent version: 6.10.0

beniwohli:
closes #1578

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -32,6 +32,14 @@ endif::[]
 [[release-notes-6.x]]
 === Python Agent version 6.x
 
+[[release-notes-6.10.1]]
+==== 6.10.1 - 2022-06-30
+
+[float]
+===== Bug fixes
+
+* Fix an issue with Kafka instrumentation and unsampled transactions {pull}1579[#1579]
+
 [[release-notes-6.10.0]]
 ==== 6.10.0 - 2022-06-22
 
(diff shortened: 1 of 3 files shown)


basepi:
This needs a CHANGELOG entry as well

PR #1585, opened 2022-07-11
Database granularity

### LINKED ISSUE #1497: [META 621] Improve Granularity for SQL Databases
[issue details](https://github.com/elastic/apm/issues/621)

[issue details](https://github.com/elastic/apm/issues/622)

There are two parts to improving the granularity for SQL Databases:
- adding new fields to agent-server protocol, for now it's mostly to future-proof the captured data as we have to keep existing fields for compatibility with UI.
- capture values for those new fields on SQL Databases, which is actually providing better granularity to the end-user.

Each step can be completed independently, but this task will be considered complete when both parts are completed.

**First step : add new service fields**

  - Add and capture values for `span.context.service.target.type` and `span.context.service.target.name` for exit spans.
  - Infer from those new fields the value of `span.destination.service.resource` and keep sending it.
  - Add `service_target_*` fields to dropped spans metrics:  add `service_target_type` and `service_target_name` next to `destination_service_resource` in `transaction.dropped_spans_stats` array, the related metrics documents should include `span.service.target.type` and `span.service.target.name` fields.
  - Handle span compression with new fields (stop relying on `resource` internally)

**Second step : capture values for new fields**

beniwohli:
## Related issues
closes #1497


basepi (approved):
So much busywork! But the increased granularity in data will be good.

Looks great!

PR #1591, opened 2022-07-21
cast redis ports to int

qeternity:

It appears possible in at least come versions of the python redis libraries (redis-py and aioredis) to end up with ports that are stored as strings, whereas apm-server is expecting an int, which results in parsing issues. This preemptively casts all redis ports to ints before building the destination info.


basepi (approved):
Looks good to me! Thanks for adding this.

PR #1593, opened 2022-07-21
Fix varmap to apply func before recursion

### LINKED ISSUE #1592: Performance Bug in the varmap function
**Describe the bug**: ...
When python code throws an error, and one of the local variables contains a very large list of lists of lists, it takes the elastic python apm client a very long time to process through the list. In our case it was taking over 60 seconds for the code to complete.

**To Reproduce**
To simply replicate the issue, I: (see code snipit  in the elastic forum, link below)

- created a simple script to create a list of list of lists of integers.
- in one method I had my "with elasticAPMTransaction:" statement.
- under this I created the list (200 integers) of list (300 integers) of list (40 integers) aka 24M items total.
- started to iterate through the outer list and picked an arbitrary spot to through an exception. This essentially mimicked the method where the real code threw an exception.
- wrapped this whole thing in a profiler and evaluated the output.
This would take 60 seconds +/- to complete.

Seems like the issue is in the varmap function in elasticapm\utils_init_.py but there maybe lots of ways to solve this issue.

- OS: [e.g. Linux]
- Python version: 3.x
- APM Server version: N/A
- Agent version: 6.7.2

**Additional context**
Created a case here https://discuss.elastic.co/t/python-client-apm-very-slow-when-local-list-variable-is-very-large/309935. See this for additional details.

basepi:

We use `varmap()` to apply [`shorten()`](https://github.com/elastic/apm-agent-python/blob/02ed26d2b41be2cd8ba7d08f3513cddfcd06cbfe/elasticapm/utils/encoding.py#L179) recursively across a data structure. `varmap()` was recursing into the data structures, and _then_ applying the `func()` (in this case, `shorten()`) on the way back out of recursion. This meant that when shortening local variables where just recursing over the data structure was a performance issue, setting `ELASTIC_APM_LOCAL_VAR_LIST_MAX_LENGTH` wouldn't actually affect the runtime, just the output.

With this change, dictionaries and lists will be shortened and _then_ we'll iterate over their contents. Shouldn't affect `_sanitize()` at all, but will speed up `shorten()` considerably.

I think the test surface on this function is sufficient without adding tests for this fix.

## Related issues
Closes #1592

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,19 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 6.x" heading
+//[float]
+//===== Features
+//
+[float]
+===== Bug fixes
+
+* Fixed a performance issue for local variable shortening via `varmap()` {pull}1593[#1593]
+
+
 [[release-notes-6.x]]
 === Python Agent version 6.x
 
--- a/elasticapm/utils/__init__.py
+++ b/elasticapm/utils/__init__.py
@@ -64,10 +64,14 @@ def varmap(func, var, context=None, name=None, **kwargs):
         return func(name, "<...>", **kwargs)
     context.add(objid)
     if isinstance(var, dict):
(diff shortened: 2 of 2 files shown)


beniwohli (approved):
Great catch! LGTM

PR #1594, opened 2022-07-22
Simplify logic in varmap

basepi:

I merged #1593 before following up on #1592 -- @kohlmanj-hs had an excellent recommendation there to simplify my logic from #1593. Thus, this PR.


beniwohli (approved):
LGTM!

PR #1596, opened 2022-07-25
label no longer fails when called without a client

### LINKED ISSUE #1595: elasticapm.label fails when trying to warn that no transaction is present
**Describe the bug**:
https://github.com/elastic/apm-agent-python/blob/main/elasticapm/traces.py#L1092

When neither Transaction nor client is present, `elasticapm.label( . . .)` will fail while trying do print a warning.
This can happen for example when testing routes with fastapi.

Open Question would be whether the warning should be printed if no client is present.
I think it makes sense to do this since this is also the default from the config.

**To Reproduce**

```
import elasticapm

elasticapm.label(foo="foo")
```

- OS: Tested on Linux and Mac but should appear everywhere else
- Python version: Tested on 3.9/3.10
- Agent version: 6.10.1

dennis-wey:

`elasticapm.label(. . .)`no longer fails when no client is present

## Related issues
closes #1595

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -1089,7 +1089,8 @@ def label(**labels):
     """
     transaction = execution_context.get_transaction()
     if not transaction:
-        if elasticapm.get_client().config.enabled:
+        client = elasticapm.get_client()
+        if not client or client.config.enabled:
             error_logger.warning("Ignored labels %s. No transaction currently active.", ", ".join(labels.keys()))
     else:
         transaction.label(**labels)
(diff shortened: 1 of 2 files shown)


beniwohli (approved):
Great catch! Thanks @dennis-wey !

basepi (approved):

dennis-wey:
@basepi @beniwohli thanks for reviewing and merging this so quickly! Would it be possible to create a bugfix release with the current changes on master?

basepi:
Yes I'll create a bugfix release, probably tomorrow.

basepi:
Version 6.10.2 is released.

PR #1605, opened 2022-08-05
Instrument aws elb serverless requests

### LINKED ISSUE #1587: AWS lambda: capture_serverless does not provide information when lambda is connected to ELB (not to API GW)
Event that lambda handler receiving in case of attachment through the ELB you can find in attachment: [elb-lambda-event.txt](https://github.com/elastic/apm-agent-python/files/9103558/elb-lambda-event.txt)

Please note that queryStringParameters are quoted (in the event which is coming from API gateway they are not).
Note that in this case requestContext contains only elb element and nothing else.

Recently AWS has introduced that lambda could be joined directly to ELB (without API gateway) and when transaction is getting recorded by capture_serverless then it has much less information comparing to the API gateway connected one.

This function should be rewritten to react accordingly. For the lambdas which are ELB api invocations - the " elasticapm.set_context(lambda: get_data_from_request( is not getting invoked.

    def __enter__(self):

..... 

 self.httpmethod = nested_key(self.event, "requestContext", "httpMethod") or nested_key(
            self.event, "requestContext", "http", "method"
        )
....

if self.httpmethod:  # API Gateway
    self.source = "api"

.....

.....
       if self.source == "api":
            elasticapm.set_context(
                lambda: get_data_from_request(

tonyman19:
## This PR lets instrument elb serverless requests.
When lambda is being connected directly to ALB (this is useful when you want to rewrite some ECS existing endpoints with their lambda counterparts)

It fixes the bugs for full url creation for api gw ver 1 (when url is formed from queryStringParameters collection):
1. full url was not properly formed with parameters (& were missing between key=value elements)
2. url encoding was not applied to key / value pairs inside url

## Related issues
closes #1587


elasticapm/contrib/serverless/aws.py
```diff
@@ -246,6 +256,18 @@ def set_metadata_and_context(self, coldstart: bool) -> None:
...
             cloud_context["origin"]["account"] = {"id": self.event["requestContext"]["accountId"]}
             cloud_context["origin"]["provider"] = "aws"
+        elif self.source == "elb":
+            elb_target_group_arn = self.event["requestContext"]["elb"]["targetGroupArn"]
+            faas["trigger"]["type"] = "http"
+            faas["trigger"]["request_id"] = self.event["headers"]["x-amzn-trace-id"]
+            service_context["origin"] = {"name": elb_target_group_arn.split(":")[5]}
+            service_context["origin"]["id"] = elb_target_group_arn
+            service_context["origin"]["version"] = self.event.get("version", "1.0")
+            cloud_context["origin"] = {}
```
basepi, line 266:
```suggestion
            service_context["origin"]["id"] = elb_target_group_arn
            cloud_context["origin"] = {}
```

Let's just omit the origin.version field here. Since we're not using API Gateway I don't think there's anything useful we can put in that field for ELB.

basepi (changes requested):
This is really excellent, thank you! Just one minor change, and I'll also be opening a PR against the [spec](https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md) so we can make sure this is consistent across the serverless agents.

basepi (approved):
I added the fix I recommended + a CHANGELOG entry

basepi:
@tonyman19 Did you sign the CLA for the email `anton@mycarly.com`? That's the email attached to your commits.

tonyman19:
Hi Colton,

Thanks for your PR review and getting back to me.
I've signed it with the ***@***.*** (same as my github
account).
***@***.*** - this is my work email which I'm using for our elastic
subscription - but the contribution will be from ***@***.***

In the attachment you can find the signed document (tried 3 times before
and everytime missed the button to finish the signature - but was able to
download PDFs :))

Best regards
Anton Matviyenko

tonyman19:
Hi Colton,

Thank you for review and quick reply. Finalized signing for [anton.matviyenko@gmail.com](mailto:anton.matviyenko@gmail.com)
What about recorded commits - is it possible to change from [anton@mycarly.com](mailto:anton@mycarly.com) to the abovementioned email?
Or is it ok the way they are ATM?
Will sign the CLA for anton@mycarly.com as well.

tonyman19:
Short question about following inconsistency that I have found. Is it by purpose or just a mistake?

```
if self.source == "api":
 ....
   cloud_context["origin"]["account"] = {"id": self.event["requestContext"]["accountId"]}
elif self.source == "sns":
 .....
   cloud_context["origin"]["account_id"] = record["Sns"]["TopicArn"].split(":")[4]
```

So first time (and everywhere else) we are creating subobject with id field ["origin"]["account"]["id"] - next time we create subelement account_id which is located under ["origin"]["account_id"].

Should we also fix that here?

tonyman19:
Regarding your removal of:

```
service_context["origin"]["version"] = self.event.get("version", "1.0")
```

version field is also absent for rest (or http integrations) v1 (check aws_api_test_data.json)

I assumed that there might be a second version of elb integration where it will be present so I did it the same way. I did it conforming the spec for api gw: https://github.com/elastic/apm/blob/main/specs/agents/tracing-instrumentation-aws-lambda.md#api-gateway-v1--v2

context.service.origin.version | e.g. 1.0 | 1.0 for API Gateway V1, 2.0 for API Gateway V2. | event.version (or 1.0 if that field is not present)
-- | -- | -- | --

basepi:
I suppose there could be a v2 for ELB some day. We did it that way for API Gateway because there were two prominent (and quite different) versions in use when we launched the integration. I think since there's only one version of ELB right now, we'll leave off the version (like we do for the other non-http triggers), and we can add it in the future if it becomes relevant.

Thanks for signing the CLA with the correct email. I can't change the email on the commits, you would have to re-configure your git client and then rebase the commits. But everything is good now that you signed the CLA for the correct email.

Good catch on the SNS `account_id` -- that was definitely a bug! I have fixed it.

PR #1609, opened 2022-08-11
Add lambda url granularity to lambda integration

basepi:

Adds `lambda url` support to our aws serverless integration. Since lambda urls just use the API v2 payload, it worked fine before, but this just adds granularity in the cloud.origin.service.name field.

## Related issues
Spec: https://github.com/elastic/apm/pull/669

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,18 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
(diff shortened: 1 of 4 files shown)


trentm (approved):
LGTM

basepi:
> We (on the nodejs agent) get the same "Build Aborted" looks-like-tests-are-red-but-they-are-fine when adding the CHANGELOG entry after having a PR number.

Yeah I don't think there's an easy way to prevent the "error" unless you wait for the test run to complete before pushing CHANGELOG. Which would be a big waste of time. 🤷‍♂️

Issue #305, opened 2018-10-23
Limit the length for non-keyword fields


alvarolobato:
Currently only the keyword fields are limited at 1024 characters, we need to also apply this limit to non keyword fields. There will also be extended fields ie. SQL that can go up to 10000 characters to non-keyword fields.

basepi:
I think this is enforced for most things at this point.

* SQL queries are shortened in [the `dbapi2` code](https://github.com/elastic/apm-agent-python/blob/00802b598cb49dc5c5c151417ddd7413a08e6339/elasticapm/instrumentation/packages/dbapi2.py#L231).
* Local vars are shortened in [the client code](https://github.com/elastic/apm-agent-python/blob/00802b598cb49dc5c5c151417ddd7413a08e6339/elasticapm/base.py#L174)

Obviously any keyword fields are already shortened.

@beniwohli I think with the age of this issue I'm going to close it as resolved. If we do run into any more issues I suppose we could write a processor that would shorten everything -- but that's probably not a precise-enough instrument, since different things need to be shortened differently. Keep me posted if you think there's still work to do here.

basepi:
Just kidding. Found some more conversation in the cross-agent meta issue, I guess this is still in flux despite the issue-open date. Pardon my noise.

trentm:
https://github.com/elastic/apm/blob/main/specs/agents/field-limits.md#long_field_max_length-configuration was spec'd a while ago. That covers a specific set of fields. (The Node.js APM agent truncates other fields at 1024.)

basepi:
`LONG_FIELD_MAX_LENGTH` introduced in #1610 and added to Django's capture_body. First of many places.

basepi:
Length of exception message field limited in https://github.com/elastic/apm-agent-python/pull/1619

basepi:
Length of Starlette bodies limited in https://github.com/elastic/apm-agent-python/pull/1549

PR #1610, opened 2022-08-11
Add body length check for Django's capture_body

### LINKED ISSUE #1478: APM agent crashes if the request body is too large
**Describe the bug**: ...
When the request body is too large, reporting to the apm server fails. 
The error message is: 
```
eapm event processor thread elasticapm.transport:286 - Failed to submit message: 'HTTP 400: {"accepted":4,"errors":[{"message":"event exceeded the permitted size." ...
```

**To Reproduce**

1. Check the body's size before reporting. If it's too large, report the body as `[Too large]` or something like that.

- OS: mac
- Python version: 3.7
- Framework and version [e.g. Django 2.1]: django==2.2.26
- APM Server version: ?
- Agent version: 5.10.1

Elastic apm configuration (in `settings.py`): 
```
ELASTIC_APM = {
	...
    "CAPTURE_BODY": "all",
}
```

**Additional context**

The issue is that Django client extracts the request body [without checking the size of the body](https://github.com/elastic/apm-agent-python/blob/main/elasticapm/contrib/django/client.py#L151-L152). I believe that adding a size check in [this](https://github.com/elastic/apm-agent-python/blob/97188c362497910762e5b1ef8fc53e4011ec75b8/elasticapm/contrib/django/client.py#L117) function should solve the issue.

basepi:

We had no restrictions on the body length in Django. This could result in the event payload being rejected by the APM Server.

As an aside, we'll need to figure out a better solution in the long term. (#305)

## Related issues
closes #1478


elasticapm/contrib/django/client.py
```diff
@@ -149,7 +150,12 @@ def get_data_from_request(self, request, event_type):
                         self.logger.debug("Can't capture request body: %s", str(e))
                         data = "<unavailable>"
                 if data is not None:
-                    result["body"] = data
+                    # This can be removed once we figure out a soluction to
+                    # https://github.com/elastic/apm-agent-python/issues/305
+                    if len(str(data)) > constants.KEYWORD_MAX_LENGTH:
```
trentm, line 155: There is https://github.com/elastic/apm/blob/main/specs/agents/field-limits.md#long_field_max_length-configuration in the specs. Is that something that is better done as part of #305?

tests/contrib/django/django_tests.py
```diff
@@ -848,7 +848,7 @@ def test_request_metrics_301_append_slash(django_elasticapm_client, client):
         **middleware_setting(
             django.VERSION,
             ["elasticapm.contrib.django.middleware.TracingMiddleware", "django.middleware.common.CommonMiddleware"],
-        )
+        ),
```
trentm, line 851: Are these changes from a code-formatter change from a while ago that is only applied as a particular file is updated?

elasticapm/contrib/django/client.py
```diff
@@ -149,7 +150,12 @@ def get_data_from_request(self, request, event_type):
...
+                        result["body"] = keyword_field(str(data))
```
trentm, line 156: Is there a potential perf concern with `str(data)`ing twice?

trentm (approved):

elasticapm/contrib/django/client.py
basepi: Ah, I had not realized that it was extended to more than `db.statement` -- I found the old `apm-dev` thread but didn't realize it had been made applicable to request body. I'll update and add that config.

tests/contrib/django/django_tests.py
basepi: Correct. I touched the file so `black` hit it. We don't run `black` on the test suite, a decision made before I joined Elastic that could probably be revisited.

elasticapm/contrib/django/client.py
basepi: Hmm. Possibly. I'll store it in a temporary variable.

trentm (approved):

PR #1613, opened 2022-08-16
Remove zerorpc (py2 only)

### LINKED ISSUE #1020: Remove py2 leftovers
This is an issue to track all of the py2 stuff we can remove once a sufficient waiting period after 6.0 has passed.

- [x] Remove `zerorpc` (py2 only)
- [x] Remove `compat.py`
- [x] Run `pyupgrade`?

Ref #928

basepi:

Removes zerorpc support. I also ran `pyupgrade` but I think @beniwohli must have already run it back when he removed all the py2 `compat` code, because it didn't make any changes, at least with `--py36-plus`.

## Related issues
closes #1020


basepi:
Note for @trentm, we haven't supported py2 since 6.0, this is just cleanup. None of these tests have been running, and if a user were trying to use py2 with the agent we'd have failed at startup so this code can't actually be used.

trentm (approved):
Does "(py2 only)" here mean that zerorpc only works with Python 2.x? Or that the APM Python agent's instrumentation of zerorpc only works with Python 2.x?

https://pypi.org/project/zerorpc/ has these specifiers:

```
Python
    Python :: 2.6
    Python :: 2.7
    Python :: 3.4
```

I guess that indirectly indicates this library isn't maintained for Python 3, given the last 3.4.x was from 2019 (https://peps.python.org/pep-0429/).

basepi:
> Does "(py2 only)" here mean that zerorpc only works with Python 2.x? Or that the APM Python agent's instrumentation of zerorpc only works with Python 2.x?

`zerorpc` only works with python 2.x (and apparently a single version of EOL 3.x, as you discovered).

PR #1616, opened 2022-08-18
use GitHub actions for linting

v1v:

use GitHub actions for linting


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,27 @@
...
+        extra_args: --help
+    - id: files
+      uses: jitterbit/get-changed-files@b17fbb00bdc0c0f63fcf166580804b4d2cdc2a42
+    - name: Configure PATH
+      run: echo "${GITHUB_WORKSPACE}/.ci/scripts" >> $GITHUB_PATH
+    - name: Precommit changes
+      run: |
+        for changed_file in ${{ steps.files.outputs.all }}; do
+          pre-commit run --files "${changed_file}"
+        done
```
v1v, line 27: I could not find a better way to run `pre-commit` for the subset of changes.

.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,27 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v3
+    - uses: pre-commit/action@646c83fcd040023954eafda54b4db0192ce70507
+      with: ## let's skip to run for all the files
+        extra_args: --help
+    - id: files
+      uses: jitterbit/get-changed-files@b17fbb00bdc0c0f63fcf166580804b4d2cdc2a42
+    - name: Configure PATH
+      run: echo "${GITHUB_WORKSPACE}/.ci/scripts" >> $GITHUB_PATH
```
v1v, line 22: There are no special scripts so it's not needed
```suggestion
```

.github/workflows/pre-commit.yml
basepi: Honestly I'd be fine with running `pre-commit` on everything. It should be very fast. But this also seems fine 🤷‍♂️

beniwohli: Wouldn't this also work without the for loop?

    pre-commit run --files ${{ steps.files.outputs.all }}

[`pre-commit run --files`](https://pre-commit.com/#pre-commit-run) can take a list of files. Not sure if we could run into some limit if the change affects hundreds of files, though (e.g. when Python 4.0 moves from spaces to tabs ;) )

v1v: good point, I'll try it in a bit :)

v1v:
Closed this PR, since it was much nicely done in https://github.com/elastic/apm-agent-python/pull/1658 :)

PR #1618, opened 2022-08-18
Fix error with elasticsearch(sniff_on_start=True)

### LINKED ISSUE #1615: enabling sniff_on_start in elasticsearch() breaks elasticserach instrumentation
**Describe the bug**:
sniff_on_start in elasticsearch with apm enabled fails the request due to a null span object
```python
ERROR File /opt/venv/lib/python3.9/site-packages/django/utils/log.py, line 224, Internal Server Error: /api/test_view
  File "/opt/venv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/opt/venv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/opt/venv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/opt/venv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/app/api/views.py", line 592, in get
    es = Elasticsearch(hosts=["elasticsearch"],sniff_on_start=True)
  File "/opt/venv/lib/python3.9/site-packages/elasticsearch/client/__init__.py", line 205, in __init__
    self.transport = transport_class(_normalize_hosts(hosts), **kwargs)
  File "/opt/venv/lib/python3.9/site-packages/elasticsearch/transport.py", line 200, in __init__
    self.sniff_hosts(True)
  File "/opt/venv/lib/python3.9/site-packages/elasticsearch/transport.py", line 365, in sniff_hosts
    node_info = self._get_sniff_data(initial)
  File "/opt/venv/lib/python3.9/site-packages/elasticsearch/transport.py", line 308, in _get_sniff_data
    _, headers, node_info = c.perform_request(
  File "/opt/venv/lib/python3.9/site-packages/elasticapm/instrumentation/packages/base.py", line 210, in call_if_sampling
    return self.call(module, method, wrapped, instance, args, kwargs)
  File "/opt/venv/lib/python3.9/site-packages/elasticapm/instrumentation/packages/elasticsearch.py", line 58, in call
    self._update_context_by_request_data(span.context, instance, args, kwargs)
AttributeError: 'NoneType' object has no attribute 'context'

```

**To Reproduce**

1. setup an elasticsearch instance or a cluser and a django app
2. create test view to create elasticsearch connection instance
 ```python
from elasticsearch import Elasticsearch
from django.http import HttpResponse
from django.views import View

class TestView(View):
    def get(self, request, *args, **kwargs):
        es = Elasticsearch(hosts=["elasticsearch"],sniff_on_start=True)
        return HttpResponse("Success !")
```
4. curl that view 

**Environment**
- OS:  debian buster docker
- Python version: 3.9
- Framework and version : django 3.2.15
- elasticserach-py: 7.17.14
- elasticsearch: 7.15.3
- APM Server version: 7.10.2
- Agent version: 6.10.1

basepi:

Checks if the span object is non-`None` before trying to access its context.

No test attached. I've been unable to get one to fail even without the fix....it may be that we're missing something in the user's edge case. In any case, I think this is a change that should fix the user's issue, and is extremely low-risk. So I think we can forgo the test.

## Related issues
Closes #1615

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -41,6 +41,7 @@ endif::[]
 
 * Differentiate Lambda URLs from API Gateway in AWS Lambda integration {pull}#1609[#1609]
 * Restrict the size of Django request bodies to prevent APM Server rejection {pull}#1610[#1610]
+* Fix error when using elasticsearch(sniff_on_start=True) {pull}#1618[#1618]
 
 
 
--- a/elasticapm/instrumentation/packages/elasticsearch.py
+++ b/elasticapm/instrumentation/packages/elasticsearch.py
@@ -52,7 +52,7 @@ class ElasticsearchConnectionInstrumentation(AbstractInstrumentedModule):
 
     def call(self, module, method, wrapped, instance, args, kwargs):
         span = execution_context.get_span()
-        if isinstance(span, DroppedSpan):
+        if not span or isinstance(span, DroppedSpan):
             return wrapped(*args, **kwargs)
 
         self._update_context_by_request_data(span.context, instance, args, kwargs)
--- a/tests/instrumentation/elasticsearch_tests.py
+++ b/tests/instrumentation/elasticsearch_tests.py
@@ -41,7 +41,6 @@ from elasticsearch.serializer import JSONSerializer
 
 import elasticapm
 from elasticapm.conf.constants import TRANSACTION
-from elasticapm.traces import execution_context
 
 pytestmark = [pytest.mark.elasticsearch]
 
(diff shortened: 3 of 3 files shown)


trentm (approved):
Seems reasonable (not knowing the agent internals at all well :). It looks like redis.py and aioredis.py have a similar guard.

There is a call to `execution_context.get_span()` in "asyncio/elasticsearch.py" as well that doesn't have a guard. Perhaps that should eventually get it as well.

PR #1619, opened 2022-08-18
Restrict length of exception messages

### LINKED ISSUE #1614: FastAPI/Starlette: Truncate exception message to prevent "event exceeded the permitted size" error from APM Server
**Describe the bug**: 
The capture_exception() method is sending the whole exception message to the APM Server and when the message exceeds the maximum size, the APM Server returns the  "event exceeded the permitted size" error. 

I'm sorry if it is not a bug, but I believe that the apm-agent should truncate the message before sending it to the APM Server.

**To Reproduce**
Create a FastAPI route that always raise an error with a large exception message (as the example below) and it will show an "event exceeded the permitted size" error on the next attempt to send events to APM server

```
@app.get('/apm_test/')
async def apm_test():
    try:
        raise Exception('test' * 1000000)
    except Exception as e:
        apm_client.capture_exception()
    return PlainTextResponse('', 200)
```

**Environment**
- OS: ubuntu 20.04
- Python version: 3.9.13
- Framework and version: fastapi==0.79.0
- APM Server version: 8.3.0
- Agent version: 6.11.0

basepi:

Restricts the length of the `message` field for exceptions captured by the agent. Continuation of the theme in #305.

## Related issues
Closes #1614

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -41,6 +41,7 @@ endif::[]
 
 * Differentiate Lambda URLs from API Gateway in AWS Lambda integration {pull}#1609[#1609]
 * Restrict the size of Django request bodies to prevent APM Server rejection {pull}#1610[#1610]
+* Restrict length of `exception.message` for exceptions captured by the agent {pull}#1619[#1619]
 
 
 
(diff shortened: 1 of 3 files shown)


trentm (approved):
LGTM.

PR #1628, opened 2022-09-07
Fix empty flushed=True request for forced flushes

### LINKED ISSUE #1627: unexpected keyword argument 'flushed' in lambda extension
**Describe the bug**: 
Call self.send(data,flushed=true) in https://github.com/elastic/apm-agent-python/blob/main/elasticapm/transport/base.py#L183 uses not declared argument.

**To Reproduce**

1. we use elastic-apm-extension-ver-1-0-0-x86_64 as lambda layer 
2. use a newer version of elastic-apm=">6.8.1"
3. Result : TypeError: send() got an unexpected keyword argument 'flushed'

- OS: N/A - use lambda
- Python version: any
- Framework and version [e.g. Django 2.1]: None
- APM Server version: 1.0.0
- Agent version: >6.8.1

basepi:

It turns out I never actually tested the functionality for sending an empty request with `flushed=True` when an explicit flush is called with no data on the buffer. The function signature was wrong (😅), and we also can't POST with a `None` body, we need an empty dict for the body.

Luckily, it should be rare to hit this issue because we usually flush before we've had a chance to send the contents of the buffer, and that `flushed=True` worked fine.

This fixes those issues and adds a test for this explicit case.

Thanks to @josef-cech-acc for catching this.

## Related issues

Closes #1627

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -180,7 +180,7 @@ class Transport(ThreadManager):
                     # No data on buffer, but due to manual flush we should send
                     # an empty payload with flushed=true query param, but only
                     # to a local APM server (or lambda extension)
-                    self.send({}, forced_flush=True)
+                    self.send("", forced_flush=True)
                 self._last_flush = timeit.default_timer()
                 buffer = self._init_buffer()
                 buffer_written = False
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -176,11 +176,11 @@ class Transport(ThreadManager):
             if flush:
                 if buffer_written:
                     self._flush(buffer, forced_flush=forced_flush)
-                elif forced_flush and "/localhost:" in self.client.config.server_url:
+                elif forced_flush and any(x in self.client.config.server_url for x in ("/localhost:", "/127.0.0.1:")):
                     # No data on buffer, but due to manual flush we should send
                     # an empty payload with flushed=true query param, but only
                     # to a local APM server (or lambda extension)
-                    self.send(None, flushed=True)
+                    self.send({}, forced_flush=True)
                 self._last_flush = timeit.default_timer()
                 buffer = self._init_buffer()
                 buffer_written = False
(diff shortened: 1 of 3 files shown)


elasticapm/transport/base.py
```diff
@@ -176,11 +176,11 @@ def _process_queue(self):
             if flush:
                 if buffer_written:
                     self._flush(buffer, forced_flush=forced_flush)
-                elif forced_flush and "/localhost:" in self.client.config.server_url:
+                elif forced_flush and any(x in self.client.config.server_url for x in ("/localhost:", "/127.0.0.1:")):
```
trentm, line 179: How much do you care about possible IPv6 usage? :)

elasticapm/transport/base.py
```diff
@@ -176,11 +176,11 @@ def _process_queue(self):
...
                     # No data on buffer, but due to manual flush we should send
                     # an empty payload with flushed=true query param, but only
                     # to a local APM server (or lambda extension)
-                    self.send(None, flushed=True)
+                    self.send({}, forced_flush=True)
```
trentm, line 183: That `Transport.send()` passes `data` to `self.http.urlopen(..., body=data, ...)`, which is https://urllib3.readthedocs.io/en/stable/reference/urllib3.connectionpool.html#urllib3.HTTPConnectionPool.urlopen which defines:

> body – Data to send in the request body, either [str](https://docs.python.org/3/library/stdtypes.html#str), [bytes](https://docs.python.org/3/library/stdtypes.html#bytes), an iterable of [str](https://docs.python.org/3/library/stdtypes.html#str)/[bytes](https://docs.python.org/3/library/stdtypes.html#bytes), or a file-like object.

So would `b''` or `''` be more appropriate than `{}` for an empty value?

elasticapm/transport/base.py
basepi: I don't think we care for this particular instance -- seems likely that most users would use `localhost` -- the only reason I added `127.0.0.1` was because the testing http server uses that as its URL base.

Those using the local APM server should just use our default `server_url`, which is `'http://localhost:8200'`

elasticapm/transport/base.py
basepi: Good catch, thanks.

trentm (approved):

PR #1629, opened 2022-09-08
fixed error: "TypeError: 'str' object does not support item assignment"

ritesd:

File "site-packages\elasticapm\handlers\structlog.py", line 63, in structlog_processor
    event_dict["service.name"] = client.config.service_name
TypeError: 'str' object does not support item assignment

I was getting above error. When i looked into code and debug I found event_dict is expected to be as dictionary object but it was 
getting passed as str object and so i put check over there and incase of str i converted to dictionary.
after the all the value assignments i converted it back to str as it is expected to return str object

below is sample configuration to generate this issue
structlog.configure(
    processors=[
        structlog.processors.add_log_level,
        structlog.processors.StackInfoRenderer(),
        structlog.dev.set_exc_info,
        structlog.processors.ExceptionPrettyPrinter(),
        structlog.processors.UnicodeDecoder(),
        structlog.processors.TimeStamper(fmt="iso"),
        structlog.processors.JSONRenderer(),
        structlog_processor
    ],
    wrapper_class=structlog.stdlib.BoundLogger,
    context_class=dict,
    logger_factory=structlog.stdlib.LoggerFactory(),
    cache_logger_on_first_use=True,
)

## Related issues
I have not found issue reported by other. I faced this problem so fixed it.
Closes #ISSUE

### DIFF
--- a/elasticapm/handlers/structlog.py
+++ b/elasticapm/handlers/structlog.py
@@ -31,6 +31,7 @@ from __future__ import absolute_import
 from elasticapm import get_client
 from elasticapm.traces import execution_context
 
+import json
 
 def structlog_processor(logger, method_name, event_dict):
     """
@@ -51,6 +52,7 @@ def structlog_processor(logger, method_name, event_dict):
     :return:
         `event_dict`, with three new entries.
     """
+    event_dict = json.loads(event_dict) if isinstance(event_dict, str) else event_dict
     transaction = execution_context.get_transaction()
     if transaction:
         event_dict["transaction.id"] = transaction.id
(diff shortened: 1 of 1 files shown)


basepi:
I don't think this is quite right. I think this error is due to you having your structlog processors in the wrong order. Ours should be before the `structlog.processors.JSONRenderer()`, as we want to enrich the event dict before it is converted to a string. I think if you reorder your list it will fix your issue.

This PR unfortunately would make our processor convert everything to JSON which is not a side effect I think we want. Hope that makes sense.

ritesd:
oh nice!, Yes by changing the order it dose solve my issue. I think JSONRenderer should always be at last. Thank you ;)

PR #1631, opened 2022-09-13
Fix some flaky tests

basepi:

Fixes a couple of flaky tests since the introduction of https://github.com/elastic/apm-agent-python/pull/1628


tests/client/client_tests.py
```diff
@@ -254,6 +254,7 @@ def test_send_remote_failover_sync(should_try, sending_elasticapm_client, caplog
     sending_elasticapm_client.httpserver.code = 202
     sending_elasticapm_client.capture_message("bar", handled=False)
     sending_elasticapm_client.close()
+    time.sleep(0.2)
```
beniwohli, line 257: It's weird that this has any effect. Calling `close()` on the agent should clear out the queue and close the connection synchronously. All the background threads should have `join`ed, so there shouldn't be anything running in the background that needs to be waited for...theoretically. Might be worth looking into this at some point.

beniwohli (approved):

tests/client/client_tests.py
basepi: Fixed it properly this time. The culprit was actually [this line](https://github.com/elastic/apm-agent-python/blob/ef9fb4ba1cf59126aff928de4e6e85d0d92933ed/tests/client/client_tests.py#L249), which makes *so much more sense* with the change that made this test flaky.

beniwohli (approved):
Great find!

PR #1632, opened 2022-09-14
Disable checks when elastic_apm is disabled

francoisfreitag:
Avoids warning about an incorrect SERVER_URL when the tool is not in use.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,19 @@ endif::[]
 //===== Bug fixes
 //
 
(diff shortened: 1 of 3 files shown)


basepi:
The tests will succeed once #1631 is merged. In the meantime, can you please sign the CLA when you get a chance? Your commits are signed with `mail@franek.fr` for reference.

francoisfreitag:
Thanks for the feedback! I believe the CLA is signed, DocuSign says it’ll be emailed to me when all parties have signed.

PR #1635, opened 2022-09-15
Doesn"t cut request body with size >= 10000

mukhamux:
b4edea3ca7885f5cbbad202a8f12325d131225ce has errors, if body size has more than 10000 bytes, it writes 10000 bytes message to elastic and call await self.app(scope, receive, wrapped_send) with same receive and invalid body, 

## Comment:
I separate receivers and now, if request has body more than 10000 bytes -> it will send to elastic only 10000 bytes and continue handle as normal request without cutting


mukhamux:
Find strange bug in API realted with apm agent is more easier than make a pull request...

mukhamux:
run elasticsearch-ci/docs

basepi:
Sorry for the confusion, but only people with write access to the repo can trigger the tests. I'll take a look at this today.

basepi (approved):
Wow this definitely was broken, thank you for the catch and the fix! 😅

basepi:
I opened a PR to add a CHANGELOG entry and fix the test to test this: https://github.com/mukhamux/apm-agent-python/pull/1

mukhamux:
> I opened a PR to add a CHANGELOG entry and fix the test: [mukhamux#1](https://github.com/mukhamux/apm-agent-python/pull/1)

merged

Issue #1636, opened 2022-09-16
Flask instrumentation does not capture transaction outcome properly during exception


LucaWintergerst:
**Describe the bug**: ...

Running flask, when an exception is raised and not caught, 

`
transaction.result: HTTP 5xx
transaction.outcome: unknown
`

this happens as the flask instrumentation calls the end transaction function 

https://github.com/elastic/apm-agent-python/blob/77c2c0d01a86a4904b669172424c926f9245beec/elasticapm/contrib/flask/__init__.py#L114

which gets the currently active transaction with `clear=True`

https://github.com/elastic/apm-agent-python/blob/77c2c0d01a86a4904b669172424c926f9245beec/elasticapm/traces.py#L985

when the request is finished, the set_transaction_outcome function is called again

https://github.com/elastic/apm-agent-python/blob/77c2c0d01a86a4904b669172424c926f9245beec/elasticapm/contrib/flask/__init__.py#L194

but it will fail to get the current transaction since it has already been ended and removed from the context

https://github.com/elastic/apm-agent-python/blob/77c2c0d01a86a4904b669172424c926f9245beec/elasticapm/traces.py#L1182

I will open a PR with one fix that works for me, but someone would need to confirm that it's the right way of fixing this.

Before:
<img width="1513" alt="Screenshot 2022-09-16 at 16 29 38" src="https://user-images.githubusercontent.com/11661400/190663115-d9cec492-8745-4758-8ab5-bd9865e0c89b.png">

After:
<img width="1511" alt="Screenshot 2022-09-16 at 16 28 40" src="https://user-images.githubusercontent.com/11661400/190662914-0e839b76-9d16-437e-bb2d-19b35e4acd62.png">

PR #1637, opened 2022-09-16
properly set transaction outcome

LucaWintergerst:
properly set transaction outcome for flask instrumentation when an exception occurs

## Related issues

Closes [#1636](https://github.com/elastic/apm-agent-python/issues/1636)


basepi (approved):
I can't see any reason this fix will be problematic. Thank you for tracking this down and fixing it!

LucaWintergerst:
And thank you for taking a look at this so quickly!

PR #1642, opened 2022-09-20
Elasticsearch 8 support

### LINKED ISSUE #1617: Update elasticsearch-py instrumentation for 8.0
While figuring out how to test my fix for https://github.com/elastic/apm-agent-python/issues/1615, I realized that with 8.0, `elasticsearch-py` has moved its transport and connection code to a new project, `elastic-transport`.

I also realized we are not testing against elasticsearch-py 8.x. 😅 

Everything may still work with the [stub file provided](https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/transport.py). But since importing that file is deprecated, we should add an instrumentation that targets `elastic-transport` directly.

/cc @sethmlarson @miriam-eid

beniwohli:

Add support and testing for elasticsearch-py 8.x

## Related issues

Closes #1617

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -76,6 +76,22 @@ services:
     volumes:
       - pyesdata7:/usr/share/elasticsearch/data
 
+  elasticsearch8:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.4.0
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:9200"]
+    ports:
+      - "9200:9200"
+    environment:
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+      - "network.host=_site_"
+      - "transport.host=127.0.0.1"
+      - "http.host=0.0.0.0"
+      - "xpack.security.enabled=false"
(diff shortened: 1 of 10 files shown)


tests/docker-compose.yml
```diff
@@ -76,6 +76,22 @@ services:
     volumes:
       - pyesdata7:/usr/share/elasticsearch/data
 
+  elasticsearch8:
+    image: docker.elastic.co/elasticsearch/elasticsearch:8.4.0
```
basepi, line 80: Is there a moving tag we can target with `8.x` instead of `8.4` so this stays up to date? I notice above we've been on 7.9.0 for awhile for 7.x testing.

basepi (approved):
Looks great. Thanks for doing this -- it's such a pain when it's *almost* the same in the new version but not *quite*.

tests/docker-compose.yml
beniwohli: No, for some reason, we don't push moving tags to docker hub, see https://hub.docker.com/_/elasticsearch/tags

basepi: 👍

PR #1648, opened 2022-09-27
fix errors in pymongo tests introduced in #1639

beniwohli:


beniwohli:
the remaining test failures are with the starlette instrumentation, probably due to https://github.com/encode/starlette/pull/1376

basepi (approved):

PR #1652, opened 2022-09-28
Use minimum of 5 seconds for interval between central config calls

### LINKED ISSUE #1645: [META 690] Ensure a minimum retry interval of 5 seconds in fetching central configuration
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/690

beniwohli:

Handling of negative values is not implemented, as this is not allowed per RFC: https://www.rfc-editor.org/rfc/rfc7234#section-5.2.2.8

## Related issues

closes #1645

### DIFF
--- a/elasticapm/transport/http.py
+++ b/elasticapm/transport/http.py
@@ -159,11 +159,9 @@ class Transport(HTTPTransportBase):
             logger.debug("HTTP error while fetching remote config: %s", str(e))
             return current_version, None, max_age
         body = response.read()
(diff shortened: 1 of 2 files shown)


beniwohli:
Regarding negative values, turns out we "accidentally" already behave as intended by the updated spec, on account of the regex failing to match negative values and falling back to the default value. I added a test to make sure we don't accidentally fix that accident.

basepi (approved):
👍

PR #1655, opened 2022-09-30
Use typing.TypeVar on decorators' type hints

### LINKED ISSUE #1654: Use typing.TypeVar on methods used as annotations
When I annotate a method with, for example, `@capture_span`, the method loses its type hints and becomes just a `Callable`.
For example, `Pylance` gives the following information tooltips (as commented to the right of the `reveal_type` calls):

```python
import elasticapm as apm
from typing_extensions import reveal_type


def test (a: int, b: str, c: float) -> bytes:
    return b"abc"

@apm.capture_span()
def test_with_apm (a: int, b: str, c: float) -> bytes:
    return b"abc"

reveal_type(test)  # Type of "test" is "(a: int, b: str, c: float) -> bytes"
reveal_type(test_with_apm)  # Type of "test_with_apm" is "(...) -> Any"
```

Just use `TypeVar` on the `__call__` method signature of annotation methods, example:
```python
from typing import TypeVar

_AnnotatedFunctionT = TypeVar("_AnnotatedFunctionT")

class capture_span(object):
    ...
    def __call__(self, func: _AnnotatedFunctionT) -> _AnnotatedFunctionT:
        self.name = self.name or get_name_from_func(func)

        @functools.wraps(func)
        def decorated(*args, **kwds):
            with self:
                return func(*args, **kwds)

        return decorated
    ...
```
After this small change, the `reveal_type` reports the expected type:
```python
import elasticapm as apm
from typing_extensions import reveal_type

@apm.capture_span()
def test_with_apm (a: int, b: str, c: float) -> bytes:
    return b"abc"

reveal_type(test_with_apm)  # Type of "test_with_apm" is "(a: int, b: str, c: float) -> bytes"
```

I've considered creating a pull request myself to solve this, but the [`CONTRIBUTING.md` file](https://github.com/elastic/apm-agent-python/blob/7d09ed8959afb2f2bd2e011969d2a6d3fdd6cd28/CONTRIBUTING.md) indicates that an issue should be opened first.

**Additional context**
Screenshot of VSCode's pylance type hints **without** the change:
apm_test_without_typevar

Screenshot of VSCode's pylance type hints **with** the change:
apm_test_with_typevar

marcoffee:

This PR adds `typing.TypeVar` type hints to decorator methods. 

## Related issues

Closes #1654

### DIFF
--- a/elasticapm/contrib/asyncio/traces.py
+++ b/elasticapm/contrib/asyncio/traces.py
@@ -30,15 +30,17 @@
 
 import functools
 from types import TracebackType
-from typing import Optional, Type
+from typing import Optional, Type, TypeVar
 
 from elasticapm.conf.constants import LABEL_RE
 from elasticapm.traces import SpanType, capture_span, execution_context
 from elasticapm.utils import get_name_from_func
 
+_AnnotatedFunctionT = TypeVar("_AnnotatedFunctionT")
+
 
 class async_capture_span(capture_span):
-    def __call__(self, func):
+    def __call__(self, func: _AnnotatedFunctionT) -> _AnnotatedFunctionT:
         self.name = self.name or get_name_from_func(func)
 
         @functools.wraps(func)
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -35,7 +35,7 @@ import json
 import os
 import platform
 import time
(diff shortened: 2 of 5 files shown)


basepi (approved):
Looks great! Thank you for implementing this!

beniwohli (approved):
Awesome!

PR #1658, opened 2022-10-04
use github actions for pre-commit checks

beniwohli:
## Related issues

Closes #ISSUE

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,16 @@
 repos:
 -   repo: https://github.com/pycqa/isort
-    rev: 5.7.0
+    rev: 5.10.1
     hooks:
     -   id: isort
 -   repo: https://github.com/ambv/black
-    rev: 22.3.0
+    rev: 22.8.0
     hooks:
     -   id: black
         language_version: python3
+        exclude: elasticapm\/utils\/wrapt
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 3.9.2
     hooks:
     -   id: flake8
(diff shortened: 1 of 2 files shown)


v1v (approved):

basepi (approved):
Looks great. Once we merge this (probably with admin privs) we'll want to update the branch protections to target the new test.

.pre-commit-config.yaml
```diff
@@ -1,15 +1,16 @@
...
 -   repo: https://github.com/ambv/black
-    rev: 22.3.0
+    rev: 22.8.0
     hooks:
     -   id: black
         language_version: python3
+        exclude: elasticapm\/utils\/wrapt
 -   repo: https://gitlab.com/pycqa/flake8
-    rev: 3.8.4
+    rev: 3.9.2
```
basepi, line 13: why not latest? 5.0.4

And we should target github, not their mirror on gitlab.

beniwohli: good catch, updated in 7823421b

kuisathaverat: to avoid have to chekc for new version you can add `.github/dependabot.yaml` file like this one

```yaml
version: 2
updates:
# GitHub actions dependencies
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
# Python dependencies
  - package-ecosystem: "pip"
    directory: "/"
    schedule:
      interval: "weekly"
```

beniwohli: @kuisathaverat from what I can tell, dependabot doesn't support updating the pre-commit hooks themselves: https://github.com/dependabot/dependabot-core/issues/1524

PR #1661, opened 2022-10-05
bump python versions and use docker slim versions

v1v:
### What does this pull request do?

Use smaller docker images (slim) and bump the version to use the supported python version for some of the scripts

### Why

image

vs 

image

Then the docker image with all the tooling is about 2x smaller:

> REPOSITORY                                          TAG                                                                IMAGE ID       CREATED          SIZE
apm-agent-python                                    python-3.6                                                         50ecd8523761   51 seconds ago   462MB

vs

> REPOSITORY                                          TAG                                                                IMAGE ID       CREATED          SIZE
apm-agent-python                                    python-3.6                                                         8e23ab62a462   16 seconds ago   961MB

### Implementation details

There were some missing packages required to run some of the tests, such as:

* `libmariadb-dev` to fix `mariadb_config: not found` when using the framework `mysqlclient-newest`
* `libpq-dev` to fix `pg_config executable not found` when using the framework `psycopg2-newest`
* `build-essential` to fix `gcc executable not found` when using the framework `twisted`
* `make` to fix the run of all the tests.
* `gpg` to fix the docker image generation.

### DIFF
--- a/tests/scripts/docker/black.sh
+++ b/tests/scripts/docker/black.sh
@@ -7,7 +7,7 @@ docker_pip_cache="/tmp/cache/pip"
 
 cd tests
 
-docker build --build-arg PYTHON_IMAGE=python:3.6 -t python-linters .
+docker build --build-arg PYTHON_IMAGE=python:3.8 -t python-linters .
 docker run \
   -e LOCAL_USER_ID=$UID \
   -e PIP_CACHE=${docker_pip_cache} \
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -7,6 +7,7 @@ RUN apt-get -qq update && apt-get -qq -y --no-install-recommends install \
     curl \
     default-mysql-server \
     gnupg2 \
+    libmariadb-dev \
     libpq-dev \
     make \
     netcat \
--- a/tests/Dockerfile
+++ b/tests/Dockerfile
@@ -5,8 +5,8 @@ RUN apt-get -qq update && apt-get -qq -y --no-install-recommends install \
     build-essential \
     ca-certificates \
     curl \
+    default-mysql-server \
     gnupg2 \
-    libmysqlclient-dev \
     libpq-dev \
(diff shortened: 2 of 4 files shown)


basepi (approved):

v1v:
I'll run a few builds in the CI to discard any issues with the full test coverage

amannocci (approved):
LGTM

Small suggestion :
```suggestion
# Use --no-cache-dir flag.
# pre-install base requirements
RUN python -m pip install --no-cache-dir --progress-bar off -U pip && pip install --no-cache-dir --progress-bar off -r /reqs-base.txt
```

tests/scripts/docker/black.sh
```diff
@@ -7,7 +7,7 @@ docker_pip_cache="/tmp/cache/pip"
 
 cd tests
 
-docker build --build-arg PYTHON_IMAGE=python:3.6 -t python-linters .
+docker build --build-arg PYTHON_IMAGE=python:3.8 -t python-linters .
```
beniwohli, line 10: I wonder if these scripts are still in use. We use pre-commit for black/isort/flake8. The only reference I could find is in [`run-tests-locally.asciidoc`](https://github.com/elastic/apm-agent-python/blob/main/docs/run-tests-locally.asciidoc#code-linter)

From what I can tell, we might as well just delete these three scripts.

beniwohli (approved):

tests/scripts/docker/black.sh
v1v: Good point, let me review it in a follow up, as I don't wanna change any further behaviours in this PR but only bump the version to use a slim one and default to a major python version in those scripts.

v1v:
> Small suggestion :

Thanks @amannocci , this change was not aimed to modify any of the internals but the docker image base image and default python version, for now I'll not make any further changes so this can be merged and in a follow up the implementation can be revisited 🙏

PR #1665, opened 2022-10-12
skaffold: first attempt to run unit-test

v1v:

Use https://skaffold.dev/docs/ to leverage the testing of several python versions against several frameworks.

## Actions

- [x] Add support for more versions.
- [x] Docs for the how to use it

## How to use it

See https://github.com/elastic/apm-agent-python/pull/1665/files#diff-92f8b8798d32cfd383b2c7d24b5940990d88714525869eb264a3f2881db12895 but in a nutshell

### Generate the `skaffold` filesystem

```bash
$ python3 .k8s/cli.py generate
```

### Generate the docker images synced with your workspace

```bash
$ python3 .k8s/cli.py build
```

### Test agains a list of frameworks

```bash
$ python3 .k8s/cli.py test -f django -f none -f opentelemetry -f psutil
```

Then the status of those jobs running on K8s are reported in the console:

image

and logs can be found in the `build` folder:

```bash
$ ls -l build 
total 176
-rw-r--r-- 1 vmartinez staff 23319 Oct 12 15:59 apm-agen-python-3-10-django-3-1.log
-rw-r--r-- 1 vmartinez staff 22924 Oct 12 15:59 apm-agen-python-3-10-django-3-2.log
-rw-r--r-- 1 vmartinez staff 22922 Oct 12 15:59 apm-agen-python-3-10-django-4-0.log
-rw-r--r-- 1 vmartinez staff 78960 Oct 12 15:59 apm-agen-python-3-10-none.log
-rw-r--r-- 1 vmartinez staff 11494 Oct 12 15:59 apm-agen-python-3-10-opentelemetry-newest.log
-rw-r--r-- 1 vmartinez staff 11606 Oct 12 15:59 apm-agen-python-3-10-psutil-newest.log
```

### Actions

- [x] Configure the `skaffold` namespace
- [x] Configure the `skaffold` labels so they are unique per user

### Follow ups
- [ ] Generate output in a json/yaml/table format with the results.
- [ ] Export junit reports from the pods
- [ ] Run frameworks against services they need to run with.
- [ ] UTs for the existing python code


v1v:
@elastic/observablt-robots , I'd like to gather some feedback from you, this is just a PoC, just wanna validate if this way of running things in K8s could help us to reduce the complexity in the CI. I'll see if I can run some GitHub actions with this and get some idea how it behaves.

There is plenty of room for improving this, ideally a common tool could be used so the code won't be hosted in each project but they will use a binary or similar tool

.k8s/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Test django-1.1 in python-3.6
+$ python3 .k8s/cli.py test --framework django-1.1 --version python-3.6
+```
+
+Then the logs can be found under the `build/` folder.
+
+## Implementation details
+
+The list of frameworks and when they run can be found in `.ci`:
```
cachedout, line 66: I think we should talk a bit more about whether or not we really need to have this kind of separation between behavior in the CI and behavior when run locally.

IMHO, one of the goals of this project should be that the test framework *defaults* to the same configuration that the CI uses. A user shouldn't even need to know anything about a `.ci/` directory in order to execute the test framework in pretty much the exact same way that the CI will do it.

v1v: I agree, let me explain a bit

This particular PR is not intrusive in terms of changing files, and it's not intended at the moment, as it's targeting the upstream branch `feature/local-and-ci-on-k8s` for now, so there will be more PRs coming up after this one, and probably those PRs will be changing this current behaviour.

The current APM Agent Python uses two different workflows:
1) on a PR basis, it runs a subset of the matrix (python 3.6 and 3.10 with some frameworks)
2) On demand or merge commits, it runs the whole matrix (python 3.6, 3.7, ..., 3.10 with all the frameworks)

In this PR I used the same matrix, which its' already defined in some files stored in the `.ci` folder. It can be potentially changed to another location, that's totally fine, and I'm not against.

The default behaviour is matching the same default configuration that the CI uses, users won't need to do anything about the `.ci` folder. They can change the behaviour to use a one or the other workflows if needed but for such they need to regenerate the skaffold files `cli.py generate --version .ci/jenkins_python_full.yml --framework .ci/jenkins_framework_full.yml` then they can run the tests as they wish 

The CLI tool can be changed if needed, I tried to make as generic as possible, so we could potentially reuse it for some other APM Agents in the future, as there are three important things to care:
1) Matrix entries one file per axis
2) Exclude entries
3) YAML Keys

This is just a PoC, so I'd like to gather about the User eXperience and whether those commands in CLI are meaningful, useful, quick enough and valuable. 

Additionally, I worked in parallel to support a GH workflow integration in https://github.com/elastic/apm-agent-python/pull/1669 , at the moment, it works only for upstream branches (some security constrains to work on PRs), see https://github.com/elastic/apm-agent-python/actions/runs/3250381993/jobs/5333991415

Did I clarify a bit the status of this PR and what I envisioned for it?

v1v:
We are not working on this atm, so I'll close this for now

PR #1674, opened 2022-10-19
Add starlette ServerErrorMiddleware instrumentation

### LINKED ISSUE #1110: trace.id is not available in fastapi exception handler
**Describe the bug**: We have fastapi framework, and we add apm_client for starlette to the application, we want to generate trace.id to correlate logs when an exception happens. I have structlog package for the logging and import `elasticapm.handlers.structlog.structlog_processor`. We set exception handlers in fastapi application for our excpected handlers and also one for `Exception` class to capture the ones that we do not expect. I can get trace.id in child exceptions but not in the exception handler that is comparing to root. 

My goal was to have log Exception types and have log correlation in them.

I am not sure if it is bug or feature request🤷‍♂️
**To Reproduce**

1. create apm agent and fastapi app, add the apm agent to application
```
from fastapi import FastAPI
from elasticapm.contrib.starlette import make_apm_client
from elasticapm.contrib.starlette import ElasticAPM
apm = make_apm_client(config=config)
app = FastAPI()
```
2. configure structlog
```
import structlog
from elasticapm.handlers.structlog import structlog_processor
def processor(logger, level, log_dict):
    return json.dumps(log_dict)

structlog.configure(processors=[structlog_processor, processor])

```
3. create an child exceptions and root and child exception
```
class AuthException(Exception):
    def __ini__(self):
        super().__init__()


@app.exception_handler(Exception)
async def fallback_handler(request , exc: Exception):
    getLogger().info("Uncaught exception during request handling happend")
    return JSONResponse({"detail": "ROOT EXCEPTION"}, status_code=500)

@app.exception_handler(AuthException)
async def validation_exception_handler(request, exc: AuthException):
    getLogger().info(f"Request Validation exception happend{sys.exc_info()}")
    return JSONResponse({"detail": "CHILD EXCEPTION"}, status_code=404)
```
4. create an endpoint to send request to see if logging formatter/processor works or not
```
@app.post("/")
async def post(body: Any = Body(...)):
    if body == 1: raise AuthException("child exception")
    if body == 2: raise ZeroDivisionError("root exception")
    return JSONResponse({"Response": "test SUCCESS"}, status_code=200)
```
- OS: Win and Linux 
- Python version: 3.7
- Framework and version [e.g. Django 2.1]: fastapi 0.63
- APM Server version: _ 
- Agent version: 6.*

basepi:

Moves the `end_transaction()` into [Starlette's ServerErrorMiddleware](https://github.com/encode/starlette/blob/858629f5188bc79d452600b1eb90eaa0045f6454/starlette/middleware/errors.py#L125) to solve #1110.

Big thanks to @magichair for help figuring this one out.

## Related issues

Closes #1110


magichair:
I've got a development environment setup from this branch too, I'm taking a look. So far I can see from the DEBUG lines that that `__call__` method gets resolved and claims to get instrumented (and un-instrumented at test cleanup). But I am seeing the same behavior that it's seemingly never called.

magichair:
Yup, ok, I'm pretty sure I fully understand here. The `call_if_sampling` is deciding whether to call the original method or call the wrapped method if we currently have a transaction recording. However, because we're trying to instrument the `__call__` on ServerErrorMiddleware, that is the outermost middleware call, and our ElasticAPM middleware has not yet `begin_transaction()`'ed.

I think this really only leaves us with the option to force ElasticAPM to be the outermost in the `middleware_stack`. Unless y'all have some flag to tell the wrappers, "no, really, always instrument this regardless if we are in a transaction or not".

elasticapm/instrumentation/packages/asyncio/starlette.py
```diff
@@ -0,0 +1,46 @@
...
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+from elasticapm import get_client
+from elasticapm.instrumentation.packages.asyncio.base import AsyncAbstractInstrumentedModule
+
+
+class StarletteServerErrorMiddlewareInstrumentation(AsyncAbstractInstrumentedModule):
+    name = "starlette"
+
+    instrument_list = [("starlette.middleware.errors", "ServerErrorMiddleware.__call__")]
```
magichair, line 38:
```suggestion
    name = "starlette"
    
    instrument_list = [("starlette.middleware.errors", "ServerErrorMiddleware.__call__")]
    
    # We defer creating the transaction to the registered middleware.
    # However, since we are instrumenting the `__call__` of the outermost ServerErrorMiddleware
    # we need this flag so that it is always instrumented, otherwise no transaction is started until the nested
    # ElasticAPM middleware is called.
    creates_transaction = True
```

There are still a couple test errors I get when I run with this locally, but this seems to solve the bulk of the problem.

basepi: Good catch! Makes perfect sense that it's not calling the wrapped version when there's no transaction. I should have spotted that. I'll add this patch and see what I can do with the rest of the tests.

beniwohli (approved):
Nice solution!

PR #1683, opened 2022-10-26
Remove twine upload for (now-removed) wheels

basepi:

Should fix the upload "failure" as seen here: https://apm-ci.elastic.co/blue/organizations/jenkins/apm-agent-python%2Fapm-agent-python-mbp/detail/v6.13.0/1/pipeline

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -374,7 +374,6 @@ def releasePackages(){
     python setup.py sdist
     echo "Uploading to ${REPO_URL} with user \${TWINE_USER}"
     python -m twine upload --username "\${TWINE_USER}" --password "\${TWINE_PASSWORD}" --skip-existing --repository-url \${REPO_URL} dist/*.tar.gz
-    python -m twine upload --username "\${TWINE_USER}" --password "\${TWINE_PASSWORD}" --skip-existing --repository-url \${REPO_URL} wheelhouse/*.whl
     """)
   }
 }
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -374,7 +374,6 @@ def releasePackages(){
     python setup.py sdist
     echo "Uploading to ${REPO_URL} with user \${TWINE_USER}"
     python -m twine upload --username "\${TWINE_USER}" --password "\${TWINE_PASSWORD}" --skip-existing --repository-url \${REPO_URL} dist/*.tar.gz
-    python -m twine upload --username "\${TWINE_USER}" --password "\${TWINE_PASSWORD}" --skip-existing --repository-url \${REPO_URL} wheelhouse/*.whl
```
beniwohli, line 377: We still produce one .whl file (the universal one), but it's in `dist/`, not in `wheelhouse/`

beniwohli: That one only matches `*.tar.gz`, doesn't it?

basepi: Oh, right. So did it not get uploaded at all for 6.13? PyPI shows ok but I didn't look closely.

beniwohli: I guess it only uploaded the tar.gz package, which is a bit slower to install than the wheel. Not sure if this warrants a 6.13.1, though.

beniwohli (approved):

PR #1690, opened 2022-11-07
Use span.update_context to avoid DroppedSpan issues in Elasticsearch instrumentation

### LINKED ISSUE #1689: KeyError: 'db' for DroppedSpan when running scan query on Elasticsearch
**Describe the bug**: ...

Elastic APM fails with a `KeyError: db key not found`.

The application where this happens is a Django project that stores/reads data from Elasticsearch. I have APM enable (APM server and Elasticsearch cluster are both running on Elastic Cloud). The library fails with the aforementioned error (shown in the snippet and screenshots below) while running a scan query on Elasticsearch. It looks like it's dropping some spans, which ends up in this case:

```
            hits = self._get_hits(result_data)
            if hits:
                span.context["db"]["rows_affected"] = hits
 
            return result_data
```

here's a screenshot of what I see on the APM Error page:

image

Few variables from the context:

image

**To Reproduce**

Unfortunately, I don't have a reproducible snippet. 

- OS: Linux (containerized)
- Python version: 3.9.15
- APM Server version: 7.17.4
- Agent version: 

```
$ pip freeze | grep elastic
django-elasticsearch-dsl==7.2.2
elastic-apm==6.13.1
elasticsearch==7.17.4
elasticsearch-dsl==7.4.0
```

**Additional context**

basepi:

We were setting the context directly instead of using `span.update_context` -- which meant if the span was dropped, it would throw a `KeyError`

## Related issues

Closes #1689

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,18 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 6.x" heading
+//[float]
+//===== Features
+//
+[float]
+===== Bug fixes
+
+* Fix error in Elasticsearch instrumentation when spans are dropped {pull}1690[#1690]
+
 [[release-notes-6.x]]
 === Python Agent version 6.x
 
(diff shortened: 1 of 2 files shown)


beniwohli (approved):
LGTM!

PR #1692, opened 2022-11-14
Lower log level for server version problems

basepi:

Currently our server version checks are fairly low stakes. If there are issues fetching the server version, we don't need to make them `warning` level. This sets them to `debug`.

I also added the request body to the JSON error case, as we ran into one of those recently and it was hard to debug since we didn't have the body in the logs.


basepi:
As an aside, as we add more server version checks in the future we may need to re-evaluate this change. The current set of server version checks are pretty low-stakes, but that may not always be the case.

beniwohli (approved):
LGTM, pending updates to log levels in `tests/transports/test_urllib3.py` :D

PR #1709, opened 2022-12-12
[WIP] Wrapper script

### LINKED ISSUE #1019: Wrapper script for no-code-changes onboarding

### LINKED ISSUE #1598: No-Code-Changes Instrumentation: Flask
Write an instrumentation for Flask to for use with the no-code-changes wrapper script + sitecustomize.py

[Meta](https://github.com/elastic/apm-agent-python/issues/1600)

basepi:

Adds the wrapper script to automate instrumentation without code changes.

Also adds support for no-code-changes instrumentation in Flask, plus a bunch of test fixes around that.

## Related issues

Closes #1019
Closes #1598


setup.cfg
```diff
@@ -163,6 +168,7 @@ markers =
     aiobotocore
     kafka
     grpc
+addopts=--random-order
```
basepi, line 171: I added this while debugging the test failures on the old flask tests. I think we should leave it on, to try to prevent adding any tests that rely on order.

tests/client/transaction_tests.py
```diff
@@ -445,7 +445,7 @@ def test_backdating_transaction(elasticapm_client):
     elasticapm_client.begin_transaction("test", start=time.time() - 1)
     elasticapm_client.end_transaction()
     transaction = elasticapm_client.events[constants.TRANSACTION][0]
-    assert 1000 < transaction["duration"] < 2000
+    assert 900 < transaction["duration"] < 2000
```
basepi, line 448: This was a windows failure:

```
[2023-01-11T19:42:05.004Z] elasticapm_client = <tests.fixtures.TempStoreClient object at 0x0000018EBF6B4188>
[2023-01-11T19:42:05.004Z] 
[2023-01-11T19:42:05.004Z]     def test_backdating_transaction(elasticapm_client):
[2023-01-11T19:42:05.004Z]         elasticapm_client.begin_transaction("test", start=time.time() - 1)
[2023-01-11T19:42:05.004Z]         elasticapm_client.end_transaction()
[2023-01-11T19:42:05.004Z]         transaction = elasticapm_client.events[constants.TRANSACTION][0]
[2023-01-11T19:42:05.004Z] >       assert 1000 < transaction["duration"] < 2000
[2023-01-11T19:42:05.004Z] E       assert 1000 < 999.971
[2023-01-11T19:42:05.004Z] 
[2023-01-11T19:42:05.004Z] tests\client\transaction_tests.py:448: AssertionError
```

I *assume* it's a side effect of `--random-order`....or maybe it's a clock issue on the VM? I don't know. I just added some wiggle room.

docs/wrapper.asciidoc
```diff
@@ -0,0 +1,55 @@
...
+----
+$ elasticapm-run --version
+elasticapm-run 6.14.0
+----
+
+Alternatively, you can run the entrypoint directly:
+
+[source,bash]
+----
+$ python -m elasticapm.instrumentation.wrapper.__init__ --version
```
beniwohli, line 31: Maybe we can get rid of the somewhat ugly `__init__`. From the docs of the `-m` flag:

> Package names (including namespace packages) are also permitted. When a package name is supplied instead of a normal module, the interpreter will execute `<pkg>.__main__` as the main module. This behaviour is deliberately similar to the handling of directories and zipfiles that are passed to the interpreter as the script argument.

So could we add a `__main__` module that calls `setup()`...?

(I didn't try this)

beniwohli (commented):
Great work!

elasticapm/instrumentation/register.py
```diff
@@ -69,6 +69,7 @@
     "elasticapm.instrumentation.packages.kafka.KafkaInstrumentation",
     "elasticapm.instrumentation.packages.grpc.GRPCClientInstrumentation",
     "elasticapm.instrumentation.packages.grpc.GRPCServerInstrumentation",
+    "elasticapm.instrumentation.packages.flask.FlaskInstrumentation",
```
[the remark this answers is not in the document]
basepi: Seems like a reasonable solution.

I did think about this, and it did cause some of my annoying testing issues. My thought was that to run into any issue without the wrapper script, they would have to create a `Client` and call `instrument()` before creating a Flask object. Which would be pretty far from our documented usage.

Even late-initializing with [`init_app()`](https://github.com/elastic/apm-agent-python/blob/43b3554a63a581961e35bab281285f8464c53765/elasticapm/contrib/flask/__init__.py#L117) should work fine because we don't call `instrument()` until we `init_app()`.

So I think this PR is fine as written. But it may be worth splitting them out now to prevent issues with other frameworks, as well as to make testing easier. I'll look into this.

elasticapm/instrumentation/register.py
```diff
@@ -94,11 +94,21 @@
...
+_wrapper_register = {
+    "elasticapm.instrumentation.packages.flask.FlaskInstrumentation",
+}
+
 
 def register(cls):
     _cls_register.add(cls)
 
 
+def register_wrapper():
```
beniwohli, line 108: I would give this function another name, e.g. `register_wrapper_instrumentations`, to make it a bit clearer what it does.

beniwohli (approved):
Just one little nitpick, otherwise 👍 . Great work!

PR #1727, opened 2023-01-25
Fix lambda performance issue

basepi:

I still don't know why, but moving the `Client` creation inside of the function execution resulted in ~18x the billable execution time of a basic "Hello World" lambda function. Instrumentation was wayyy slower.

This PR moves the instrumentation back out to setup time, and deprecates the use of kwargs with `capture_serverless`.

The core performance fix is in the first commit: https://github.com/elastic/apm-agent-python/pull/1727/commits/ccb5b67b805008c54edafe869be78773be437941

Also implements minor spec changes from https://github.com/elastic/apm/pull/749

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,6 +42,7 @@ endif::[]
 ===== Bug fixes
 
 * Fix an async issue with long elasticsearch queries {pull}1725[#1725]
+* Fix a major performance issue with AWS Lambda Support {pull}1727[#1727]
 
 
 [[release-notes-6.x]]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,7 +42,7 @@ endif::[]
 ===== Bug fixes
 
 * Fix an async issue with long elasticsearch queries {pull}1725[#1725]
-* Fix a major performance issue with AWS Lambda Support {pull}1727[#1727]
+* Fix a performance issue with AWS Lambda Support {pull}1727[#1727]
 
 
 [[release-notes-6.x]]
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -42,7 +42,8 @@ endif::[]
 ===== Bug fixes
 
 * Fix an async issue with long elasticsearch queries {pull}1725[#1725]
-* Fix a performance issue with AWS Lambda Support {pull}1727[#1727]
+* Fix a performance issue with our AWS Lambda integration {pull}1727[#1727]
+* Mark `**kwargs` config usage in our AWS Lambda integration as deprecated {pull}1727[#1727]
 
 
 [[release-notes-6.x]]
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -57,6 +57,17 @@ _AWSLambdaContextT = TypeVar("_AWSLambdaContextT")
 
 
 def capture_serverless(func: Optional[callable] = None, **kwargs) -> callable:
+    """
+    Decorator for instrumenting AWS Lambda functions.
+
+    Example usage:
+
+        from elasticapm import capture_serverless
+
+        @capture_serverless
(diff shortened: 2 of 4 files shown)


elasticapm/contrib/serverless/aws.py
```diff
@@ -50,11 +51,65 @@
...
+    if "service_version" not in kwargs and "ELASTIC_APM_SERVICE_VERSION" not in os.environ:
+        kwargs["service_version"] = os.environ.get("AWS_LAMBDA_FUNCTION_VERSION")
+
+    global INSTRUMENTED
+    client = get_client()
+    if not client:
+        client = Client(**kwargs)
+    if not client.config.debug and client.config.instrument and client.config.enabled and not INSTRUMENTED:
+        elasticapm.instrument()
+        INSTRUMENTED = True
```
basepi, line 94: Technically this `INSTRUMENTED` flag is not required. Instrumentation is idempotent, so multiple calls don't matter. I added it purely for performance reasons.

CHANGELOG.asciidoc
```diff
@@ -42,6 +42,7 @@ endif::[]
 ===== Bug fixes
 
 * Fix an async issue with long elasticsearch queries {pull}1725[#1725]
+* Fix a performance issue with AWS Lambda Support {pull}1727[#1727]
```
trentm, line 45: I'm not sure user's will notice, but should the change deprecating[^1] use of kwargs to `capture_serverless` be called out here.

[^1]: Or perhaps not "deprecated" yet? The `PendingDeprecationWarning` below says it'll be deprecated later... as opposed to deprecated now and *remove* in a subsequent major.

elasticapm/contrib/serverless/aws.py
```diff
@@ -50,11 +51,67 @@
...
-class capture_serverless(object):
+def capture_serverless(func: Optional[callable] = None, **kwargs) -> callable:
+    if not func:
+        # This allows for `@capture_serverless()` in addition to
+        # `@capture_serverless` decorator usage
+        return functools.partial(capture_serverless, **kwargs)
+
+    if kwargs:
+        warnings.warn(
+            PendingDeprecationWarning(
```
basepi, line 67: Open question as to whether this needs to wait 18 months. I don't think anyone was using this with kwargs, it makes *way* more sense to use env variables in lambda environments, but it technically will be a breaking change. 🤷‍♂️

trentm (commented):
I don't know any of this well-enough to do a decent review. I agree it would be good to get the fix out, even if not fully understood.

CHANGELOG.asciidoc
basepi: We use `DeprecationWarning` when support is removed, and `PendingDeprecationWarning` in the period leading up to its removal.

elasticapm/contrib/serverless/aws.py
```diff
@@ -50,11 +51,67 @@
...
+    def decorated(*args, **kwds):
+        if len(args) == 2:
+            # Saving these for request context later
+            event, context = args
+        else:
+            event, context = {}, {}
+
+        if not client.config.debug and client.config.instrument and client.config.enabled:
+            with capture_serverless_context(func, name, client, event, context) as sls:
+                sls.response = func(*args, **kwds)
```
beniwohli, line 106: I wonder if it would make sense to move the code from `__exit__()` that accesses `response` to here, which would avoid us having to store the response on the `sls` object.

beniwohli (approved):
Great job!

Do you think a similar pattern would also make sense for the `capture_span` decorator?

elasticapm/contrib/serverless/aws.py
basepi: I think I'm going to leave it. While I agree that it makes the context manager slightly more inscrutable to use, the reality is that the context manager is not designed to be used anywhere else, and I love how clean the `decorated()` function is.

I considered actually not using a context manager at all -- it's not really needed, as it's only used in one place. But I think the readability that it adds is worth it.

PR #1728, opened 2023-01-25
fix: "tracestate" parsing must allow whitespace around list-members

trentm:
Per https://w3c.github.io/trace-context/#tracestate-header-field-values
the tracestate header can have whitespace around list-members. E.g.
    'vendor1=foo, vendor2=bar ,\t vendor3=baz '

Refs: https://github.com/elastic/apm-agent-rum-js/issues/1334

---

I'm fairly confident on the regex changes, but not at all on possible wider impact. I haven't followed closely how `_parse_tracestate` is used in the agent code.

### DIFF
--- a/elasticapm/utils/disttracing.py
+++ b/elasticapm/utils/disttracing.py
@@ -234,8 +234,9 @@ class TraceParent(object):
             return elastic_state
         else:
             # Remove es=<stuff> from the tracestate, and add the new es state to the end
-            otherstate = re.sub(r"(?:,|^)es=([^,]*)", "", self.tracestate)
-            otherstate = otherstate.lstrip(",")
+            otherstate = re.sub(r"(?:,|^)\s*es=([^,]*)\s*(?:,|$)", "", self.tracestate)
+            otherstate = otherstate.lstrip(",")  # in case `es=` was the first entry
+            otherstate = re.sub(r",,", ",", otherstate)  # remove potential double commas
             # No validation of `otherstate` required, since we're downstream. We only need to check `es=`
             # since we introduced it, and that validation has already been done at this point.
             if otherstate:
--- a/elasticapm/utils/disttracing.py
+++ b/elasticapm/utils/disttracing.py
@@ -212,7 +212,7 @@ class TraceParent(object):
 
         ret = {}
         try:
-            state = re.search(r"(?:,|^)\s*es=([^,]*?)\s*(?:,|$)", tracestate).group(1).split(";")
+            state = re.search(r"(?:,|^)\s*es=([^,]*)\s*(?:,|$)", tracestate).group(1).split(";")
         except IndexError:
             return {}
         for keyval in state:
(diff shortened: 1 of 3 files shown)


elasticapm/utils/disttracing.py
```diff
@@ -208,7 +212,7 @@ def _parse_tracestate(self, tracestate) -> Dict[str, str]:
 
         ret = {}
         try:
-            state = re.search(r"(?:,|^)es=([^,]*)", tracestate).group(1).split(";")
+            state = re.search(r"(?:,|^)\s*es=([^,]*?)\s*(?:,|$)", tracestate).group(1).split(";")
```
basepi, line 215: Is this question mark superfluous? The `*` already makes it optional, yes?

```suggestion
            state = re.search(r"(?:,|^)\s*es=([^,]*)\s*(?:,|$)", tracestate).group(1).split(";")
```

basepi (approved):

basepi:
@trentm it won't let me ask you for review since you're the author...what do you think of my changes?

elasticapm/utils/disttracing.py
trentm: That `?` is for a non-greedy match, otherwise, I think (I'd need to test), the `([^,]*)` could slurp up the trailing whitespace rather than leaving it to the `\s*`.

trentm:
```
>>> tracestate = 'foo=bar, es=blah , bing=bang '
>>> re.search(r"(?:,|^)\s*es=([^,]*)\s*(?:,|$)", tracestate).group(1)
'blah '
>>> re.search(r"(?:,|^)\s*es=([^,]*?)\s*(?:,|$)", tracestate).group(1)
'blah'
```

basepi: Ah, right, thank you. In that case I can undo my "fix" in https://github.com/elastic/apm-agent-python/pull/1728/commits/a609b1a52df7dded2eb7a27720ef223daa875f54

trentm: Ah, I hadn't see the `.strip()` call. I think the issue with that if this were for **general** tracestate parsing, is that a `value` here can have *leading* whitespace (per https://www.w3.org/TR/trace-context/#value). So it would have to be a `.rstrip()` if kept.   I'm nitpicking though, because IIUC the Python impl *only* ever processing the `es` vendor value and we'll never care about leading whitespace.

trentm: A possible alternative for parsing this all, if the regex is deemed getting too complex, would be to:
- `str.find('es=')`
- if there was a hit, then `.split(',')` and `.trim()` each entry
- then pick the one(s) starting with 'es='.

trentm:
@basepi Should I merge this? Or should  there be a changelog entry or something?
I'm away tomorrow, so feel free to take this one over.

elasticapm/utils/disttracing.py
basepi: I think it's fine. But thanks for the thoughtful suggestions.

PR #1733, opened 2023-01-31
Fix some flaky tests

basepi:

I've noticed two tests that have been more flaky than usual recently, presumably due to `--random-order`:

- test_label_while_no_transaction
  - The real problem here is `test_get_transaction` was not cleaning up the execution context
- test_memcache_span_compression
  - This one is less clear. But I'm guessing that since we were testing with `5ms` span compression, performance blips on the docker containers could result in a failure. I just bumped the span compression threshold up.

I've also seen `test_flushed_arg_with_wait` fail recently but I was never able to reproduce locally.

The above tests should be less flaky now.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -168,7 +168,7 @@ markers =
     aiobotocore
     kafka
     grpc
-addopts=--random-order
+addopts=--random-order -v
 
 [isort]
 line_length=120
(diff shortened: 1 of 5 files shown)


beniwohli (approved):
Good finds!

PR #1745, opened 2023-02-08
Add update-specs updatcli workflow

reakaleek:

Moves the update specs workflow from apm-server to the consumer using updatecli to have a "dependabot"-like experience.

## How to test

```
GITHUB_TOKEN=<token>\
GIT_USER=<username> \
GIT_EMAIL=j<email> \
updatecli diff -c .ci/update-specs.yml
```

## Related issues

- https://github.com/elastic/observability-robots/issues/1516

### DIFF
--- a/.ci/update-specs.yml
+++ b/.ci/update-specs.yml
@@ -69,7 +69,7 @@ targets:
     sourceid: error.json
     kind: file
     spec:
-      file: upstream/json-specs/error.json
+      file: tests/upstream/json-specs/error.json
       forcecreate: true
   metadata.json:
     name: metadata.json
@@ -77,7 +77,7 @@ targets:
     sourceid: metadata.json
     kind: file
     spec:
-      file: upstream/json-specs/metadata.json
+      file: tests/upstream/json-specs/metadata.json
       forcecreate: true
   metricset.json:
     name: metricset.json
@@ -85,7 +85,7 @@ targets:
     sourceid: metricset.json
     kind: file
     spec:
-      file: upstream/json-specs/metricset.json
+      file: tests/upstream/json-specs/metricset.json
       forcecreate: true
   span.json:
     name: span.json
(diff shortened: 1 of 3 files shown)


.ci/update-specs.yml
```diff
@@ -0,0 +1,105 @@
...
+    spec:
+      file: https://raw.githubusercontent.com/elastic/apm-data/main/input/elasticapm/docs/spec/v2/metricset.json
+  span.json:
+    kind: file
+    spec:
+      file: https://raw.githubusercontent.com/elastic/apm-data/main/input/elasticapm/docs/spec/v2/span.json
+  transaction.json:
+    kind: file
+    spec:
+      file: https://raw.githubusercontent.com/elastic/apm-data/main/input/elasticapm/docs/spec/v2/transaction.json
```
reakaleek, line 42: @axw 
Is it okay for this to be a "static list" of files, or do we need something more dynamic here that considers file additions/deletions?

v1v (approved):

basepi (approved):

.ci/update-specs.yml
axw: What are our other options? The set of files won't change often, but they might still change. I think we could go ahead with this, and look at bundling the schema documents into a Compound Schema Document: http://json-schema.org/understanding-json-schema/structuring.html#bundling

axw (approved):

.ci/update-specs.yml
axw: Opened https://github.com/elastic/apm-data/issues/15 to improve this

reakaleek: Another option could be a shell script that does the cloning and copying of the files. Which feels a bit hacky in the updatecli way, but is certainly possible.

Regardless,  a compound schema sounds really good! This will also account for deletions.

I actually encountered a case where `log.json` was not removed (Also in the current implementation). (See https://github.com/elastic/apm-agent-nodejs/pull/3153#pullrequestreview-1289704768) A single file would fix this behaviour.

PR #1749, opened 2023-02-09
Required Status Check

reakaleek:

Creates the status check `test / all` which can be set as required status check.

This requires the matrix items to fail in case of failing tests.

## Related issues
- https://github.com/elastic/observability-robots/issues/1571

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -139,5 +139,4 @@ jobs:
       - chunks-3
       - windows
     steps:
-      - run: |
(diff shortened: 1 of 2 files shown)


reakaleek:
@basepi 

I noticed that windows checks do not return a failing exit code. not sure if this is also true for the docker runs.

Can you help me accomplish this?

basepi (approved):

beniwohli (approved):

PR #1753, opened 2023-02-13
Set status check to success in case it's an only-docs PR

reakaleek:

Sets the `test / all` status check to green in case it's a only-docs PR and the test.yml workflow is not triggered.

This is needed to be able to set `test / all` as required status check.

## Related issues
- #1749

### DIFF
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -5,9 +5,10 @@ name: test
 
 on:
   pull_request:
-    paths:
-      - '**/*.md'
-      - '**/*.asciidoc'
+    paths-ignore: # This expression needs to match the paths ignored on test.yml.
+      - '**'
+      - '!**/*.md'
+      - '!**/*.asciidoc'
 
 permissions:
   contents: read
--- a/.github/workflows/test-docs.yml
+++ b/.github/workflows/test-docs.yml
@@ -1,0 +1,16 @@
+---
+# This workflow sets the test / all status check to success in case it's a docs only PR and test.yml is not triggered
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: test
+
+on:
+  pull_request:
+    paths:
+      - '**/*.md'
+      - '**/*.asciidoc'
+
+jobs:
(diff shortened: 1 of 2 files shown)


.github/workflows/test-docs.yml
```diff
@@ -0,0 +1,19 @@
+---
+# This workflow sets the test / all status check to success in case it's a docs only PR and test.yml is not triggered
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: test
+
+on:
+  pull_request:
+    paths:
```
v1v, line 8: Will this particular workflow run even if changes contain other files that are not `.md` or `.asciidoc`? 

I've seen this pattern with `paths-ignore`:
```
    paths-ignore: # This expression needs to match the paths ignored on ci.yml.
      - '**' # Ignore all file changes, except for the below.
      - '!**.md'
      - '!**.asciidoc'
      - ```

Comment (reakaleek) [reply .github/workflows/test-docs.yml:8]: Ah, yes. Thank you for the hint. I was following the github docs and there it's kind of reversed.. which is why it works in their usecase.

After thinking about it.. your suggestion is necessary, so it won't trigger in case other files are edited as well.

Comment (reakaleek) [reply .github/workflows/test-docs.yml]: https://github.com/elastic/apm-agent-python/pull/1753/commits/33867e5ca2f606c6e6c672bc02808a40f2e79247

**Review by v1v — APPROVED** [review]
LGMT, just a nit comment

File: .github/workflows/test-docs.yml [hunk 2]
```diff
@@ -0,0 +1,20 @@
+---
+# This workflow sets the test / all status check to success in case it's a docs only PR and test.yml is not triggered
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: test
+
+on:
+  pull_request:
+    paths-ignore: # This expression needs to match the paths ignored on test.yml.
```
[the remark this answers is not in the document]
reakaleek: https://github.com/elastic/apm-agent-python/pull/1753/commits/68a1953f159ee907f9026081c9e87994940164da

Issue #1739, opened 2023-02-03
[GRPC] Handle the null span case


sublime1809:
**Describe the bug**: ...

**To Reproduce**

1. Setup ElasticAPM in Django
2. Integrate with Google PubSub client
3. Try to publish message from your script
4. Stacktrace:
```
Exception in thread Thread-CommitBatchPublisher:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.8/site-packages/google/cloud/pubsub_v1/publisher/_batch/thread.py", line 274, in _commit
    response = self._client._gapic_publish(
  File "/usr/local/lib/python3.8/site-packages/google/cloud/pubsub_v1/publisher/client.py", line 272, in _gapic_publish
    return super().publish(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/google/pubsub_v1/services/publisher/client.py", line 781, in publish
    response = rpc(
  File "/usr/local/lib/python3.8/site-packages/google/api_core/gapic_v1/method.py", line 113, in __call__
    return wrapped_func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/google/api_core/retry.py", line 349, in retry_wrapped_func
    return retry_target(
  File "/usr/local/lib/python3.8/site-packages/google/api_core/retry.py", line 191, in retry_target
    return target()
  File "/usr/local/lib/python3.8/site-packages/google/api_core/timeout.py", line 120, in func_with_timeout
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/google/api_core/grpc_helpers.py", line 72, in error_remapped_callable
    return callable_(*args, **kwargs)
  File "/usr/local/lib/python3.8/site-packages/grpc/_interceptor.py", line 247, in __call__
    response, ignored_call = self._with_call(request,
  File "/usr/local/lib/python3.8/site-packages/grpc/_interceptor.py", line 287, in _with_call
    call = self._interceptor.intercept_unary_unary(continuation,
  File "/usr/local/lib/python3.8/site-packages/elasticapm/contrib/grpc/client_interceptor.py", line 97, in intercept_unary_unary
    client_call_details = self.attach_traceparent(client_call_details, span)
[...]

PR #1761, opened 2023-02-23
[GRPC] Handle non-existent transaction in server interceptor

ckoehn:

This change handles the case when APM is not recording/disabled and `client.begin_transaction()` just returns `None`. I wasn't able to figure out how to add a test for this. Do you have an advice/hint?

## Related issues

* https://github.com/elastic/apm-agent-python/issues/1739
* Introduced in the initial gRPC instrumentation (https://github.com/elastic/apm-agent-python/pull/1703)

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,19 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 6.x" heading
+//[float]
+//===== Features
+
+[float]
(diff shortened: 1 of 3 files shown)


beniwohli (approved):
Hi @ckoehn 

Good catch, the fix looks good. Testing this is indeed somewhat tricky, as our setup with a GRPC server in a subprocess is a bit convoluted. I added a test that does the trick, albeit a bit unsatisfactory. It fails without your patch, but will also pass if a transaction isn't recorded by our fake APM Server for other reasons. But I think it's acceptable for now.

beniwohli:
huh. not sure why this was auto-merged before the tests completed :/

PR #1765, opened 2023-02-28
Fix default botocore handler

### LINKED ISSUE #1764: botocore instrumentation creating empty services
When using `boto3` to call out to services outside of the specially handled services (S3, SQS, DynamoDB, SNS), the `botocore` instrumentation creates spans with an incorrect `service.name`. This results in extra, empty services in both the service list and the service map.

basepi:

The `handle_default` handler was not correctly nesting its destination context, which resulted in spans with an incorrect `service.name`. This meant users would see empty services in their service list (and service map) for each AWS service type, but with no transactions attached to those services. 

## Related issues

Ref #1054 
Closes #1764

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,19 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
+// When the next release happens, nest these changes under the "Python Agent version 6.x" heading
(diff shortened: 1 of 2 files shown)


beniwohli (approved):
good catch!

beniwohli:
@basepi heads up, looks like you made this PR from your local `main` branch, so there's probably some cleaning up necessary on your side.

PR #1767, opened 2023-03-06
action: test-docs workflow should create basic junit report

v1v:
## What does this pull request do?

Generate junit test report file with a skipped entry for the `test-docs`

## Why 

otherwise, the test-reporter will fail as no artifacts can be found

## Issue

https://github.com/elastic/apm-pipeline-library/pull/2121 is the one testing this in isolation.

image

See https://github.com/elastic/apm-pipeline-library/runs/11788165082

PR #1768, opened 2023-03-06
chore: test-reporter

v1v:
Just wanna double check whether the fix in https://github.com/elastic/apm-agent-python/pull/1767 works


v1v:
and it works

https://github.com/elastic/apm-agent-python/runs/11792936944

PR #1770, opened 2023-03-07
Add a default span name

basepi:

We list `name` as a required argument in our [`capture_span` docs](https://www.elastic.co/guide/en/apm/agent/python/current/api.html#api-capture-span). However, if a span name is not provided, we silently set it to `None`, and it gets dropped by the APM Server (as it's a [required field](https://github.com/elastic/apm-data/blob/534660e4a1517a6bc2fbc58b1df22e92bcce7180/input/elasticapm/docs/spec/v2/span.json#L876)).

I didn't think this was acceptable behavior, so I copied what the Node agent does, and set the default to `"unnamed"` so we don't silently drop spans.

This also matches my spec change here: https://github.com/elastic/apm/pull/772

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,19 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
+// Unreleased changes go here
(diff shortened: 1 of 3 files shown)


beniwohli (approved):
Looks good!

PR #1775, opened 2023-03-10
Lambda support for `event` formed as list

### LINKED ISSUE #1746: capture_serverless makes lambda fail when input is not a JSON-dict
**Describe the bug**: 

Hi! 

I have encountered a bug when using `@capture_serverless` decorator on a lambda inside a Step Function that processes the output of a Map state.

Elastic APM Python package version: 6.14.0 (latest)

**To Reproduce**

1. Implement a Step Function with a Map state followed by a lambda function that takes the output of the map. As specified in AWS docs: `When complete, the output of the Map state is a JSON array, where each item is the output of an iteration.` - https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html#:~:text=the%20ItemSelector%20field.-,When%20complete%2C%20the%20output%20of%20the%20Map%20state%20is%20a%20JSON%20array%2C%20where%20each%20item%20is%20the%20output%20of%20an%20iteration.,-For%20more%20information 
2. Add the `capture_serverless` setup to your lambda.

Expected: lambda to run and log to APM Server

Actual: Error from Step Function logs:
```
{
    "errorMessage": "'list' object has no attribute 'get'",
    "errorType": "AttributeError",
    "stackTrace": [
      "  File \"/var/task/elasticapm/contrib/serverless/aws.py\", line 116, in decorated\n    with _lambda_transaction(func, name, client, event, context) as sls:\n",
      "  File \"/var/task/elasticapm/contrib/serverless/aws.py\", line 147, in __enter__\n    trace_parent = TraceParent.from_headers(self.event.get(\"headers\", {}))\n"
    ]
  }
```

I think the fix is as simple as updating `elasticapm.xxx.aws`  `_lambda_transaction` to accept `event` being either a dictionary or a list, as this is an acceptable input to lambdas within Step Functions.

- OS: [e.g. Linux] - AWS Lambda function for Python
- Python version: 3.8
- APM Server version: 6.14.0

**Severity**
- Critical issue as this breaks all our lambdas processing events coming from Map states in Step Functions.

basepi:

Lambda functions can be invoked by other AWS services, such as Step Functions. In this case, the `event` will be the JSON object received from the invoking service. This means that `event` could be a list, instead of a dict, which would cause an exception to be raised. This fixes that issue, and is also a useful test of our `event` defaults.

I manually tested this with Step Functions in AWS, and also wrote a unit test.

## Related issues

Closes #1746

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -40,6 +40,7 @@ endif::[]
 ===== Bug fixes
 
 * Fix spans being dropped if they don't have a name {pull}1770[#1770]
+* Fix AWS Lambda support when `event` is a list {pull}1775[#1775]
 
 
 
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -144,6 +144,11 @@ class _lambda_transaction(object):
         """
         Transaction setup
         """
+        if isinstance(self.event, list):
+            # When `event` is a list, it's likely the output of another AWS
+            # service like Step Functions, and is unlikely to be standardized
+            # in any way. We just have to rely on our defaults in this case.
+            self.event = {}
         trace_parent = TraceParent.from_headers(self.event.get("headers", {}))
 
         global COLD_START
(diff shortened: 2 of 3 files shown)


fjoanr:
Looks good to me!

beniwohli (approved):
Looks good!

PR #1780, opened 2023-03-21
Add wrapper instrumentation for Django

### LINKED ISSUE #1597: No-Code-Changes Instrumentation: Django
Write an instrumentation for Django to for use with the no-code-changes wrapper script + sitecustomize.py

[Meta](https://github.com/elastic/apm-agent-python/issues/1600)

beniwohli:
We instrument `Apps.populate` to insert our own app at the top of the list. This takes care of initialization.

As Django requires a client of type DjangoClient, we can't initialize the client in sitecustomize already. Hence, we only load a limited config object to check enabled/instrument config values, and move client initialization into the actual instrumentation.

## Related issues

Closes #1597
Ref #1709


basepi (approved):
This is great! Glad it was so easy.

As for testing, I'm a bit torn. I think right now this instrumentation is simple enough that we can probably skip an end to end test. But if for some reason this gets more complicated or brittle in the future, I think an end to end test (similar to the one I do for the wrapper script, where we actually shell out) would be worthwhile.

PR #1784, opened 2023-03-27
Add support for partial transaction reporting in Lambda

### LINKED ISSUE #1734: [META 753] Support Partial Transaction for AWS Lambda
See meta issue for the description and details:
- Meta issue: https://github.com/elastic/apm/issues/753

basepi:

The lambda extension now has support for early reporting of partial transactions. This covers the case where our lambda runtime times out before we can properly end and report the transaction. In this case, the lambda extension will report the transaction with a `timeout` result.

## Related issues

Closes #1734

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -385,7 +385,8 @@ class Transaction(BaseSpan):
         return self.trace_parent.span_id
 
     def to_dict(self) -> dict:
-        self.context["tags"] = self.labels
+        context = self.context.copy()
+        context["tags"] = self.labels
         result = {
             "id": self.id,
             "trace_id": self.trace_parent.trace_id,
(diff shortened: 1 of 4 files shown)


elasticapm/transport/base.py
```diff
@@ -273,8 +273,7 @@ def _flush(self, buffer, forced_flush=False):
             fileobj = buffer.fileobj  # get a reference to the fileobj before closing the gzip file
             buffer.close()
 
-            # StringIO on Python 2 does not have getbuffer, so we need to fall back to getvalue
-            data = fileobj.getbuffer() if hasattr(fileobj, "getbuffer") else fileobj.getvalue()
+            data = fileobj.getbuffer()
```
beniwohli, line 276: Good catch!

elasticapm/traces.py
```diff
@@ -385,13 +385,14 @@ def ensure_parent_id(self) -> str:
         return self.trace_parent.span_id
 
     def to_dict(self) -> dict:
-        self.context["tags"] = self.labels
+        context = self.context.copy()
```
beniwohli, line 388: `copy()` is shallow, which is fine as long as we only operate on the top level dictionary (which, if I didn't overlook something, we currently do). Maybe still worth a note here to change to `deepcopy()` if we ever have to modify data deeper in the context object.

elasticapm/transport/http.py
```diff
@@ -70,19 +70,22 @@ def __init__(self, url: str, *args, **kwargs) -> None:
         self._http = None
         self._url = url
 
-    def send(self, data, forced_flush=False):
+    def send(self, data, forced_flush=False, custom_url=None, override_headers=None):
```
beniwohli, line 73: Is there a reason to use `custom` for the URL override, and `override` for the custom headers? :D

beniwohli (approved):
Just some minor comments. Great work!

elasticapm/transport/http.py
basepi: Haha, good catch. It went through a lot of iterations over the PR. I'll clean them up.

PR #1785, opened 2023-03-29
Fix flakey pylons test

basepi:

Somewhere in the test suite, we're leaking a `Client` object. I haven't been able to reproduce this, but this Pylons test in particular relies on a new `Client` being created inside of the test, and was failing because of the `get_client()` code.

This test was also leaking a `Client` object, so I fixed that too. 😅

### DIFF
--- a/tests/contrib/pylons/tests.py
+++ b/tests/contrib/pylons/tests.py
@@ -40,7 +40,7 @@ def example_app(environ, start_response):
     raise ValueError("hello world")
 
 
-def test_init():
+def test_init(close_client):
     config = {
         "elasticapm.server_url": "http://localhost/api/store",
         "elasticapm.service_name": "p" * 32,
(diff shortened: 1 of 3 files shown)


tests/contrib/pylons/tests.py
```diff
@@ -51,3 +52,4 @@ def test_init():
     assert middleware.client.config.service_name == "p" * 32
     assert middleware.client.config.secret_token == "a" * 32
     assert middleware.client.config.metrics_interval.total_seconds() == 0
+    elasticapm.get_client().close()
```
beniwohli, line 55: If we want to make sure that the client gets closed even when the test fails, we'll have to wrap the test in a `try/finally`, and put the `close()` call into the `finally` block.

beniwohli (approved):
Looks good. Alternatively, we could just use https://docs.python.org/3/library/contextlib.html#contextlib.closing, but this is fine too :)

beniwohli:
Right, didn't consider that you can't call `get_client()` before yielding. This solution is great!

PR #1792, opened 2023-04-04
Optimize CI time

reakaleek:

~Adds the `--cache-from` option.~
~This way the build time will be shorter. If there are changes in the Dockerfile, it will still be built utilising the cache.~
~Will only work after https://github.com/elastic/apm-pipeline-library/pull/2174 is merged.~

Use daily built public docker image instead of building on every run.

## Related issues

- depends on https://github.com/elastic/apm-pipeline-library/pull/2174

### DIFF
--- a/tests/scripts/docker/run_tests.sh
+++ b/tests/scripts/docker/run_tests.sh
@@ -47,12 +47,14 @@ fi
 
 # CASS_DRIVER_NO_EXTENSIONS is set so we don't build the Cassandra C-extensions,
 # as this can take several minutes
+
 DOCKER_BUILDKIT=1 docker build \
+  --progress=plain \
   --cache-from="elasticobservability/apm-agent-python-testing:${1}" \
-  --build-arg BUILDKIT_INLINE_CACHE=1 \
   --build-arg PYTHON_IMAGE="${1/-/:}" \
-  -t "apm-agent-python:${1}" \
+  --tag "apm-agent-python-testing:${1}"
   .
+
 PYTHON_VERSION=${1} docker-compose run \
   -e PYTHON_FULL_VERSION=${1} \
   -e LOCAL_USER_ID=$LOCAL_USER_ID \
--- a/tests/scripts/docker/run_tests.sh
+++ b/tests/scripts/docker/run_tests.sh
@@ -48,12 +48,14 @@ fi
 # CASS_DRIVER_NO_EXTENSIONS is set so we don't build the Cassandra C-extensions,
 # as this can take several minutes
 
-DOCKER_BUILDKIT=1 docker buildx build \
-  --progress=plain \
-  --cache-from="elasticobservability/apm-agent-python-testing:${1}" \
-  --build-arg PYTHON_IMAGE="${1/-/:}" \
-  --tag "apm-agent-python:${1}" \
-  .
+if ! ${CI}; then
+  DOCKER_BUILDKIT=1 docker build \
+    --progress=plain \
+    --cache-from="elasticobservability/apm-agent-python-testing:${1}" \
+    --build-arg PYTHON_IMAGE="${1/-/:}" \
+    --tag "apm-agent-python:${1}" \
+    .
+fi
 
 PYTHON_VERSION=${1} docker-compose run \
   -e PYTHON_FULL_VERSION=${1} \
--- a/tests/scripts/docker/run_tests.sh
+++ b/tests/scripts/docker/run_tests.sh
@@ -47,7 +47,7 @@ fi
 
 # CASS_DRIVER_NO_EXTENSIONS is set so we don't build the Cassandra C-extensions,
 # as this can take several minutes
-DOCKER_BUILDKIT=1 docker build --cache-from=elasticobservability/apm-agent-python:${1} --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
+DOCKER_BUILDKIT=1 docker build --cache-from=elasticobservability/apm-agent-python-testing:${1} --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
 PYTHON_VERSION=${1} docker-compose run \
   -e PYTHON_FULL_VERSION=${1} \
   -e LOCAL_USER_ID=$LOCAL_USER_ID \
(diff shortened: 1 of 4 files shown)


basepi:
How will this work for PRs? Is this only for nightly tests?

I think what @beniwohli was requesting in https://github.com/elastic/observability-robots/issues/1623 was that we still create the docker image on each test run, but we do it up front and use the same image for each of the matrix runs (as we currently build it in every matrix run individually). But I don't think https://github.com/elastic/apm-pipeline-library/pull/2174 is creating images for every commit pushed to a PR, so I'm not sure it solves the problem.

But I could be missing something! Thanks for your work on this!

beniwohli:
@basepi these base images don't really change from PR to PR. Updating them daily or even weekly is fine (to pick up changes in the underlying Python base images). We still execute a `pip install -r reqs-base.txt` on every run, so if a PR adds a new test dependency, it'll be installed.

reakaleek:
@basepi 

I'm with you here. This setup is not the most ideal. 

My preferred setup would be utilising the [--cache-from](https://docs.docker.com/engine/reference/commandline/build/#cache-from) option of the `docker build` command. This way we would benefit from caching the image and the ability to rebuild the image on every commit if there are changes.

However, we have this limitation that we cannot use secrets in workflows that are triggered on forked PRs.
Hence, we cannot do a docker login and won't be able to push the image.

As @beniwohli mentioned the base image does not change that much. But this setup will surely create discrepancy when you want to test changes on the docker image itself.

Let me experiment a bit more and I will try to figure out if I can somehow GitHub's artifacts API to achieve the desired caching behavior.

reakaleek:
@basepi and @beniwohli 

I think I got something intereseting here in https://github.com/elastic/apm-agent-python/pull/1793 which is inspired by https://depot.dev/blog/docker-layer-caching-in-github-actions

In this [run](https://github.com/elastic/apm-agent-python/actions/runs/4615344598/jobs/8159136799) you can see the building utilizes the docker layer cache and its time is reduced to ~16 seconds.

cc @elastic/observablt-ci

tests/scripts/docker/run_tests.sh
```diff
@@ -47,7 +47,7 @@ fi
 
 # CASS_DRIVER_NO_EXTENSIONS is set so we don't build the Cassandra C-extensions,
 # as this can take several minutes
-docker build --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
+
```
v1v, line 50: Given a PR that changes the `Dockerfile`, how will it be possible to change the PR in the CI? I'm not sure whether that's something we do often.

reakaleek: I just added this a couple of minutes ago.

https://github.com/elastic/apm-agent-python/pull/1792/commits/1cd662e614829d9062c7e560b47eaa02b5864df4

this way we can utilise the cache and still build it on a individual commit.

reakaleek:
Alright, I think I got a good balance now.

We can build the images daily and publish it dockerhub like suggested in https://github.com/elastic/apm-pipeline-library/pull/2174.

Here, we can utilise the `--cache-from` option, which will speed up the build significantly, but you will still be able to make changes to the docker image for individual commits, if necessary.

tests/scripts/docker/run_tests.sh
reakaleek: https://github.com/elastic/apm-agent-python/pull/1792#issuecomment-1496992110

tests/scripts/docker/run_tests.sh
```diff
@@ -47,7 +47,7 @@ fi
 
 # CASS_DRIVER_NO_EXTENSIONS is set so we don't build the Cassandra C-extensions,
 # as this can take several minutes
-docker build --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
+DOCKER_BUILDKIT=1 docker build --cache-from=elasticobservability/apm-agent-python:${1} --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
```
v1v, line 50: `elasticobservability/apm-agent-python-testing` ?

```suggestion
DOCKER_BUILDKIT=1 docker build --cache-from=elasticobservability/apm-agent-python-testing:${1} --build-arg PYTHON_IMAGE=${1/-/:} -t apm-agent-python:${1} . # replace - with : to get the correct docker image
```

beniwohli:
Looking at a bunch of build logs, it looks like we're still spending a lot of time in docker `RUN` commands that should be cacheable, e.g. https://github.com/elastic/apm-agent-python/actions/runs/4617102660/jobs/8162941988?pr=1792#step:3:125. Am I misunderstanding something?

reakaleek:
> Looking at a bunch of build logs, it looks like we're still spending a lot of time in docker `RUN` commands that should be cacheable, e.g. https://github.com/elastic/apm-agent-python/actions/runs/4617102660/jobs/8162941988?pr=1792#step:3:125. Am I misunderstanding something?

Yes. I still need to get following PRs merged to see the result.
- https://github.com/elastic/apm-agent-python/pull/1794
- https://github.com/elastic/apm-pipeline-library/pull/2174

beniwohli (approved):
Ah, gotcha. LGTM, then :)

reakaleek:
In the latest run attempt it's still not utilising the cache as expected.

https://github.com/elastic/apm-agent-python/actions/runs/4617333116?pr=1792

But the images are definitely accessible from https://hub.docker.com/r/elasticobservability/apm-agent-python-testing/tags

Will need to debug a little further.

reakaleek:
Okay, I give up. I tried multiple approaches and multiple combinations of using `--cache-from` without success.
Locally, it was working just fine. It seems like it just won't work in a GH Actions workflow.

Nevertheless, in it's current state the workflow run time decreased from ~10 minutes to ~4 minutes.
However, the ability to create changes in the docker set up for a specific commit is not given.

@basepi @beniwohli let me know what you think and if we can merge it for now.

If changing the docker setup frequently becomes a necessity, let's revisit it in the future.

WDYT?

basepi (approved):
Sounds fine to me. I'll re-request Beni's review and we can get this merged.

beniwohli (approved):
Really great work, thanks for taking this on, @reakaleek!

PR #1798, opened 2023-04-11
trying to sift through our test dependency mess

beniwohli:
## Related issues

Closes #ISSUE


beniwohli:
@basepi I'll defer moving setup.cfg to pyproject.toml to a later PR. That will allow us to only define the top level testing dependencies in pyproject.toml, and generate `reqs-base.txt` with all dependencies and transitive dependencies pinned, using `pip-compile` from the pip-tools repo.

basepi (approved):

PR #1800, opened 2023-04-12
combine coverage data and upload a html report

beniwohli:

So far, we generate coverage data for every step of the build matrix. To be useful, these coverage files have to be combined and reported on.

As an added benefit, we can then delete the large coverage files and only keep the relatively small HTML report.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -140,3 +140,42 @@ jobs:
       - windows
     steps:
       - run: test $(echo '${{ toJSON(needs) }}' | jq -s 'map(.[].result) | all(.=="success")') = 'true'
+
+  coverage:
+    name: Combine & check coverage.
+    needs: all
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -140,3 +140,42 @@ jobs:
...
+
+          # Report again and fail if under 84%.
+          python -Im coverage report --fail-under=84
+
+      - name: Upload HTML report
+        uses: actions/upload-artifact@v3
+        with:
+          name: html-coverage-report
+          path: htmlcov
+      - uses: geekyeggo/delete-artifact@v2
```
kuisathaverat, line 179: For actions that are not in the `actions/*` namespace we usually set an SHA instead of a version to avoid updating to un verified versions.
 
```suggestion
      - uses: geekyeggo/delete-artifact@54ab544f12cdb7b71613a16a2b5a37a9ade990af
```

basepi (approved):
Excellent!

PR #1807, opened 2023-04-19
add instrumentation for redis.asyncio

### LINKED ISSUE #1799: Instrument redis.asyncio
aioredis has been moved into `redis.asyncio`, and aioredis itself is deprecated/dormant.

While moving the code over to redis-py, enough changes where introduced to necessitate a separate instrumentation module.

beniwohli:

aioredis has been discontinued and moved into redis.asyncio, but with enough changes to warrant its own instrumentation.

## Related issues

Closes #1799


basepi (approved):
This looks good. My only question (without having looked at the async redis code) is whether we should be using `AsyncAbstractInstrumentedModule` or `await`ing any of these calls? Outside of the `async_capture_span` this module is not very async. That may be correct, especially with passing integration tests, but I wanted to point it out.

PR #1821, opened 2023-04-28
Update the tests_require block to match reqs-base.txt

basepi:

@sethmlarson ran into issues getting the tests running using the `tests_require` block in setup.cfg. It's way out of date, this PR just matches it to `reqs-base.txt` for now. I even took out the framework extras for simplicity.

At some point we need to figure out how to automate this. Or just drop the test_requires block altogether? If it's out of date, it's useless.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -40,39 +40,6 @@ install_requires =
     certifi
     wrapt>=1.14.1
 test_suite=tests
-tests_require =
-    pytest==7.0.1 ; python_version == '3.6'
-    pytest==7.3.0 ; python_version > '3.6'
-    pytest-random-order==1.1.0
-    pytest-django==4.4.0
-    coverage==6.2 ; python_version == '3.6'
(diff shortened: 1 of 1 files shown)


beniwohli (approved):

beniwohli:
@basepi one way to automate this would be to use `test_requires` as canonical source, and either write some tooling to copy it to `reqs-base.txt`, or use `pip-tools`. Unfortunately, I don't think `pip-compile` supports `tests_require`, so we'd need to do some pre-/post-processing.

basepi:
I planned to update our [CONTRIBUTING.md](https://github.com/elastic/apm-agent-python/blob/main/CONTRIBUTING.md) as well, but it turns out it already just references the `tests/requirements/reqs*.txt` files and doesn't use `setup.py test` anyway, so I just removed the block.

beniwohli (approved):
+0 -59, nice!

PR #1822, opened 2023-05-01
Fix tests for urllib3 v2.0.1+

### LINKED ISSUE #1816: urllib3 v2.0 compatibility
urllib3 v2.0 came out during PyConUS 2023 and is resulting in hanging tests on windows, and test failures on Linux. We need to update our code/tests to ensure compatibility for this new version.

basepi:

urllib3 v2.0+ introduced a change to args ordering that broke some tests. This fixes those tests (plus a stack frame test that was failing under a new version of VSCode's python extension).

The hangs that were breaking test runs for urllib3 v2.0.0 were due to a regression that has been [fixed](https://github.com/urllib3/urllib3/pull/2992) in v2.0.1. Big thanks to @pquentin, @sethmlarson, and @graingert for help with this!

## Related issues

Closes #1816

### DIFF
--- a/elasticapm/instrumentation/packages/urllib3.py
+++ b/elasticapm/instrumentation/packages/urllib3.py
@@ -61,7 +61,12 @@ def update_headers(args, kwargs, instance, transaction, trace_parent):
     :param trace_parent: the TraceParent object
     :return: an (args, kwargs) tuple
     """
-    if len(args) >= 4 and args[3]:
+    from urllib3._version import __version__ as urllib3_version
+
+    if urllib3_version.startswith("2") and len(args) >= 5 and args[4]:
+        headers = args[4].copy()
+        args = tuple(itertools.chain((args[:4]), (headers,), args[5:]))
+    elif len(args) >= 4 and args[3]:
         headers = args[3].copy()
         args = tuple(itertools.chain((args[:3]), (headers,), args[4:]))
     elif "headers" in kwargs and kwargs["headers"]:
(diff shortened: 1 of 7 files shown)


elasticapm/instrumentation/packages/urllib3.py
```diff
@@ -61,7 +61,12 @@ def update_headers(args, kwargs, instance, transaction, trace_parent):
     :param trace_parent: the TraceParent object
     :return: an (args, kwargs) tuple
     """
-    if len(args) >= 4 and args[3]:
+    from urllib3._version import __version__ as urllib3_version
+
+    if urllib3_version.startswith("2") and len(args) >= 5 and args[4]:
+        headers = args[4].copy()
+        args = tuple(itertools.chain((args[:4]), (headers,), args[5:]))
+    elif len(args) >= 4 and args[3]:
```
pquentin, line 69: I was initially surprised to see version dependent code here since urllib3 2.0 is supposed to be mostly compatible with urllib3 1.26. Looking more closely, I understand now that you need to set trace headers irrelevant of the way they were set by the caller (arg, kwarg, instance parameter). Since urllib3 2.0 added a `body` field, it moved `headers` from 4th positional parameter to 5th positional parameter, and you have to detect this. cc @sethmlarson 

Is my understanding right? Should you update the docstring to mention urllib3 2.0?

basepi: Yes, you understood correctly. I probably should throw a comment in there for better documentation.

beniwohli (approved):
Nice work!

PR #1837, opened 2023-05-25
[doc] Align agents logs documentation

SylvainJuge:
update logs docs for python agent, part of https://github.com/elastic/apm/issues/800

### DIFF
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -1,26 +1,28 @@
 [[logs]]
 == Logs
 
-Log correlation allows you to navigate to all logs belonging to a particular trace, and vice-versa -- for a specific log, see in which context it has been logged, and which parameters the user provided.
+Elastic Python APM Agent provides the following log features:
+
+- <<log-correlation-ids>> : Automatically inject correlation IDs that allow navigation between logs, traces and services.
+- <<log-reformatting>> : Automatically reformat plaintext logs in {ecs-logging-ref}/intro.html[ECS logging] format.
+
+Those features are part of {observability-guide}/application-logs.html[Application log ingestion strategies].
+
+The {ecs-logging-python-ref}/intro.html[`ecs-logging-python`] library can also be used to use the ECS logging format without an APM agent.
+When deployed with the Python APM agent, the agent will provide <<log-correlation-ids,log correlation>> IDs.
+
+[float]
+[[log-correlation-ids]]
+=== Log correlation
+
+{apm-guide-ref}/log-correlation.html[Log correlation] allows you to navigate to all logs belonging to a particular trace
+and vice-versa: for a specific log, see in which context it has been logged and which parameters the user provided.
 
 The Agent provides integrations with both the default Python logging library,
 as well as http://www.structlog.org/en/stable/[`structlog`].
 
 * <<logging-integrations>>
 * <<log-correlation-in-es>>
-
-[float]
-[[log-correlation-ids]]
-=== Log correlation
-
-The easiest way to integrate your logs with APM is to use the
-https://github.com/elastic/ecs-logging-python[`ecs-logging`] library, which
-is also provided by Elastic. This library provides formatters for both `logging`
-and `structlog` which create ECS-compatible logs and will include the tracing
-information required for log correlation in kibana. Coupled with something like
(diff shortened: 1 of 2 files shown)


SylvainJuge:
@elasticmachine run elasticsearch-ci/docs

docs/logging.asciidoc
```diff
@@ -174,3 +176,15 @@ Then, you could use a grok pattern like this (for the
   ]
 }
 ----
+
+[float]
+[[log-reformatting]]
+=== Log reformatting (experimental)
+
+The agent can automatically reformat application logs to ECS format, without adding a dependency to `ecs-logging-python`, modifying the application
```
beniwohli, line 184: Huh, unless I missed something, we _do_ depend on `ecs-logging-python` for log re-formatting:

https://github.com/elastic/apm-agent-python/blob/862b6e105bd2e42a5015c864ee0f9935bef024bb/elasticapm/conf/__init__.py#L412-L423

SylvainJuge: - In order to have ECS reformatting in python, do you need the application to have a direct `ecs-logging-python` dependency ?
- Or does that feature allows to reformat application logs without modifying the application dependencies (beyond adding the agent) ?

In Java, the agent embeds a copy of `ecs-logging-java` and instrument the application loggers at runtime, so from user perspective there is nothing to add beyond the agent.

basepi: Currently, the user must add the `ecs-logging-python` dependency. This is something we need to change, I think we should add `ecs-logging-python` as a direct dependency.

docs/logging.asciidoc
```diff
@@ -1,34 +1,36 @@
...
 
-* <<logging-integrations>>
-* <<log-correlation-in-es>>
+Those features are part of {observability-guide}/application-logs.html[Application log ingestion strategies].
+
+The {ecs-logging-python-ref}/intro.html[`ecs-logging-python`] library can also be used to use the {ecs-logging-ref}/intro.html[ECS logging] format without an APM agent.
+When deployed with the Python APM agent, the agent will provide <<log-correlation-ids,log correlation>> IDs.
 
 [float]
-[[ecs-logging]]
```
basepi, line 13: This is a breaking change, but I suppose it's probably worth standardizing the links between agents.

basepi (approved):

docs/logging.asciidoc
basepi: https://github.com/elastic/apm-agent-python/pull/1840

basepi: Even with that change, this needs to be re-worded:

```suggestion
The agent can automatically reformat application logs to ECS format, by modifying the application
```

SylvainJuge: That's definitely a noticeable difference here, so from what I understand: ECS reformatting is only available when ecs-logging-python is already added to the application.

Does this means all the users have to do is to add a dependency on their application to `ecs-logging-python` or do they also need to configure `ecs-logging-python` in the application ?
- if yes, then it means the agent provides all the configuration for ecs-logging-python
- if no, then it means users still have to configure ecs-logging-python, and the agent only provides the log correlation part.

Also the sentence below will also have to be re-worded to account for this:
> logging configuration and making the application always use ECS log format. In short, it provides the benefits of ECS logging at runtime without any 
change to the application.

basepi: I added it as a dependency in #1840, that's why I added the suggested change above. Once #1840 is merged and released, users will get `ecs_logging` automatically, and the configuration value will start working as expected when set to `override`.

SylvainJuge: In this case, I think the best next move is to do this doc change in two steps:
- document the current implementation in this PR
- update documentation in #1840 to indicate that starting from version 6.16.0 (or whenever it's released), adding the `ecs-logging-python` dependency is no longer requires as it is already included by the agent.

If you agree on this I should be able to modify this PR and make it merge-able, then once merged I will let you update #1840 documentation. Does it sounds like a good plan for you ?

basepi: Sounds fine.

SylvainJuge:
@basepi since you approved before I made post-review changes, do you mind re-reviewing the current state so I can merge it ?

basepi (approved):

PR #1840, opened 2023-05-26
Add ecs_logging to dependencies

basepi:

Adds [`ecs_logging`](https://github.com/elastic/ecs-logging-python) as a dependency, for easier log correlation setup.

## Related issues

Ref #1837

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -46,6 +46,7 @@ endif::[]
 * Add instrumentation for redis.asyncio {pull}1807[#1807]
 * Add support for urllib3 v2.0.1+ {pull}1822[#1822]
 * Add `service.environment` to log correlation {pull}1833[#1833]
+* Add `ecs_logging` as a dependency {pull}1840[#1840]
 
 [float]
 ===== Bug fixes
--- a/docs/logging.asciidoc
+++ b/docs/logging.asciidoc
@@ -181,7 +181,9 @@ Then, you could use a grok pattern like this (for the
 [[log-reformatting]]
 === Log reformatting (experimental)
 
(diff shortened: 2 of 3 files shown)


SylvainJuge (approved):

beniwohli (approved):

sparrowt:
It seems odd for this to be a required dependency when in fact it is only needed if using an optional feature that is still marked as a 'technical preview' https://www.elastic.co/guide/en/apm/agent/python/current/configuration.html#config-log_ecs_reformatting

At least it doesn't have transient dependencies, but it feels strange to require everyone to install this just to use the main elastic apm agent, purely in order to make it easier for the minority to opt in to an extra feature.

Perhaps this could be an `extra` instead?

xrmx:
@sparrowt What does concern you about this dependency? I think that in this case the tradeoff is very much in favor of the convenience of having it always available instead of max efficiency with dependency size (I guess).

sparrowt:
I don't have any major concerns specifically about _this_ dependency, but I guess my reasoning is a combination of:
- being required to install unnecessary dependencies is generally undesirable, obviously build size can be a factor (sure it's small in this case but it all adds up), but also security patching required if a vuln were discovered, etc.
- this feels very much an "optional extra" which seems like the perfect use case for the python packaging `extras` which many other packages use to avoid forcing everyone to install packages which are only needed for some optional sub-feature (this seems especially appropriate given things like _"Note that this is a very blunt instrument that could have unintended side effects"_ [here](https://www.elastic.co/guide/en/apm/agent/python/current/configuration.html#config-log_ecs_reformatting) and it being designated as a 'technical preview')

P.S. typing lacks subtlety etc. so to be clear, please take this in a friendly way and it is not a super major blocker 😃 I'm just explaining the reasoning behind the question & generally many people (myself included) prefer not to add deps unless necessary.

PR #1850, opened 2023-06-11
Add MASK_URL for url sanitization

basepi:
Python 3.11.4 added more strict URL checking when calling `urllib.parse.urlsplit`. This caused the url sanitization to fail when the url contained brackets, as with `[REDACTED]` when redacting passwords in URLs.

I've added a new `constants.MASK_URL` for use in URL masking, without the square brackets.


beniwohli (approved):
Good catch and nice fix!

PR #1852, opened 2023-06-12
implement server_ca_cert_file config option

### LINKED ISSUE #1801: Implement server_ca_cert_file config option
In some deployments, using `server_cert` for pinning the server certificate is not easily doable, and a custom certificate store is preferable. Some Elastic APM agents already support this, namely [Node.js](https://www.elastic.co/guide/en/apm/agent/nodejs/current/configuration.html#server-ca-cert-file), Ruby and Golang.

beniwohli:

While `server_cert` allows usage of self-signed certs for simple cases (there is only one certificate), it's not of great use for larger deployments. Allowing the user to provide a CA certificate with which server certificates are signed is a better fit for such situations.

## Related issues

Closes #1801


basepi (approved):
Looks great!

PR #1863, opened 2023-06-27
add pkg-config to test image

beniwohli:
newest release of mysqlclient requires pkg-config to build.

I also had to switch netcat to netcat-traditional, as the netcat package isn't available in the Python 3.11 base image

### DIFF
--- a/.ci/.matrix_exclude.yml
+++ b/.ci/.matrix_exclude.yml
@@ -198,3 +198,6 @@ exclude:
     FRAMEWORK: grpc-1.24
   - VERSION: python-3.11
     FRAMEWORK: grpc-1.24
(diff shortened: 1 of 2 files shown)


beniwohli:
Bit of a catch-22 here. Can't merge this before tests pass, but tests won't pass before we merge this and docker images are rebuilt.

@basepi how about we disable the mysqlclient test in this PR, merge, wait for rebuild, and then re-enable the test in a follow-up PR?

basepi:
> @basepi how about we disable the mysqlclient test in this PR, merge, wait for rebuild, and then re-enable the test in a follow-up PR?

Sounds workable to me.

basepi (approved):

PR #1870, opened 2023-07-05
Add support for grpc aio server interceptor

### LINKED ISSUE #1869: GRPC aio (async) server instrumentation support
Today I was trying to use the elastic APM for my GRPC service, but when it didn't work I investigated and found out that the instrumentation is implemented only for the "default" version of the grpc server, and not the aio (async) server which is what I currently use in my service.

To solve this, I modified the code a bit to work for an async server and it is currently working fine for me. I'm about to start a PR with the async server instrumentation, is there anything I should be aware about this in the meantime? My idea is just to copy and adapt the existing code, possibly extracting common code to avoid duplication.

felipou:

Add support for grpc aio server (async) instrumentation using an interceptor mostly copied from the "regular", sync, version. For now it's still missing tests, but I'm creating the PR just to check if anything breaks in the CI.

## Related issues

Closes #1869


basepi:
I approved it so tests will run.

felipou:
Thanks @basepi! I saw some problems and fixed them.

I also just tried to push an attempt at the tests. Not sure it will work, I haven't been able to setup my local dev environment yet, will try to do that sometime soon, but if you'd like to run the CI as it is!

felipou:
Turns out it was quite easy to run the tests locally, so I was able to fix some issues and now the tests are running properly.

basepi:
@felipou Can you please sign the [CLA](https://www.elastic.co/contributor-agreement)?

felipou:
Ok, done

elasticapm/instrumentation/packages/grpc.py
```diff
@@ -74,3 +76,30 @@ def call(self, module, method, wrapped, instance, args, kwargs):
...
+class GRPCAsyncServerInstrumentation(AbstractInstrumentedModule):
+    name = "grpc_async_server_instrumentation"
+    creates_transactions = True
+    instrument_list = [("grpc", "aio.server")]
+
+    def get_instrument_list(self):
+        import grpc
+
+        # Check against the oldest version that I believe has the expected API
+        if parse_version(grpc.__version__) >= parse_version("1.33.1") and hasattr(grpc, "aio"):
```
felipou, line 90: This is how I'm checking if we should try to instrument the async version, but I'm not sure what is the recommended way for doing this kind of thing here. I've thought of leaving just the check for the `aio` module's existence (and perhaps add a check for a `server` attribute inside `grpc.aio`), but since I wrote the version check first, I just left it there too.

elasticapm/instrumentation/register.py
```diff
@@ -70,6 +70,7 @@
     "elasticapm.instrumentation.packages.kafka.KafkaInstrumentation",
     "elasticapm.instrumentation.packages.grpc.GRPCClientInstrumentation",
     "elasticapm.instrumentation.packages.grpc.GRPCServerInstrumentation",
+    "elasticapm.instrumentation.packages.grpc.GRPCAsyncServerInstrumentation",
```
beniwohli, line 73: Could you move this to the list below that is gated by the Python 3.7 version check? We don't support asyncio tracing in Python 3.6.

elasticapm/instrumentation/packages/grpc.py
beniwohli: Actually, the whole check here is somewhat unnecessary. While instrumenting, we already check if the module/function we want to wrap is available or not. If we cannot import it, we simply skip it :)

elasticapm/instrumentation/register.py
felipou: Sure, done!

elasticapm/instrumentation/packages/grpc.py
```diff
@@ -74,3 +76,30 @@ def call(self, module, method, wrapped, instance, args, kwargs):
...
         return wrapped(*args, **kwargs)
+
+
+class GRPCAsyncServerInstrumentation(AbstractInstrumentedModule):
+    name = "grpc_async_server_instrumentation"
+    creates_transactions = True
+    instrument_list = [("grpc", "aio.server")]
+
+    def get_instrument_list(self):
+        import grpc
```
basepi, line 87: Can you please surround this with a `try:except` similar to [this example](https://github.com/elastic/apm-agent-python/blob/0eaa44a199365785c61ffe8a5adcf7b7a58a7a8d/elasticapm/instrumentation/packages/elasticsearch.py#L50-L62)?

basepi (changes requested):
One small fix

elasticapm/instrumentation/packages/grpc.py
felipou: Ok, done! Should I also use this technique to check if the `grpc.aio` is present? Or just leave as I've done, with the version check?

basepi: I think we're good. Anything in the `call()` will only be called if we decide to instrument. And anything in the instrument list that isn't able to import will be skipped.

Before 1.33.1, is `grpc.aio` just not able to be imported? Because now that I think about it, if we just set `instrument_list` to something that is only available in 1.33.1+, we can just let the instrumentation system attempt to import it and then abort.

If it's present in <1.33.1 but not working, then we need the version check as you've coded it. But if it's just missing, can probably skip the whole version check.

felipou: I didn't actually check by testing or anything like that, but the release changelog says: "Allow asyncio API to be imported as grpc.aio.", so I'm guessing it actually didn't exist before that version. And the date is close to the first commit in the history of file `__init__.py` in the `grpc/aio` folder. So I think it's a good idea to skip the version check like you suggested! I'll push a commit removing it now.

basepi (approved):

PR #1873, opened 2023-07-10
add Python 3.12 to test matrix

### LINKED ISSUE #1857: Add 3.12 betas to CI
3.12.0b3 is out, we should add 3.12 to our CI well before the release date to catch any issues that may arise.

beniwohli:
currently fails with a segmentation fault. Could be related to https://github.com/python/cpython/issues/106092

## Related issues

Closes #1857


basepi:
I guess we should probably wait to merge this until the segfault goes away? I don't want a whole bunch of extra noise on nightlies.

Did you narrow it to a certain test or is the segfault more general than that?

beniwohli:
The segfault happens with a different test every time. It takes a bit longer with coverage disabled, but it still crashes.

Issue #1835, opened 2023-05-24
Switch to trusted publishers on PyPI


basepi:
Currently the CI is pushing packages to PyPI using a key generated by @beniwohli

PyPI now supports [trusted publishers](https://docs.pypi.org/trusted-publishers/). We should switch to this system.

PR #1874, opened 2023-07-10
ci: use trusted publisher instead of token flow

amannocci:
## What is the change being made?

* Use [trusted publisher](https://docs.pypi.org/trusted-publishers/using-a-publisher/) instead of token flow.

## How has this been tested?

* The workflow has been tested using [`https://test.pypi.org/legacy/`](https://github.com/elastic/apm-agent-python/actions/runs/5507168368).

## Related Issues

Closes https://github.com/elastic/apm-agent-python/issues/1835


.github/workflows/release.yml
```diff
@@ -20,32 +20,19 @@ jobs:
...
       - name: Upload
-        run: |
-          python -m pip install --user twine
-          python setup.py sdist
-          echo "Uploading to ${REPO_URL} with user ${TWINE_USER}"
-          python -m twine upload --username "${TWINE_USER}" --password "${TWINE_PASSWORD}" --skip-existing --repository-url ${REPO_URL} dist/*.tar.gz
-          python -m twine upload --username "${TWINE_USER}" --password "${TWINE_PASSWORD}" --skip-existing --repository-url ${REPO_URL} dist/*.whl
-        env:
-          REPO_URL: "https://upload.pypi.org/legacy/"
+        uses: pypa/gh-action-pypi-publish@f5622bde02b04381239da3573277701ceca8f6a0
```
beniwohli, line 33: I think we still need a step to run `python setup.py sdist`. I think it makes most sense to add it here: https://github.com/elastic/apm-agent-python/blob/main/.github/workflows/packages.yml#L26

beniwohli (approved):
Just a note about generating the source distribution (`sdist`), otherwise LGTM!

.github/workflows/release.yml
amannocci: Does it make sense since We are already relying on this step to generate artifacts and then downloading those artifacts before upload?

beniwohli: exactly. And since `sdist` generates a source distribution package, it makes sense to do it in the `packages.yml` workflow. I'm not sure why we did it here to begin with, probably it was just copy&pasted from the old Jenkins setup.

amannocci: Ah, I got it. You mean to download the artifacts already upload in the other workflow rather than running the step within the same workflow?
I guess it because there are extra steps to download artifacts from another workflow.

beniwohli: Don't we already do that? The whl file we generate in the packages workflow is uploaded in the release workflow AFAIK

basepi (approved):
I have the same concern as @beniwohli but otherwise this looks good.

.github/workflows/release.yml
basepi: Yeah, I don't care where those steps are, but we need to run `python setup.py sdist` somewhere. If you look at this [doc](https://docs.pypi.org/trusted-publishers/using-a-publisher/), it's the `# retrieve your distributions here` piece of the puzzle. We need to generate those packages somewhere so that the publish action can publish them.

amannocci: Currently, we have a job `packages` within the workflow `release` that executes the `python setup.py sdist`.
The `publish-pypi` download the result of the `packages` job which is a list of distribution package.
AFAIU, the `python setup.py sdist` wasn't needed because it was already performed in another job within the `release` workflow.

In our case (within the `release` workflow), we can remove the `python setup.py sdist` step from the `publish-pypi` job and rely on the `packages` job to do it or rely on the `publish-pypi` job to do it and remove the `packages` job.
The first option avoid duplicate CI code since those steps are already present in the `packages` workflow definition.

> Ah, I got it. You mean to download the artifacts already upload in the other workflow rather than running the step within the same workflow?
> I guess it because there are extra steps to download artifacts from another workflow.
In some case, there is also another option which is to download the result of the `build` job from the `packages` workflow.

.github/workflows/release.yml
```diff
@@ -20,32 +20,19 @@ jobs:
       - test
       - packages
     runs-on: ubuntu-latest
-    env:
-      PYPI_SECRET_PATH: secret/apm-team/ci/apm-agent-python-pypi-prod
+    environment: release
```
reakaleek, line 23: Q: What is the purpose of the environment here?

reakaleek (approved):

.github/workflows/release.yml
amannocci: A: It allows to apply specific restriction about who trigger the workflow and is allowed to publish an artifact.

> Configuring an environment is optional, but strongly recommended: with a GitHub environment, you can apply additional restrictions to your trusted workflow, such as requiring manual approval on each run by a trusted subset of repository maintainers.

kuisathaverat: The security applied by using environments is a side effect. I would not recommend using that approach.

reakaleek: > The security applied by using environments is a side effect. I would not recommend using that approach.

From what I read here, it's designed for this usecase https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment.

I was just surprised to see no protection rules as checked in the environment settings. Hence, I did not understand why this was created.

.github/workflows/release.yml
beniwohli: Just to make sure we are on the same page: currently, there are two release artifacts that need to be built and pushed to PyPI:

 * a source distribution, `.tar.gz`, generated by `python setup.py sdist`
 * a "universal wheel" distribution, `.whl`, generated by `python setup.py bdist_wheel` (this command needs the `wheel` package to be installed during the build)

We used to generate a lot more wheel packages for all kinds of different platforms, but since we removed any platform-dependent code, only one universal wheel is needed. Because the package build process was much more complicated in the past, it was split into a separate job, and run on every commit.

I guess since the package building is much simpler these days, we could remove the `packages.yml` job completely and only build the packages in the release workflow.

beniwohli (approved):
LGTM!

PR #1890, opened 2023-08-23
fix "IndexError when "url" not found in args or kwargs"

### LINKED ISSUE #1884: IndexError when "url" not found in args or kwargs.
### Overview

I've found an issue in line 46 of `call()` method in `AioHttpClientInstrumentation(...)` class.

https://github.com/elastic/apm-agent-python/blob/da93e7af448abcac367d216e2d20a584051f6e50/elasticapm/instrumentation/packages/asyncio/aiohttp_client.py#L44-L47

I'm getting an `IndexError` exception due to lack or "url" in both kwargs and in args[1]. The the reason is that the argument containing urls is called "str_or_url".

https://github.com/aio-libs/aiohttp/blob/4b59d55e9e79f5a0b1932d6dc9f6b12a33d19266/aiohttp/client.py#L325-L328

By default the code is running fine, but this issue will appear in cases where someone will try to use `ClientSession._request()` method directly AND use keyword-arguments.

### How to recreate the bug?

This is a general example on how to recreate the bug. Lets assume that somewhere in my code I want to connect to some external http rest-api service using aiohttp library. I'll be using custom made session object based on ClientSession object from aiohttp library.

```python
from aiohttp import ClientSession, ClientResponse

class CustomSession(ClientSession):

    async def _request(self, method: str, str_or_url: StrOrURL, **kwargs: Any) -> ClientResponse:
        # put some extra code here, like add retry functionality
        client_response = await super()._request(method=method, str_or_url=url, **kwargs) # << ISSUE IS HERE
        return client_response
```
the above code is valid and it works as long as there is no apm agent running as middleware. With apm agent added, the code has to be written as below:
```python
        client_response = await super()._request(method, url, **kwargs)
```

### How to fix it?

Do search for both `url` and `str_or_url` in kwargs. It's simple fix that can fit in same line, I bet.

nbarbey:
Closes #1884

### DIFF
--- a/elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
+++ b/elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
@@ -43,9 +43,10 @@ class AioHttpClientInstrumentation(AsyncAbstractInstrumentedModule):
 
     async def call(self, module, method, wrapped, instance, args, kwargs):
         method = kwargs.get("method", args[0])
-        url = str(kwargs.get("url", kwargs.get("str_or_url", None)))
+        url = kwargs.get("url", kwargs.get("str_or_url", None))
         if url is None:
             url = args[1]
+        url = str(url)
 
         signature = " ".join([method.upper(), get_host_from_url(url)])
         url = sanitize_url(url)
--- a/elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
+++ b/elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
@@ -42,9 +42,8 @@ class AioHttpClientInstrumentation(AsyncAbstractInstrumentedModule):
     instrument_list = [("aiohttp.client", "ClientSession._request")]
 
     async def call(self, module, method, wrapped, instance, args, kwargs):
-        method = kwargs["method"] if "method" in kwargs else args[0]
-        url = kwargs["url"] if "url" in kwargs else args[1]
-        url = str(url)
+        method = kwargs.get("method", args[0])
+        url = str(kwargs.get("url", kwargs.get("str_or_url", args[1])))
 
         signature = " ".join([method.upper(), get_host_from_url(url)])
         url = sanitize_url(url)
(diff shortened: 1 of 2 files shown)


basepi:
@nbarbey I opened a PR against your PR, with a test and a little fix: https://github.com/nbarbey/apm-agent-python/pull/1

If you open PRs from a new branch instead of your `main` branch, it allows me to push changes directly to your PR, rather than having to open a new PR against your PR. Should make it easier next time. Thanks for the contribution!

nbarbey:
Ok I will be more careful next time about branch name.
I merge your PR in my fork. It seems it updated this PR automagically, nice !

basepi:

basepi (approved):

elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
```diff
@@ -42,9 +42,10 @@ class AioHttpClientInstrumentation(AsyncAbstractInstrumentedModule):
     instrument_list = [("aiohttp.client", "ClientSession._request")]
 
     async def call(self, module, method, wrapped, instance, args, kwargs):
-        method = kwargs["method"] if "method" in kwargs else args[0]
-        url = kwargs["url"] if "url" in kwargs else args[1]
-        url = str(url)
+        method = kwargs.get("method", args[0])
+        url = str(kwargs.get("url", kwargs.get("str_or_url", None)))
+        if url is None:
+            url = args[1]
```
basepi, line 48:
```suggestion
        url = kwargs.get("url", kwargs.get("str_or_url", None))
        if url is None:
            url = args[1]
        url = str(url)
```

basepi (changes requested):
One more change I missed to get the tests passing.

elasticapm/instrumentation/packages/asyncio/aiohttp_client.py
basepi: @nbarbey Can you commit this suggestion so we can get this merged?

basepi (approved):

basepi:

PR #1893, opened 2023-08-25
Add type hints to execution_context

### LINKED ISSUE #1856: Enhancement Request: Add elasticapm.set_custom_span_context()
**Description**
Currently, the Elastic APM Python library restricts custom context to the transaction and error level. However, there is a use case where the user would like to enrich their spans with metadata, and storing it on the transaction level is not suitable. It would be beneficial to have a method like elasticapm.set_custom_span_context() that applies to the currently active span only.

**Context**
In an internal Slack conversation, it was discussed that custom context can be set for spans using an undocumented argument extra on capture_span(). However, this argument is not namespaced, which means users could accidentally overwrite the internal context.

**Proposed Solution**
To address the limitations and provide a more intuitive way of setting custom context for spans, it is suggested to add the elasticapm.set_custom_span_context() method to the Elastic APM Python library. This method should allow users to specify custom metadata for the currently active span without the risk of overwriting internal context.

**Expected Behavior**
The elasticapm.set_custom_span_context() method should enable users to set custom metadata for the currently active span only. The custom context should be namespaced to avoid conflicts with internal context. This enhancement would provide more flexibility in enriching spans with metadata and better serve the use case mentioned in the Slack conversation.

basepi:

Add type hints to the execution context system.

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -39,7 +39,7 @@ import warnings
 from collections import defaultdict
 from datetime import timedelta
 from types import TracebackType
-from typing import Any, Dict, List, Optional, Sequence, Tuple, Type, TypeVar, Union
+from typing import Any, Callable, Dict, List, Optional, Sequence, Tuple, Type, TypeVar, Union
 
 import elasticapm
 from elasticapm.conf import constants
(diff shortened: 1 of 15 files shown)


elasticapm/traces.py
```diff
@@ -1266,7 +1266,34 @@ def set_context(data, key="custom"):
...
+            data[LABEL_RE.sub("_", k)] = data.pop(k)
+
+    if key in span.context:
+        span.context[key].update(data)
+    else:
+        span.context[key] = data
+
+
 set_custom_context = functools.partial(set_context, key="custom")
+set_custom_span_context = functools.partial(set_span_context, key="custom")
```
trentm, line 1296: The intake API schema (span.json) doesn't include `context.custom`, so will these be ingested, or just dropped?

basepi: Good question! I will do some investigation.

basepi: You're correct; the data is dropped. I think rather than spending the time on apm-data getting this added, I'm going to recommend labels to the user.

basepi:
I ditched everything but the type hints I added.

trentm (approved):

PR #1904, opened 2023-09-19
Merge forward from 6.x to main

basepi:
I didn't notice that #1903 wasn't targeted to `main`. This PR makes sure that change isn't lost when we release.


basepi:
DO NOT click `update branch`.

trentm (approved):

Issue #1851, opened 2023-06-12
Skipping PostgreSQL dollar-quoted string constants extremely slow for queries with many query parameters


gkoller:
**Describe the bug**:

PostgreSQL has a something called [dollar-quoted string constants](https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING). `apm-agent-python`'s [instrumentation.packages.dbapi2.scan](https://github.com/elastic/apm-agent-python/blob/aedd6b1bec8607a45c5cc39557f232e5231f55ff/elasticapm/instrumentation/packages/dbapi2.py#L132) function has logic to detect these dollar quoted string constants and skip them.

However, this logic/code path is also triggered when it comes across query parameters. After all, query parameters also start with a dollar character (`$`). For queries with a very large number of query parameters (eg, 3000-4000) this code path, as determined by running the app under PyCharm with [yappi](https://pypi.org/project/yappi/), is a severe bottle neck. Such a query can be handled in just  tens of miliseconds (30-40 ms) by PostgreSQL, but it subsequently takes `apm-agent-python` more than 10 sec. (!!!). Almost all the time is spent in the [skip_to](https://github.com/elastic/apm-agent-python/blob/aedd6b1bec8607a45c5cc39557f232e5231f55ff/elasticapm/instrumentation/packages/dbapi2.py#L57) function.

Leading to graphs like these:

<img width="1221" alt="screenshot_20230612_113424" src="https://github.com/elastic/apm-agent-python/assets/4290278/2adf526e-d6e8-451c-996a-274338df5ed7">

What kind of a query takes up to 3000 query parameters? Queries with large [VALUES lists](https://www.postgresql.org/docs/15/queries-values.html).

```sql
SELECT *
FROM (VALUES ($1::varchar, $2::varchar, $3::varchar),
             ($4::varchar, $5::varchar, $6::varchar),
             ($7::varchar, $8::varchar, $9::varchar),
             ...
             ($2992::varchar, $2993::varchar, $2994::varchar),
             ($2995::varchar, $2996::varchar, $2997::varchar),
             ($2998::varchar, $2999::varchar, $3000::varchar))
```

BTW the above query is a simplified example. In our use case that large `VALUES` list is part of a much more elaborate `CTE` (`WITH` statement)

**To Reproduce**

1. Use SQL queries with a large `VALUES` list consisting of query parameters.

**Environment (please complete the following information)**
- OS: 
[...]

PR #1905, opened 2023-09-27
Optimise handling PostgreSQL dollar quotes

alexmojaki:
Helps with https://github.com/elastic/apm-agent-python/issues/1851

This avoids slowly looping in Python, relying more on optimised C methods. For this toy benchmark:

```python
from elasticapm.instrumentation.packages.dbapi2 import tokenize, scan
import time

sql = f"Hello $q${'Peter Pan $ ' * 1000000}$q$ at Disney World"
tokens = tokenize(sql)
start = time.time()
list(scan(tokens))
end = time.time()
print(end - start)
```

it's about 3-4x faster with this PR, and about 10x faster if there's no inner `$`.

This is nice, but probably not enough. Both with and without this PR, the code assumes that anything between 2 dollar signs is potentially a dollar quote. But according to https://www.postgresql.org/docs/15/sql-syntax-lexical.html#SQL-SYNTAX-DOLLAR-QUOTING:

> The tag, if any, of a dollar-quoted string follows the same rules as an unquoted identifier, except that it cannot contain a dollar sign

From further up:

> SQL identifiers and key words must begin with a letter **(a-z, but also letters with diacritical marks and non-Latin letters)** or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($).

Checking that the dollar quote is valid should avoid a lot of pointless searching for a closing quote, which should resolve the use case in the issue. It also seems important for correctness.

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -37,6 +37,7 @@ endif::[]
 ===== Features
 
 * Collect the `configured_hostname` and `detected_hostname` separately, and switch to FQDN for the `detected_hostname`. {pull}1891[#1891]
+* Improve postgres dollar-quote detection to be much faster {pull}1905[#1905]
 
 [float]
 ===== Bug fixes
--- a/elasticapm/instrumentation/packages/dbapi2.py
+++ b/elasticapm/instrumentation/packages/dbapi2.py
@@ -54,21 +54,6 @@ class Literal(object):
         return "<Literal {}{}{}>".format(self.literal_type, self.content, self.literal_type)
 
 
-def skip_to(start, tokens, value_sequence):
-    i = start
-    while i < len(tokens):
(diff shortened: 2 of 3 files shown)


alexmojaki:
@cla-checker-service recheck?

basepi (approved):
This looks excellent! Thank you for adding tests as well.

Do you have any interest in tackling the "quote correctness check" piece that is remaining?

alexmojaki:
> Do you have any interest in tackling the "quote correctness check" piece that is remaining?

Not really, sorry :)

basepi:
No problem. I really appreciate this contribution! Big improvement.

basepi (approved):

trentm (approved):

basepi:

PR #1910, opened 2023-10-02
Add python 3.12 to CI and metadata

basepi:


basepi:
Python 3.12 is not on docker yet, thus the extremely failing tests.

trentm (approved):

PR #1914, opened 2023-10-04
Fix race condition with extra metadata + partial transaction extra span

### LINKED ISSUE #1902: [Lambda] Agent is reporting a span for partial transaction reporting in lambda
Because we're sending the partial transaction registration synchronously, it's creating a short span, named `POST 127.0.0.1:8200`.
image

We need to drop this span or not create it in the first place.

basepi:

Eliminates the `add_metadata` interaction with the transport thread. We now generate our own metadata for sending the partial transaction, and also save the extra metadata in the `Client` object so that it makes it into every future metadata build.

Also, fixed the extra span being recorded when we send the partial transaction. Fixes #1902

### DIFF
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -433,10 +433,11 @@ class _lambda_transaction(object):
             and os.environ.get("ELASTIC_APM_LAMBDA_APM_SERVER")
             and ("localhost" in self.client.config.server_url or "127.0.0.1" in self.client.config.server_url)
         ):
+            transaction = execution_context.get_transaction()
             transport = self.client._transport
             logger.debug("Sending partial transaction and early metadata to the lambda extension...")
             data = transport._json_serializer({"metadata": self.client.build_metadata()}) + "\n"
-            data += transport._json_serializer({"transaction": execution_context.get_transaction().to_dict()})
+            data += transport._json_serializer({"transaction": transaction.to_dict()})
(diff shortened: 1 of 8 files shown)


elasticapm/contrib/serverless/aws.py
```diff
@@ -400,10 +400,9 @@ def set_metadata_and_context(self, coldstart: bool) -> None:
         metadata["service"]["node"] = {"configured_name": os.environ.get("AWS_LAMBDA_LOG_STREAM_NAME")}
         # This is the one piece of metadata that requires deep merging. We add it manually
         # here to avoid having to deep merge in _transport.add_metadata()
-        if self.client._transport._metadata:
-            node_name = nested_key(self.client._transport._metadata, "service", "node", "name")
-            if node_name:
-                metadata["service"]["node"]["name"] = node_name
```
trentm, line 406: This `metadata.service.node.name` was a separate issue? AFAIK there is no such field in the intake schema.

trentm (approved):
I'll trust you on the `.pause_sampling`. I don't know the Python agent code well enough to grok that.

elasticapm/contrib/serverless/aws.py
basepi: It was an error I found while fixing the metadata bug. I was looking for the wrong field; the code did nothing.

lahsivjar (approved):
Tested with the branch and didn't see any errors.

trentm (approved):

PR #1915, opened 2023-10-05
chore(typing): Add explicit None return type

orsinium:

If a function doesn't return anything, annotate its return type as None. It will make sure that the callee doesn't try to assign the function result to a variable.

> It is recommended but not required that checked functions have annotations for all arguments and the return type. For a checked function, the default annotation for arguments and for the return type is Any.
>
> https://peps.python.org/pep-0484/#the-meaning-of-annotations

## Related issues

-

### DIFF
--- a/elasticapm/traces.py
+++ b/elasticapm/traces.py
@@ -70,20 +70,20 @@ _AnnotatedFunctionT = TypeVar("_AnnotatedFunctionT", bound=FuncType)
 class ChildDuration(object):
     __slots__ = ("obj", "_nesting_level", "_start", "_duration", "_lock")
 
-    def __init__(self, obj: "BaseSpan"):
+    def __init__(self, obj: "BaseSpan") -> None:
         self.obj = obj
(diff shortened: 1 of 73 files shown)


elasticapm/traces.py
```diff
@@ -894,7 +894,7 @@ def action(self):
 
 
 class Tracer(object):
-    def __init__(self, frames_collector_func, frames_processing_func, queue_func, config, agent: "elasticapm.Client"):
+    def __init__(self, frames_collector_func, frames_processing_func, queue_func, config, agent: "elasticapm.Client") -> None:
```
basepi, line 897:
```suggestion
    def __init__(
        self, frames_collector_func, frames_processing_func, queue_func, config, agent: "elasticapm.Client"
    ) -> None:
```

basepi (changes requested):
Just one change to make `black` happy

basepi (approved):

basepi:

PR #1918, opened 2023-10-11
Remove log shipping from Flask docs, start deprecation

basepi:
## What does this pull request do?

Starts the deprecation process for the log shipping in Flask.

PR #1921, opened 2023-10-18
Mark the old logs-as-messages shipper as deprecated

basepi:

Finishes what was started in https://github.com/elastic/apm-agent-python/pull/1918

### DIFF
--- a/CHANGELOG.asciidoc
+++ b/CHANGELOG.asciidoc
@@ -29,6 +29,21 @@ endif::[]
 //===== Bug fixes
 //
 
+=== Unreleased
+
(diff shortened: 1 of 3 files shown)


pheki:
Will it be removed or deprecated in v7.0? The warning says states that it's gonna be deprecated, but the CHANGELOG states it will be removed

basepi:
I plan to remove it in v7.0. The language around `PendingDeprecationWarning` vs `DeprecationWarning` has always been confusing to me. But yes I'll fix the wording in the code.

pheki:
Ok, thanks!

PR #1938, opened 2023-11-24
feat(dbapi2): include object name in procedure call spans

### LINKED ISSUE #1937: dbapi2 span does not include object name when calling a procedure

Aggregating metrics of stored procedures is not possible because the span name lacks the object name of the procedure being called.
Also, the span.action is only set to `exec` when a procedure is executed through `callproc`, not connection.execute etc.
Now it's set to `query` for EXECUTE statements, depending on the dbapi2 method used.

* Set span.action to the EXEC_ACTION constant for all EXEC/EXECUTE statements
* Include the object name / procedure name in the span.name, such as `EXECUTE sp_who` instead of just `EXECUTE`

Custom spans can be used, but that duplicates the number of spans for each EXECUTE statement

**Additional context**

Example of current output:

Screenshot 2023-11-24 at 18 39 31

anderssonjohan:

* include object name in span name when calling procedures
* consistently use span action `exec` when calling procedures
 
## Related issues

Closes #1937

### DIFF
--- a/elasticapm/instrumentation/packages/dbapi2.py
+++ b/elasticapm/instrumentation/packages/dbapi2.py
@@ -203,6 +203,13 @@ def extract_signature(sql):
 
 QUERY_ACTION = "query"
 EXEC_ACTION = "exec"
+PROCEDURE_STATEMENTS = ["EXEC", "EXECUTE", "CALL"]
+
+
+def extract_action_from_signature(signature, default):
+    if signature.split(" ")[0] in PROCEDURE_STATEMENTS:
+        return EXEC_ACTION
+    return default
 
 
 class CursorProxy(wrapt.ObjectProxy):
@@ -235,6 +242,7 @@ class CursorProxy(wrapt.ObjectProxy):
             signature = sql_string + "()"
         else:
             signature = self.extract_signature(sql_string)
+            action = extract_action_from_signature(signature, action)
 
         # Truncate sql_string to 10000 characters to prevent large queries from
         # causing an error to APM server.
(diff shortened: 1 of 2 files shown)


basepi:
Please merge/rebase this against `main` to get the tests passing. It appears you have pushes from maintainers disabled.

elasticapm/instrumentation/packages/dbapi2.py
```diff
@@ -170,30 +170,44 @@ def extract_signature(sql):
...
+            object_name = look_for_table(sql, "FROM")
         except Exception:
-            table_name = ""
+            object_name = ""
+    elif sql_type in ["EXEC", "EXECUTE"]:
+        sql_type = "EXECUTE"
+        end = second_space if second_space > first_space else len(sql)
+        object_name = sql[first_space + 1 : end]
+    elif sql_type == "CALL":
+        first_paren = sql.find("(", first_space)
```
basepi, line 192: I'm not certain this line is quite right. It assumes that if there is no parenthesis, we'll grab from the first space to the second to last character (as `first_paren` will be `-1`). What if there's no trailing semicolon? Maybe add a test case for `("CALL get_car_stats_by_year", "CALL get_car_stats_by_year()"),`?

basepi (changes requested):
One question below. But overall this looks great!

anderssonjohan:
Thanks for the comments, I'll add the tests.

Regarding the access I don't see a checkbox.. probably because of the cross-org issue mentioned here: https://github.com/orgs/community/discussions/5634

elasticapm/instrumentation/packages/dbapi2.py
anderssonjohan: Good catch! I added the test case and refactored the mentioned part.

Note: In the current commit I have `object_name = procedure_name + "()"`. The added parens are there to be consistent with the already existing code in `_trace_sql` for calls coming from `def callproc`.
If `object_name = procedure_name` is desired, the pre-existing code in _trace_sql should be updated accordingly.

basepi: I think your version is fine. Thanks!

basepi (approved):

Issue #1872, opened 2023-07-06
support for asynchronous psycopg3


Antonkoch999:
psycopg3 was released some time ago

You started supporting psycopg in version 6.16, but only synchronous operations.
Tell me, please, is it possible to implement support for asynchronous operations too?

https://www.psycopg.org/psycopg3/docs/advanced/async.html#async

basepi:
We should definitely add support for async psycopg3 as well. Thanks for the request!

zikphil:
@basepi Good morning. Do we know if/when we should expect monitoring options for AsyncIO versions of DBAPIv2 compliant objects?

PR #1944, opened 2023-11-30
Added Async support for DBAPI2 w/ implementation for psycopg

zikphil:
Quick and dirty patch to introduce proper DBAPIv2 support for AsyncIO-based drivers, like psycopg. Cannot attest to full functionality but I will run on it for a while and report back or adjust if need be.

Update: Not sure why but when tried to reset my branch to run pre-commit it just closed the previous PR. So here is the new one with pre-commit hooks @basepi 

Closes https://github.com/elastic/apm-agent-python/issues/1872


basepi:

Issue #1701, opened 2022-11-29
[Starlette] ordering of adding the ElasticAPM is no relevant and must be last middleware to be added


bjoernmeier:
**Describe the bug**:

Since elastic-apm>=6.13.0 we did not get transaction information anymore from our FastAPI application. It turned out that since then the order of adding multiple layers of middleware is relevant. If a middleware is added after `elasticapm.contrib.starlette.ElasticAPM` the transaction information is missing when the transaction is closed here https://github.com/elastic/apm-agent-python/blob/main/elasticapm/instrumentation/packages/asyncio/starlette.py#L52. Meaning the call here https://github.com/elastic/apm-agent-python/blob/main/elasticapm/base.py#L324 returns None in such a case.

Reordering of adding the middleware layers does resolve the issue.

**To Reproduce**

1. create a FastAPI app:
 ```
import time

from fastapi import FastAPI
from elasticapm.contrib.starlette import make_apm_client, ElasticAPM
from starlette.middleware import Middleware
from starlette.middleware.base import BaseHTTPMiddleware


class CustomHeaderMiddleware(BaseHTTPMiddleware):
    async def dispatch(self, request, call_next):
        start_time = time.time()
        response = await call_next(request)
        process_time = time.time() - start_time
        response.headers["X-Process-Time"] = str(process_time)
        return response

apm = make_apm_client()
app = FastAPI()

# this ordering works
# app.add_middleware(CustomHeaderMiddleware)
app.add_middleware(ElasticAPM, client=apm)

# this order does not work
app.add_middleware(CustomHeaderMiddleware)

@app.get("/ok")
async def ok():
    return {"details": "ok"}


@app.get("/raise")
async def raise_route():
    raise ValueError("oho")

 ```

2. install requirements: elastic-apm, fastapi, uvicorn
3. run the app with an APM server running and define how to reach it:
```
ELASTIC_APM_SERVER_URL=http://localhost:8200
ELASTIC_APM_SECRET_TOKEN=
ELASTIC_APM_SERVICE_NAME=my-service
ELASTIC_APM_ENVIRONMENT=test
```

run it with: `uvicorn app:app`
4. check if transactions are send and complete in APM, which is not the case

**Environment (please complete the following information)**
- OS: Debian bullseye and OS X
- Python version: 3.10
- Framework and version: elastic-apm-6.13.2 fastapi-0.88.0 starlette-0.22.0 uvicorn-0.20.0
[...]

PR #1956, opened 2024-01-23
Fix:Update starlette.asciidoc, incorrect middleware order

mariocandela:

The documentation contains an error, you need to put the APM middleware before BaseHTTPMiddleware.


mariocandela:
> I don't think this is correct:
> 
> > Using BaseHTTPMiddleware will prevent changes to [contextlib.ContextVar](https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar)s from propagating upwards.
> 
> So, BaseHTTPMiddleware needs to be above (i.e. before) our middleware in the list so that we have an unbroken contextvar chain from the endpoint to our middleware.
> 
> Does this not match your experience?
> 
> Edit: see also the original issue: #1701
Yesterday, we integrated the APM agent across all our microservices, placing the agent's middleware as the last element. No data was reaching the APM server in this configuration. However, when we placed it as the first element, everything worked correctly. Could it be that one of our custom middlewares was invalidating the context variable?

Thanks for your time Colton :1st_place_medal:

basepi:
Are you passing in the middleware as a list or using the `add_middleware()` function repeatedly? I think that may be where the confusion is -- the original issue #1701 was using `add_middleware`, which appears to insert the middleware at the front of the list. Thus, the required order would be swapped based on how you're adding the middleware.

I think we need to expand this warning with examples of both.

v1v:
run docs-build

basepi:
@mariocandela Can you answer my question above?

mariocandela:
> @mariocandela Can you answer my question above?

Hi @basepi, sorry for the delay, we use middleware as a list.
Initially, we had the agent's middleware placed last, but data wasn't reaching the APM server in this setup. However, when we moved it to the first position, everything started working correctly.

I haven't had the chance to delve deeper into the issue yet. It's possible that our custom middleware, which utilizes dependency injection, could be causing the problem. I'll make sure to allocate some time to debug the APM and figure out what's going on with the contextvar.

Thanks for your time

basepi:
Alright, you've confirmed my suspicion. Our middleware must be first in the list if you're passing in a list of middleware, but last if you're adding each one manually with add_middleware. I'll add a commit to your PR clarifying that point.

docs/starlette.asciidoc
```diff
@@ -43,7 +43,7 @@ app.add_middleware(ElasticAPM)
 ----
 
 WARNING: If you are using any `BaseHTTPMiddleware` middleware, you must add them
-*before* the ElasticAPM middleware. This is because `BaseHTTPMiddleware` breaks
+*after* the ElasticAPM middleware. This is because `BaseHTTPMiddleware` breaks
 `contextvar` propagation, as noted
 https://www.starlette.io/middleware/#limitations[here].
```
basepi, line 48:
```suggestion
WARNING: `BaseHTTPMiddleware` breaks `contextvar` propagation, as noted
https://www.starlette.io/middleware/#limitations[here]. This means the
ElasticAPM middleware must be above any `BaseHTTPMiddleware` in the final
middleware list. If you're calling `add_middleware` repeatedly, add the
ElasticAPM middleware last. If you're passing in a list of middleware,
ElasticAPM should be first on that list.
```

basepi (changes requested):
One change. Thanks again for talking through this with me!

mariocandela:
> One change. Thanks again for talking through this with me!

thank you for your time!

basepi (approved):

basepi:
run docs-build

PR #1958, opened 2024-01-26
Add ability to run full matrix manually and on push

reakaleek:

- Add ability to run full matrix manually
- Runs full matrix on push (currently only `main`)

## Related issues

Closes #ISSUE

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,8 +11,6 @@ permissions:
 jobs:
   test:
     uses: ./.github/workflows/test.yml
-    with:
-      full-matrix: true
 
   packages:
     uses: ./.github/workflows/packages.yml
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,6 +11,8 @@ permissions:
 jobs:
   test:
     uses: ./.github/workflows/test.yml
+    with:
+      full-matrix: true
 
   packages:
     uses: ./.github/workflows/packages.yml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,7 @@
 name: test # The name must be the same as in test-docs.yml
 
 on:
+  workflow_call: ~
   pull_request:
     paths-ignore:
       - "**/*.md"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,7 +1,12 @@
 name: test # The name must be the same as in test-docs.yml
 
 on:
-  workflow_call: ~
+  workflow_call:
+    inputs:
+      full_matrix:
+        description: "Run the full matrix"
+        required: true
+        type: boolean
   pull_request:
     paths-ignore:
       - "**/*.md"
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -11,6 +11,8 @@ permissions:
 jobs:
   test:
     uses: ./.github/workflows/test.yml
+    with:
+      full-matrix: true
```
v1v, line 15: IIUC tags are generated from commits in the `main` branch, and those commits are already verified in the CI - I'd assume the matrix runs for all the merge-commits in `main`.

What's the rationale for running them again?

reakaleek: Currently, the full matrix does not run on push to main. Only the partial matrix.

Currently, only the scheduled job executes the full matrix.

v1v: > Currently, the full matrix does not run on push to main. Only the partial matrix.

What if they run by default on the main, so there is no need to run them on the releases.

There is a chance, releases might be affected by the test flakiness, and I think the release process should be as smoothy as possible if the other validations run by default.

What do you think?

reakaleek: Makes sense.

reakaleek: https://github.com/elastic/apm-agent-python/pull/1958/commits/6021774820d3fd25bfde33bc8e0e3aca1c04996d

reakaleek: I will do the suggested comment command like in in apm-agent-nodejs in a follow-up.

.github/workflows/test.yml
```diff
@@ -31,9 +36,9 @@ jobs:
         uses: elastic/apm-pipeline-library/.github/actions/version-framework@current
         with:
           # Use .ci/.matrix_python_full.yml if it's a scheduled workflow, otherwise use .ci/.matrix_python.yml
-          versionsFile: .ci/.matrix_python${{ github.event_name == 'schedule' && '_full' || '' }}.yml
+          versionsFile: .ci/.matrix_python${{ (github.event_name == 'schedule' || github.event_name == 'push' || inputs.full-matrix) && '_full' || '' }}.yml
           # Use .ci/.matrix_framework_full.yml if it's a scheduled workflow, otherwise use .ci/.matrix_framework.yml
-          frameworksFile: .ci/.matrix_framework${{ github.event_name == 'schedule' && '_full' || '' }}.yml
+          frameworksFile: .ci/.matrix_framework${{ (github.event_name == 'schedule' || github.event_name == 'push' || inputs.full-matrix) && '_full' || '' }}.yml
```
v1v, line 41: Will this happen for push events on any other upstream branch but `main`?  I don't know how often those branches are created, but I wanted to note it.

reakaleek: The trigger is configured to `main`. Hence, it will only happen for the main branch.

```yaml
  push:
    branches:
      - main
    paths-ignore:
      - "**/*.md"
      - "**/*.asciidoc"
```

v1v (approved):

v1v (approved):

basepi (approved):

PR #1976, opened 2024-02-22
instrumentation: make query scanning for dollar quotes a bit more correct

xrmx:

Dollar quotes follow the same rules as SQL identifiers so:

SQL identifiers and key words must begin with a letter (a-z, but also letters with diacritical marks and non-Latin letters) or an underscore (_). Subsequent characters in an identifier or key word can be letters, underscores, digits (0-9), or dollar signs ($).

Given we are not going to write a compliant SQL parser at least handle query parameters that are simple to catch since they have a digit right after the opening $.

## Related issues

Refs #1851


basepi (approved):
Nice!

PR #1982, opened 2024-03-04
contrib/serverless/aws: normalize headers with v1 event format

### LINKED ISSUE #1980: Missing header normalization on AWS lambda
**Describe the bug**: 

It looks we are missing some case normalization in the AWS Lambda instrumentation. Even if headers are documented as being already lowercase in [AWS docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html), it looks like there are cases where this is not.

Nodejs agent already handles the issue here:
https://github.com/elastic/apm-agent-nodejs/blob/2b69514d0b2c97bb974346cce975cf0cc958d559/lib/lambda.js#L705-L725

Outcome of this is that we are not getting any traceparent or tracestate header and so we are not able to correlate transactions.

**To Reproduce**

1. run apm agent instrumentation for aws lambda
2. inspect the event and see headers are not normalized
3. see that we are not able to get traceparent or tracestate header

- OS: [e.g. Linux] aws lambda
- Python version: -
- Framework and version [e.g. Django 2.1]: -
- APM Server version: -
- Agent version: latest

xrmx:

If the event has been generated from elb or api gateway and is in the v1 format normalize headers even if they already should per https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html

Event discrimination logic courtesy of nodejs apm client.

## Related issues

Closes #1980

### DIFF
--- a/elasticapm/contrib/serverless/aws.py
+++ b/elasticapm/contrib/serverless/aws.py
@@ -135,6 +135,18 @@ def prep_kwargs(kwargs=None):
     return kwargs
 
 
+def should_normalize_headers(event: dict) -> bool:
+    """
+    Helper to decide if we should normalize headers or not depending on the event
+
+    Even if the documentation says that headers are lowercased it's not always the case for format version 1.0
+    https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html
+    """
+
+    request_context = event.get("requestContext", {})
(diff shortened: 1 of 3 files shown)


trentm (approved):
LGTM, with the caveat that I haven't tried it, nor have I learn this module ;)

basepi (approved):

PR #1983, opened 2024-03-04
Remove hound integration

xrmx:

We have already pre-commit running on CI and it doesn't look to be running at the moment.

## Related issues

None


v1v:
> @v1v Is it correct that the integrations with hound is not enabled right?

That\'s correct, I didn't even knew we had hound. If needed we can figure out how to integrate it

basepi (approved):

PR #1987, opened 2024-03-06
feat: add typing on some public api functions

pbabics:

- Adds typing on **some** of the publicly exposed API functions
- Explicitly export API in `__init__.py`


xrmx:
Thanks for the PR. We should probably add some mypy run in CI to check we are not messing up anything.

pbabics:
Hello, I've checked mypy output and there is a lot of errors because most of the code either does not have typing or has wrong/incomplete typing

xrmx:
> Hello, I've checked mypy output and there is a lot of errors because most of the code either does not have typing or has wrong/incomplete typing

Right, maybe we can think of an incremental strategy for typing. Let me think about it a bit.

PR #1990, opened 2024-03-07
ci: add missing full-matrix input when using test workflow in release

xrmx:

Should fix the following error when tagging a release:

The workflow is not valid. .github/workflows/release.yml (Line: 13, Col: 11):
 Input full-matrix is required, but not provided while calling.

See https://github.com/elastic/apm-agent-python/actions/runs/8179350148/workflow

## Related issues

None


reakaleek (approved):
Thank you for fixing!

Issue #1994, opened 2024-03-07
Make it possible to build test artifacts


xrmx:
**Is your feature request related to a problem? Please describe.**

We are not able to catch problems in the build workflows until it's too late. Since our release process requires to cut a new release and the fact that we are distributed that'll adds tons of latency for getting a package out of the building.

Recent incidents:

https://github.com/elastic/apm-agent-python/pull/1990
https://github.com/elastic/apm-agent-python/pull/1993 and https://github.com/elastic/apm-agent-python/pull/1996

**Describe the solution you'd like**

I'd like to being able to run the release workflows pushing test packages / lambdas / you name it to test instances so we can check that the workflows are working.

Ideally this would be run every time we touch .github/workflows to catch dependabot regressions :)

The emphasis here is running the very same github workflows and not on publishing test artifacts.

**Describe alternatives you've considered**

None I can think of.

**Additional context**

In python we have a test package index (https://test.pypi.org/) where we can push things without bothering anyone.

v1v:
I enabled the test.pypi as explained in https://docs.github.com/en/actions/deployment/security-hardening-your-deployments/configuring-openid-connect-in-pypi

<img width="867" alt="image" src="https://github.com/elastic/apm-agent-python/assets/2871786/4d8b4dca-7704-4b0d-94ea-ab01df47e6ad">

PR #1998, opened 2024-03-08
ci: run the release when merges to main

v1v:

Use https://test.pypi.org/ for testing the Release GitHub workflow

## Related issues

Closes https://github.com/elastic/apm-agent-python/issues/1994

## Test

I created a feature branch called `test/release` -> https://github.com/elastic/apm-agent-python/actions/runs/8204978343

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,6 +4,8 @@ on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      - main
 
 permissions:
   contents: read
@@ -31,10 +33,16 @@ jobs:
         with:
           name: packages
           path: dist
-      - name: Upload
+      - name: Upload pipy
+        if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@2f6f737ca5f74c637829c0f5c3acd0e29ea5e8bf
         with:
           repository-url: https://upload.pypi.org/legacy/
+      - name: Upload pipy test
+        if: github.ref == 'refs/heads/main'
+        uses: pypa/gh-action-pypi-publish@2f6f737ca5f74c637829c0f5c3acd0e29ea5e8bf
+        with:
+          repository-url: https://test.pypi.org/legacy/
 
   build-distribution:
     uses: ./.github/workflows/build-distribution.yml
@@ -59,6 +67,7 @@ jobs:
           name: build-distribution
           path: ./build
       - name: Publish lambda layers to AWS
+        if: startsWith(github.ref, 'refs/tags')
         run: |
           # Convert v1.2.3 to ver-1-2-3
           VERSION=${GITHUB_REF_NAME/v/ver-}
@@ -66,6 +75,7 @@ jobs:
 
           ELASTIC_LAYER_NAME="elastic-apm-python-${VERSION}" .ci/publish-aws.sh
       - uses: actions/upload-artifact@v4
+        if: startsWith(github.ref, 'refs/tags')
         with:
           name: arn-file
           path: ".arn-file.md"
@@ -106,6 +116,7 @@ jobs:
           ${{ steps.setup-docker.outputs.name }}:${{ steps.setup-docker.outputs.tag }}
           ${{ steps.setup-docker.outputs.name }}:latest
       - name: Docker push
+        if: startsWith(github.ref, 'refs/tags')
         run: |-
           docker push ${{ steps.setup-docker.outputs.name }}:${{ steps.setup-docker.outputs.tag }}
           docker push ${{ steps.setup-docker.outputs.name }}:latest
@@ -115,6 +126,7 @@ jobs:
       contents: write
     needs:
       - publish-lambda-layers
+    if: startsWith(github.ref, 'refs/tags')
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
@@ -133,7 +145,7 @@ jobs:
 
   notify:
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ always() && startsWith(github.ref, 'refs/tags') }}
     needs:
       - publish-lambda-layers
       - publish-pypi


.github/workflows/release.yml
```diff
@@ -31,10 +29,16 @@ jobs:
         with:
           name: packages
           path: dist
-      - name: Upload
+      - name: Upload pipy
```
xrmx, line 32: It's pypi

.github/workflows/release.yml
```diff
@@ -31,10 +29,16 @@ jobs:
...
+        if: startsWith(github.ref, 'refs/tags')
         uses: pypa/gh-action-pypi-publish@2f6f737ca5f74c637829c0f5c3acd0e29ea5e8bf
         with:
           repository-url: https://upload.pypi.org/legacy/
+      - name: Upload pipy test
```
xrmx, line 37: same

.github/workflows/release.yml
```diff
@@ -4,22 +4,20 @@ on:
...
       - "v*.*.*"
+    branches:
+      - main
+      - test/release
 
 permissions:
   contents: read
 
 jobs:
-  test:
```
xrmx, line 12: is it possible to keep this `if: startsWith(github.ref, 'refs/tags')` ?

xrmx (commented):
Thanks @v1v !

.github/workflows/release.yml
v1v: Ups... I should not commit this change at all

.github/workflows/release.yml
```diff
@@ -4,11 +4,15 @@ on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      - main
+      - test/release
```
v1v, line 9:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -4,11 +4,14 @@ on:
...
     tags:
       - "v*.*.*"
+    branches:
+      - main
 
 permissions:
   contents: read
 
 jobs:
+
```
v1v, line 14:
```suggestion
```

.github/workflows/release.yml
```diff
@@ -133,7 +151,7 @@ jobs:
 
   notify:
     runs-on: ubuntu-latest
-    if: always()
+    if: ${{ always() }}
```
v1v, line 154:
```suggestion
    if: always()
```

.github/workflows/release.yml
```diff
@@ -13,11 +13,17 @@ permissions:
 
 jobs:
 
+  test:
```
xrmx: will adding `if: startsWith(github.ref, 'refs/tags')` work? Otherwise we'll always runs the full matrix tests right?

v1v: That's correct, I can skip the `test` for the `main` branch. That's already part of the default workflow with the scheduler.

v1v:
@xrmx , I might need your help with this approach:

> PyPI does not allow for a filename to be reused, even once a project has been deleted and recreated.

Then, a second run failed:

https://github.com/elastic/apm-agent-python/actions/runs/8229823579/job/22501787751

What do you suggest?

.github/workflows/release.yml
v1v: See https://github.com/elastic/apm-agent-python/pull/1998/commits/054de934800ac2dbc9e1d37c8c8793c9cf3e9dd6

A bit complicated, but that's how I managed to skip those tests for the releases without changing the `test.yaml`.

What do you think?

xrmx:
@v1v can we postfix a `+$GITHASH` to the release version? See https://peps.python.org/pep-0440/#adding-local-version-identifiers, the paragraph before last.

.github/workflows/release.yml
xrmx: LGTM

v1v:
> can we postfix a +$GITHASH to the release version? See https://peps.python.org/pep-0440/#adding-local-version-identifiers, the paragraph before last.

Nice, but I have no clue how to do it; I can try to figure out but if you know how to, would you mind contributing to this PR? 🙏

xrmx:
> > can we postfix a +$GITHASH to the release version? See https://peps.python.org/pep-0440/#adding-local-version-identifiers, the paragraph before last.
> 
> Nice, but I have no clue how to do it; I can try to figure out but if you know how to, would you mind contributing to this PR? 🙏

Sure

> > can we postfix a +$GITHASH to the release version? See https://peps.python.org/pep-0440/#adding-local-version-identifiers, the paragraph before last.
> 
> Nice, but I have no clue how to do it; I can try to figure out but if you know how to, would you mind contributing to this PR? 🙏

Sure, will take a look later today

xrmx:
@v1v What do you think about somethings like this

```
diff --git a/setup.py b/setup.py
index 23ebec33..235bf4d8 100644
--- a/setup.py
+++ b/setup.py
@@ -100,7 +100,11 @@ def get_version():
     for line in version_file:
         if line.startswith("__version__"):
             version_tuple = ast.literal_eval(line.split(" = ")[1])
-            return ".".join(map(str, version_tuple))
+            version_str = ".".join(map(str, version_tuple))
+            local_version = os.getenv("ELASTIC_CI_LOCAL_VERSION")
+            if local_version:
+                return f"{version_str}+{local_version}"
+            return version_str
     return "unknown"
 ```
 
and then exposing to the packages workflow an ELASTIC_CI_LOCAL_VERSION environment variable with the value of `${{ github.run_id }}`?

**Comment by v1v** [comment]
no single quotes, but empty instead, single quotes will be empty in this particular case, I think that's the way to set empty env variables

File: .github/workflows/release.yml [hunk 8]
```diff
@@ -4,15 +4,18 @@ on:
   push:
     tags:
       - "v*.*.*"
+    branches:
+      - main
 
 permissions:
   contents: read
 
 jobs:
```
v1v, line 13:
```suggestion
# Override the version if there is no tag release.
env:
  ELASTIC_CI_LOCAL_VERSION: ${{ startsWith(github.ref, 'refs/tags') && '' || github.run_id }}

jobs:
```

.github/workflows/release.yml
```diff
@@ -4,15 +4,22 @@ on:
...
       - "v*.*.*"
+    branches:
+      - main
 
 permissions:
   contents: read
 
+# Override the version if there is no tag release.
+env:
+  ELASTIC_CI_LOCAL_VERSION: ${{ startsWith(github.ref, 'refs/tags') && '' || github.run_id }}
```
trentm, line 15: Do you need *ordering* of these published releases? Or is just any unique-ish identifier sufficient?

xrmx: I don't think we care about ordering or the packages built. My goal here is just to check that the github workflow for releases works fine.

trentm: Oh, I guess this (`run_id`) works for ordering. All good.

setup.py
```diff
@@ -100,7 +100,11 @@ def get_version():
     for line in version_file:
         if line.startswith("__version__"):
             version_tuple = ast.literal_eval(line.split(" = ")[1])
-            return ".".join(map(str, version_tuple))
+            version_str = ".".join(map(str, version_tuple))
```
v1v, line 103: I picked your changes from https://github.com/elastic/apm-agent-python/pull/2000

🥺 , I wanted to validate if things were working fine! 🙏

setup.py
```diff
@@ -100,7 +100,11 @@ def get_version():
...
+            local_version = os.getenv("ELASTIC_CI_LOCAL_VERSION")
+            if local_version:
+                return f"{version_str}-dev{local_version}"
```
v1v, line 106: as discussed in https://github.com/elastic/apm-agent-python/pull/2000#issuecomment-1988917372

.github/workflows/release.yml
```diff
@@ -91,30 +104,36 @@ jobs:
...
-          ${{ steps.setup-docker.outputs.name }}:latest
+          ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+          ${{ env.DOCKER_IMAGE_NAME }}:latest
       - name: Docker push
+        if: startsWith(github.ref, 'refs/tags')
         run: |-
-          docker push ${{ steps.setup-docker.outputs.name }}:${{ steps.setup-docker.outputs.tag }}
-          docker push ${{ steps.setup-docker.outputs.name }}:latest
+          docker push ${{ env.DOCKER_IMAGE_NAME }}:${{ steps.setup-docker.outputs.tag }}
+          docker push ${{ env.DOCKER_IMAGE_NAME }}:latest
```
v1v, line 129: >   -a, --all-tags                Push all tags of an image to the repository

```suggestion
          docker push --all-tags ${{ env.DOCKER_IMAGE_NAME }}
```

reakaleek (approved):

xrmx (approved):
Thanks!

PR #2000, opened 2024-03-11
setup: take a local version from environment variable

xrmx:

Make it possible to add a local version to the current agent version at build time. This will be useful because we want to continuously build the agent on CI and avoid version conflict when publishing to test pypi.
So when returning the agent version add anything that has been passed in ELASTIC_CI_LOCAL_VERSION environment variable.

## Related issues

Refs #1994
Refs #1998


trentm:
@xrmx Aside: I think there is an issue either with the "labeler" workflow in this repo, or with your team membership, because I would not expect PRs from you to be labelled "community" and "triage".

trentm (dismissed):
Looks fine.

Have you considered "dev releases" per https://peps.python.org/pep-0440/#developmental-releases rather than "local versions" https://peps.python.org/pep-0440/#local-version-identifiers ?

The docs on local versions includes:
> Local version identifiers SHOULD NOT be used when publishing upstream projects to a public index server, but MAY be used to identify private builds created directly from the project source. Local version identifiers SHOULD be used by downstream projects when releasing a version that is API compatible with the version of the upstream project identified by the public version identifier, but contains additional changes (such as bug fixes). As the Python Package Index is intended solely for indexing and hosting upstream projects, it MUST NOT allow the use of local version identifiers.

I'm not sure of the full intent of these "releases from main".

v1v:
I tested your changes and:

```
25hWARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
         '6.[21](https://github.com/elastic/apm-agent-python/actions/runs/8236346346/job/22522564642#step:6:22).3+8[23](https://github.com/elastic/apm-agent-python/actions/runs/8236346346/job/22522564642#step:6:24)6346346' is an invalid value for Version. Error: Can't use  
         PEP 440 local versions. See                                            
         https://packaging.python.org/specifications/core-metadata for more     
         information.
```

in https://github.com/elastic/apm-agent-python/actions/runs/8236346346/job/22522564642#step:6:26

xrmx:
I don't think test.pypi.org is considerd a public index server since the packages name ownership is not shared with the production one so you have no guarantee you are instaling the same package. I haven't checked if it'll accept a package with the local version though. I can use a .postX maybe.

v1v:
>  I can use a .postX maybe.

https://github.com/pypi/legacy/issues/731 , they suggest to remove `+...` , so I tried with your changes and It works:

> Checking dist/elastic_apm-6.21.3.dev8236509058-py2.py3-none-any.whl: PASSED with warnings

https://github.com/elastic/apm-agent-python/actions/runs/8236509058

xrmx:
@v1v I've updated the PR to look for an ELASTIC_CI_POST_VERSION environment variable and append it as `.post$POST_VERSION`

v1v:
done

`Checking dist/elastic_apm-6.21.3.post8236877975-py2.py3-none-any.whl: PASSED with warnings`

in https://github.com/elastic/apm-agent-python/actions/runs/8236877975/job/22524360434

I cherry-picked your changes in https://github.com/elastic/apm-agent-python/pull/1998

trentm:
Would `.devXXX` be more appropriate than `.postXXX` versions?

xrmx:
> Would `.devXXX` be more appropriate than `.postXXX` versions?

I see x.y.z.dev as early releases of a x.y.z release and not something created after it. In fact we are bumping the version of the package just before tagging the release so to me does not make much sense to release x.y.z.devXXX after x.y.z.

trentm:
> I see x.y.z.dev as early releases of a x.y.z release and not something created after it.

Yah, for ordering you are right. IIUC, you are not ever publishing these to `pypi.org`, only possibly to `test.pypi.org`, so all good.



For the "spirit" of those releases, I think ".dev" fits better. From https://packaging.python.org/en/latest/specifications/version-specifiers/#post-releases

> Post-releases
> Some projects use post-releases to address minor errors in a final release that do not affect the distributed software (for example, correcting an error in the release notes).

vs.

> Developmental releases
> Some projects make regular developmental releases, and system packagers (especially for Linux distributions) may wish to create early releases directly from source control which do not conflict with later project releases.

trentm:
> What about something like`post0.devXXX`? So it's clear that these are development releases?

Nah. I think your using `.postNNN` is just fine. You are only intending to publish to **test.pypi.org**, correct?

xrmx:
> > What about something like`post0.devXXX`? So it's clear that these are development releases?
> 
> Nah. I think your using `.postNNN` is just fine. You are only intending to publish to **test.pypi.org**, correct?

Yes, this stuff will be only published to test.pypi.org.

trentm (approved):

PR #2002, opened 2024-03-12
instrumentation/urllib3: fix urllib3 2.0.1+ crash with many args

### LINKED ISSUE #1928: AttributeError: 'Retry' object has no attribute 'copy'
The `elasticapm` Python client fails with `AttributeError: 'Retry' object has no attribute 'copy'` during HTTP retries.

The Elastic APM UI points towards the following lines being responsible for the crash https://github.com/elastic/apm-agent-python/blob/c9a1f7a74b4b8e39af06f9b383037a3ae49ca42b/elasticapm/instrumentation/packages/urllib3.py#L66-L68 

**To Reproduce**

Configure a `requests` session with [urllib3 retries](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html) as described below. Then point it at a flaky endpoint.

``` python
class TimeoutHTTPAdapter(HTTPAdapter):
    def __init__(self, retries: int, *args: Any, **kwargs: Any) -> None:
        self.timeout = kwargs.pop("timeout", 30)
        self._retries = retries
        super().__init__(*args, **kwargs)

    def send( 
        self, request: requests.PreparedRequest, **kwargs: Any
    ) -> requests.Response:
        kwargs.setdefault("timeout", self.timeout)
        return super().send(request, **kwargs)

def session_with_retries(
    retries: int = 3,
    backoff_factor: float = 1,
    timeout: int = 30,
    status_forcelist: Collection[int] = (408, 429, 500, 502, 503, 504),
) -> requests.Session:

    retry_strategy = Retry(
        total=retries,
        backoff_factor=backoff_factor,
        status_forcelist=status_forcelist,
        allowed_methods=Retry.DEFAULT_ALLOWED_METHODS.union(["POST", "PATCH"]),
    )
    adapter = TimeoutHTTPAdapter(
        retries=retries, max_retries=retry_strategy, timeout=timeout
    )

    s = requests.Session()
    s.mount("http://", adapter)
    s.mount("https://", adapter)
    return s
```

**Environment**
- OS: Linux
- Python version: Python 3.11
- Framework and version: `urllib3==2.0.6` and `requests==2.31.0`
- APM Server version: 8.10.4
- Agent version: `elastic-apm==6.19.0`

**Background**
* This appears to be related to the recent urllib3 fixes https://github.com/elastic/apm-agent-python/pull/1822 and https://github.com/elastic/apm-agent-python/issues/1816

xrmx:

In c1dd69e0a5b9d1ec020666a97e0b7cb788b86c4f we changed the logic of update_headers taking into account a new body arg before the one for headers.
The problem is that HTTPConnectionPool.urlopen did not change at all, only HTTPConnectionPool.request did so the old login update_headers was fine.

## Related issues

Closes #1928


trentm (approved):

xrmx:
@tonyman19 need a couple more commits to get in and then I'll cut a new release

PR #2057, opened 2024-06-04
github-action: run buildkite action with GH secrets

v1v:

Run the Buildkite pipeline for the microbenchmarks without waiting for. 

There is no need to wait for, since it uses a different CI - logs can be accessed through the UI, the GitHub action will provide the link to it.

## Related issues

Closes #ISSUE

### DIFF
--- a/.github/workflows/microbenchmark.yml
+++ b/.github/workflows/microbenchmark.yml
@@ -16,32 +16,16 @@ permissions:
 jobs:
   microbenchmark:
     runs-on: ubuntu-latest
-    # wait up to 1 hour
-    timeout-minutes: 60
+    timeout-minutes: 5
     steps:
-      - id: buildkite
-        name: Run buildkite pipeline
-        uses: elastic/apm-pipeline-library/.github/actions/buildkite@current
+      - name: Run microbenchmark
+        uses: elastic/oblt-actions/buildkite/run@v1.5.0
         with:
-          vaultUrl: ${{ secrets.VAULT_ADDR }}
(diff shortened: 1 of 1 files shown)


reakaleek (approved):

.github/workflows/microbenchmark.yml
```diff
@@ -16,32 +16,16 @@ permissions:
...
+          pipeline: "apm-agent-microbenchmark"
+          token: ${{ secrets.BUILDKITE_TOKEN }}
+          wait-for: false
+          env-vars: |
             script=.ci/bench.sh
             repo=apm-agent-python
             sha=${{ github.sha }}
             BRANCH_NAME=${{ github.ref_name }}
-
-      - if: ${{ failure() }}
```
xrmx, line 38: Why is slack failure notification removed? I don't think it would a problem but am curious

v1v: The Buildkite pipeline will be triggered using GitHub actions, and the GitHub action won't wait for the build to finish, see `wait-for: false`. The Slack notifications are gonna be handled in the Buldkite pipeline itself.

There is one main reason:
- Avoid paying twice, GitHub runners waiting for the build to finish, Buildkite runner actually running the microbenchmark.

While the notifications can be anyway done in Buildkite itself, at the moment the microbenchmarks are broken because a recent change in the git https cloning that replaced the git ssh/git cloning.

xrmx: Thanks for the explanation

xrmx (approved):

PR #2085, opened 2024-07-18
elasticapm/transport: specific shutdown handling for http transport

### LINKED ISSUE #1975: "Closing the transport connection timed out." caused by race condition
**Describe the bug**: Occasionally, when using the `elasticapm.Client` (without a framework), during process shutdown (in the `atexit` handler), the transport thread will block forever while trying to send data to the APM server and subsequently be killed by the thread manager, after the configured timeout is reached. This causes "Closing the transport connection timed out." to be printed to the command line and the messages remaining in the buffer to be lost.

This seems to be caused by a race condition involving the `atexit` handler of the `elasticapm.Client` and the `weakref.finalize` of `urllib3.connectionpool.HTTPConnectionPool` (which uses an `atexit` handler under the hood) which calls `_close_pool_connections`. A timeline causing this bugs looks like this:
1. The process is about to shutdown. `atexit` handlers are called.
2. `_close_pool_connections` is called while all connections are in the pool. All existing connections are disposed.
3. The `elasticapm.Client` `atexit` handler is called, sending the "close" event to the transport thread.
4. The transport thread handles the "close" event, flushing the buffer and trying to send remaining data to the APM server.
5. `urlopen` will block the transport thread forever while waiting to get a connection from the connection pool (since the poolmanager uses `block=True` and no pool timeout is configured, this will block forever, because the only way to get a connection from the pool this way, is for someone else to put a connection into the pool).
6. The thread manager kills the thread after the configured timeout is reached, printing the error message and loosing all data in the buffer

The reason why this does occur consistently, is because `_close_pool_connections` will not clean up connections which are currently in use (e.g. connections being used in another thread). If a request is in progress when `_close_pool_connections` is called, the associated connection "survives" the cleanup and will be added back to the pool afterwards and can be reused by the transport thread (which may be a bug/unintended behavior of urllib3 since it claims `HTTPConnectionPool` is thread safe).

**To Reproduce**

The following minimal example reproduces the issue:

```py
import time

import elasticapm

# NOTE: You should be able to remove the "config" argument in your environment
client = elasticapm.Client(
    service_name="<SERVICE>",
    server_url="<APM_SERVER_URL>",
    secret_token="<SECRET_TOKEN>",
    config={
        "SERVER_CA_CERT_FILE": "<INTERNAL_CA_FILE_PATH>",
        "GLOBAL_LABELS": {"Tenant": "<TENANT>"},
    },
)

client.capture_message("Test")

# Give the client time to resolve all internal network requests, ensuring
# that all urllib connections are in the pool when the atexit handlers are called
time.sleep(10)
```

As is the case with race conditions, you might have to fiddle with the sleep timing a little bit. 10 seconds work quite reliable in my environment, but you may need a few more/less seconds, depending on your environment.

**Environment**
- OS: Windows 10
- Python version: 3.11.7
- package versions: `urllib3==2.2.1`
- APM Server version: 8.11.3
- Agent version: `elastic-apm==6.20.0`

**Additional context**

A workaround for my use case is to use a custom Transport class which uses a non-blocking pool. I don't know enough about the `elastic-apm` code base to know whether or not this causes issues in other parts of the package, but it seems to resolve the issue for me without causing any other major issues.

```py
def get_import_string(cls) -> str:
    module = cls.__module__
    if module == "builtins":
        # avoid outputs like 'builtins.str'
        return cls.__qualname__
    return module + "." + cls.__qualname__

class NonBlockingTransport(Transport):
    def __init__(self, *args, **kwargs) -> None:
        super(NonBlockingTransport, self).__init__(*args, **kwargs)
        self._pool_kwargs["block"] = False

# Use like this:
client = elasticapm.Client(
    ...,
    config={
        ...,
        "TRANSPORT_CLASS": get_import_string(NonBlockingTransport),
    },
)
```

xrmx:

We have a race condition at http transport shutdown where our atexit handler is racing against urllib3 ConnectionPool weakref finalizer. Having the urllib3 finalizer called before atexit would lead to have our thread hang waiting for send any eventual queued data via urllib3 pools that are closed.
Force the creation of a new PoolManager so that we are always able to flush.

## Related issues

Closes #1975

### DIFF
--- a/tests/transports/test_urllib3.py
+++ b/tests/transports/test_urllib3.py
@@ -30,6 +30,7 @@
 
 
 import os
+import time
 
 import certifi
 import mock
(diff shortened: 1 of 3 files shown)


basepi (dismissed):

tests/transports/test_urllib3.py
```diff
@@ -517,3 +518,89 @@ def test_fetch_server_info_flat_string(waiting_httpserver, caplog, elasticapm_cl
...
+    waiting_httpserver.serve_content(code=202, content="", headers={"Location": "http://example.com/foo"})
+    transport = Transport(
+        waiting_httpserver.url, client=elasticapm_client, headers=elasticapm_client._transport._headers
+    )
+    transport.start_thread()
+
+    with mock.patch("os.getpid") as getpid_mock:
+        getpid_mock.return_value = 0
+        transport.close()
+        time.sleep(0.1)
```
xrmx, line 548:
```suggestion
```

basepi (dismissed):

xrmx:
It looks like the tests that replicates the submission timeout is flaky on windows:

```
=========================== short test summary info ===========================
FAILED tests/transports/test_urllib3.py::test_close_timeout_error_without_flushing - assert True is False
 +  where True = <bound method Event.is_set of <threading.Event object at 0x00000224A10BE0D0>>()
 +    where <bound method Event.is_set of <threading.Event object at 0x00000224A10BE0D0>> = <threading.Event object at 0x00000224A10BE0D0>.is_set
 +      where <threading.Event object at 0x00000224A10BE0D0> = <elasticapm.transport.http.Transport object at 0x00000224A10CDDF0>._flushed
===== 1 failed, 745 passed, 63 skipped, 138 warnings in 90.95s (0:01:30) ======
```

basepi:
> It looks like the tests that replicates the submission timeout is flaky on windows

Honestly, I've been extremely tempted to remove our windows support entirely. I don't think anyone uses it, and the majority of our flakey tests have been on windows. (You can easily search for some of our other tests we skip on windows due to flakiness.)

tests/transports/test_urllib3.py
```diff
@@ -517,3 +518,88 @@ def test_fetch_server_info_flat_string(waiting_httpserver, caplog, elasticapm_cl
...
+    transport.close()
+
+
+def test_close_timeout_error_without_flushing(waiting_httpserver, caplog, elasticapm_client):
+    elasticapm_client.server_version = (8, 0, 0)  # avoid making server_info request
+    waiting_httpserver.serve_content(code=202, content="", headers={"Location": "http://example.com/foo"})
+
+    with caplog.at_level("INFO", logger="elasticapm.transport.http"):
+        with mock.patch.object(Transport, "_max_flush_time_seconds", 0):
+            with mock.patch.object(Transport, "_flush") as flush_mock:
```
xrmx, line 576: We are not calling this mock, it fails randomly but not very often on linux too

xrmx:
> > It looks like the tests that replicates the submission timeout is flaky on windows
> 
> Honestly, I've been extremely tempted to remove our windows support entirely. I don't think anyone uses it, and the majority of our flakey tests have been on windows. (You can easily search for some of our other tests we skip on windows due to flakiness.)

This time the problem was the test itself :)

basepi (approved):

PR #2116, opened 2024-09-04
ci: fix upload of codecov data

xrmx:

From latest release upload artifact will ignore hidden files by default, configure to don't ignore them when uploading coverage data.

Should fix:

```
Run actions/download-artifact@v3
Starting download for coverage-reports
Error: Unable to find an artifact with the name: coverage-reports
```

### DIFF
--- a/.github/workflows/run-matrix.yml
+++ b/.github/workflows/run-matrix.yml
@@ -38,3 +38,4 @@ jobs:
         with:
           name: coverage-reports
           path: "**/.coverage*"
+          include-hidden-files: true


.github/workflows/run-matrix.yml
```diff
@@ -38,3 +38,4 @@ jobs:
         with:
           name: coverage-reports
           path: "**/.coverage*"
+          include-hidden-files: true
```
v1v, line 41: It's such a coincidence that I've been reading about this default behaviour recently.

v1v (approved):

.github/workflows/run-matrix.yml
v1v: If possible, can we add `- uses: actions/checkout@v4` with `persist-credentials: false`:

> The auth token is persisted in the local git config. This enables your scripts to run authenticated git commands. The token is removed during post-job cleanup. Set persist-credentials: false to opt-out.

https://github.com/actions/checkout/tree/v4/

PR #2160, opened 2024-11-25
Use build to create distributions

### LINKED ISSUE #2159: Update built distributions name to adhere to PEP 625
```
This email is notifying you of an upcoming deprecation that we have determined may affect you as a result of your recent upload to 'elastic-apm'.
In the future, PyPI will require all newly uploaded source distribution filenames to comply with [PEP 625](https://peps.python.org/pep-0625/). Any source distributions already uploaded will remain in place as-is and do not need to be updated.
Specifically, your recent upload of 'elastic-apm-6.23.0.post12008437098.tar.gz' is incompatible with PEP 625 because it does not contain the normalized project name 'elastic_apm'.
In most cases, this can be resolved by upgrading the version of your build tooling to a later version that supports PEP 625 and produces compliant filenames.
```

xrmx:

Instead of calling setup.py. This fixes the name of built distributions to match pep625, i.e. elastic_apm instead of elastic-apm.

## Related issues

Closes #2159

### DIFF
--- a/dev-utils/make-packages.sh
+++ b/dev-utils/make-packages.sh
@@ -3,14 +3,10 @@
 # Make a Python APM agent distribution
 #
 
-echo "::group::Install wheel"
-pip install --user wheel
+echo "::group::Install build"
+pip install --user build
 echo "::endgroup::"
 
-echo "::group::Building universal wheel"
-python setup.py bdist_wheel
-echo "::endgroup::"
-
-echo "::group::Building source distribution"
-python setup.py sdist
+echo "::group::Building packages"
+python -m build
 echo "::endgroup::"


v1v (approved):

v1v:
How will this change look like in https://test.pypi.org/project/elastic-apm/ and  https://pypi.org/project/elastic-apm/?

xrmx:
> How will this change look like in https://test.pypi.org/project/elastic-apm/ and https://pypi.org/project/elastic-apm/?

It will build wheels and tarballs named `elastic_apm` instead of `elastic-apm`:

```
-rw-rw-r-- 1 rm rm 169731 mag 31 10:09 elastic-apm-6.22.2.tar.gz
-rw-rw-r-- 1 rm rm 351466 nov 25 12:06 elastic_apm-6.23.0-py2.py3-none-any.whl
-rw-rw-r-- 1 rm rm 169730 nov 25 12:06 elastic_apm-6.23.0.tar.gz
````

basepi (approved):

PR #2194, opened 2025-01-13
Fix sanic cookies headers serialization

xrmx:

This is a follow up of #2190 that fixes the full matrix.

We are getting a Cookie instance from Sanic that by default get translated to a dict that is not a proper type for an header. Instead convert it to its string representation.

Also use a syntax for setting cookies that work on older versions.


xrmx (commented):
/test matrix

estolfo (approved):

xrmx (commented):
/test matrix

PR #2216, opened 2025-02-07
Python3.13 support

### LINKED ISSUE #2213: Python 3.13 support
Python 3.13 raises an exception with how we buffer things to send to the apm server. When the gzip object is closed it raises an exception because the BytesIO used as its backing storage is already gone. 

Filed an upstream issue https://github.com/python/cpython/issues/129726 so waiting for some feedback here before reworking our code.

xrmx:

This is draft because the change in the buffering must be measured

## Related issues

Closes #2213

### DIFF
--- a/elasticapm/transport/base.py
+++ b/elasticapm/transport/base.py
@@ -153,7 +153,7 @@ class Transport(ThreadManager):
                     buffer_written = True
                     self._counts[event_type] += 1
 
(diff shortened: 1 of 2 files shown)


h0rn3t:
same thing fix plz

injectionsuccesfully:
+

lsaint (approved):

zffocussss:
@xrmx Is it fixed?

PR #2225, opened 2025-02-28
add target in case there is no host in target

ajaygupta2790:
The current grpc instrumentation expects host value with a port that may not always be true. In our project, we usean internal host without a port value
```
target = kwargs.get("target") or args[0]
if ":" in target:
    host, port = target.split(":")
    try:
        port = int(port)
    except ValueError:
        port = None
else:
    host, port = None, None
```
This results in span showing None value and we cannot track the external host value. All grpc hosts are clubbed together under https://none/ value

The proposal is to keep the target value as is in case there is no ':' in the host and keep the port as None


xrmx:
Thanks, care to add a test too please?

ajaygupta2790:
test case file is missing for grpc, need some suggestions on where to start @xrmx

tests/instrumentation/grpc_tests.py
```diff
@@ -0,0 +1,158 @@
...
+from elasticapm.conf import constants
+from elasticapm.conf.constants import TRANSACTION
+from elasticapm.traces import capture_span
+from elasticapm import Client
+from elasticapm.contrib.grpc.client_interceptor import _ClientInterceptor
+from elasticapm.contrib.grpc.server_interceptor import _ServerInterceptor
+from elasticapm.contrib.grpc.async_server_interceptor import _AsyncServerInterceptor
+from tests.fixtures import TempStoreClient, instrument
+from tests.instrumentation.test_pb2 import UnaryUnaryRequest, UnaryUnaryResponse
+from tests.instrumentation.test_pb2_grpc import TestServiceServicer, TestServiceStub, add_TestServiceServicer_to_server
```
xrmx, line 48: These modules should be renamed otherwise test runners think these contains tests and runs them

PR #2233, opened 2025-03-13
feat: enable Dockerfile check for dependabot

kruskall:

keep docker image up to date

## Related issues

Closes #ISSUE

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -30,6 +30,12 @@ updates:
         patterns:
           - "*"
 
+  # docker
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+
   # GitHub composite actions
   - package-ecosystem: "github-actions"
     directory: "/.github/actions/packages"


trentm:
ping @xrmx

xrmx:
Don't we have already renovate for that?

xrmx:
And I don't merge renovate PRs often because they bump the major python version image that I don't want to.

xrmx:
> Sorry for reopening and for the late reply 😓
> 
> > Don't we have already renovate for that?
> 
> I think renovate config is only enabled for chainguard images:
> 
> https://github.com/elastic/apm-agent-python/blob/a15ec54f72ebae5c5f31c3feb8979130cb6901ae/renovate.json#L4
> 
> I'll leave it up to you and the python agent team to decide whether you want to extend the config to other docker images or want to use renovate but the image should be kept up to date.

I would like to use just one tool for the same purpose

> > I don't merge renovate PRs often because they bump the major python version image that I don't want to.
> 
> I don't think this is ok :( I see #2206 is open for example. Maybe the config can be updated to not bump major versions ? cc @jamietanna

That would be great of course

xrmx:
Handing the update to dependabot has been merged in https://github.com/elastic/apm-agent-python/pull/2246 . Now I just need to configure it :sweat:

PR #2246, opened 2025-03-26
dependabot: use directories and use docker ecosystem

v1v:

Use `dependabot` for updating docker images stored in our internal docker registry using [this](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/configuring-access-to-private-registries-for-dependabot).

Use `directories` to manage the dependencies for the GitHub workflow sand composite actions.

### Why

We already have secrets stored for accessing the internal docker registry. Use one dependency management tool with native support for GitHub actions and the [GH secrets access control](https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions/). 

This has been already tested in the past in a sandbox repository and also in another GH internal repositories.

### Actions

- [x] Create Dependabot GitHub secret using CasC.

## Related issues

Done also in https://github.com/elastic/apm-agent-nodejs/pull/4539


xrmx:
Would it be possible to configure dependabot to just bump patch releases for a specified image?

PR #2254, opened 2025-03-31
Revert "ci: pin actions to specific commits"

v1v:
Reverts elastic/apm-agent-python#2236

I prefer that we don't pin versions coming from the `actions` for now.

Let's agree on this. At the moment, it requires some manual actions in our end to validate change for something we follow the GH security hardening:
- https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions

### DIFF
--- a/.github/actions/build-distribution/action.yml
+++ b/.github/actions/build-distribution/action.yml
@@ -6,7 +6,7 @@ description: Run the build distribution
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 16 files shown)


xrmx (approved):

v1v:
> not all of these are first party action

I see two actions?
- `actions-ecosystem`
- `precommit`

I'll pin them in a follow-up. I think the other ones are either coming from `actions` or the `elastic` org.

PR #2261, opened 2025-04-04
tests: enable transports tests poking with request_size for Python 3.12+

xrmx:

I don't know if python 3.12 changed gzip compression or what but we get a flush call for each message we send so since our message is compressed to 10 bytes lower the limit to 9 bytes to pass the check deciding to flush the buffer.

## Related issues

### DIFF
--- a/tests/transports/test_base.py
+++ b/tests/transports/test_base.py
@@ -157,24 +157,26 @@ def test_api_request_time_dynamic(mock_send, caplog, elasticapm_client):
     assert mock_send.call_count == 0
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 12), reason="Failing locally on 3.12.0rc1")  # TODO py3.12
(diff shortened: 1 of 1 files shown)


tests/transports/test_base.py
```diff
@@ -157,43 +157,46 @@ def test_api_request_time_dynamic(mock_send, caplog, elasticapm_client):
...
 
 
-@pytest.mark.skipif(sys.version_info >= (3, 12), reason="Failing locally on 3.12.0rc1")  # TODO py3.12
+def _cleanup_flush_mock_buffers(mock_flush):
+    args, kwargs = mock_flush.call_args
+    buffer = args[0]
+    fileobj = buffer.fileobj
+    buffer.close()
+    compressed_data = fileobj.getbuffer()
+    compressed_data.release()
```
xrmx, line 166: Maybe these should just be fileobj.close() ?

basepi (approved):

PR #2265, opened 2025-04-09
Add official support for Python 3.13

### LINKED ISSUE #2213: Python 3.13 support

xrmx:

And start running CI against it.

## Related issues

Closes #2213


xrmx:
This needs #2266

basepi (approved):

basepi (approved):

PR #2311, opened 2025-06-03
docs: update options to enable with uwsgi

### LINKED ISSUE #2304: Slow restarts with chain-reload
With `elastic-apm` installed the following configuration `uwsgi` takes a lot to reload workers:

```
[uwsgi]
module = djangouwsgi.wsgi
http = :8000
enable-threads = true
chain-reload = true
touch-chain-reload = /tmp/uwsgi.reload
pidfile = /tmp/uwsgi.pid
processes = 2
threads = 2
master = true
vacuum = true
```

As they are killed after a 60s timeout:

```
Fri May 30 08:18:56 2025 - *** /tmp/uwsgi.reload has been touched... chain reload !!! ***
Fri May 30 08:18:56 2025 - chain next victim is worker 1
Gracefully killing worker 1 (pid: 1235583)...
Fri May 30 08:19:57 2025 - worker 1 (pid: 1235583) is taking too much time to die...NO MERCY !!!
worker 1 killed successfully (pid: 1235583)
```

Adding `lazy-apps` fixes the issue so we should investigate this and update the documentation if we are gonna to require that option.

xrmx:

We need to have py-call-uwsgi-fork-hooks set in order to have proper handling of our threads.
While at it say that threads are enabled by default since a few releases.

## Related issues

Closes #2304

### DIFF
--- a/docs/reference/django-support.md
+++ b/docs/reference/django-support.md
@@ -24,7 +24,7 @@ For apm-server 6.2+, make sure you use version 2.0 or higher of `elastic-apm`.
 
 
 ::::{note}
-If you use Django with uwsgi, make sure to [enable threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads).
+If you use Django with uwsgi, make sure to [enable threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads) (enabled by default since 2.0.27) and [py-call-uwsgi-fork-hooks](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#py-call-uwsgi-fork-hooks).
 ::::
 
 
--- a/docs/reference/flask-support.md
+++ b/docs/reference/flask-support.md
@@ -24,7 +24,7 @@ For apm-server 6.2+, make sure you use version 2.0 or higher of `elastic-apm`.
 
 
 ::::{note}
-If you use Flask with uwsgi, make sure to [enable threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads).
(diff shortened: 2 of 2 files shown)


docs/reference/django-support.md
```diff
@@ -24,7 +24,7 @@ For apm-server 6.2+, make sure you use version 2.0 or higher of `elastic-apm`.
 
 
 ::::{note}
-If you use Django with uwsgi, make sure to [enable threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads).
+If you use Django with uwsgi, make sure to [enable threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads) (enabled by default since 2.0.27) and [py-call-uwsgi-fork-hooks](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#py-call-uwsgi-fork-hooks).
```
colleenmcginnis, line 27: Does it make sense to say they need to enable threads and enable py-call-uwsgi-fork-hooks or is there some other verb associated with py-call-uwsgi-fork-hooks? If the former, maybe:

```suggestion
If you use Django with uwsgi, make sure to enable [threads](http://uwsgi-docs.readthedocs.org/en/latest/Options.html#enable-threads) (enabled by default since 2.0.27) and [py-call-uwsgi-fork-hooks](https://uwsgi-docs.readthedocs.io/en/latest/Options.html#py-call-uwsgi-fork-hooks).
```

xrmx: `enable-threads` is the option you need to enable so I think that's the reason the verb is on the link but both versions work for me

colleenmcginnis: I'll leave it up to you!

colleenmcginnis (approved):
✨

PR #2464, opened 2025-10-30
Cleanup elasticsearch testing

xrmx:

Drop testing against ancient versions, add testing for 9.x, bump docker images to latest and while at it drops the version from the docker-compose.yml.


.ci/.matrix_framework.yml
```diff
@@ -29,9 +29,9 @@ FRAMEWORK:
   - pyodbc-newest
   - memcached-newest
   - pylibmc-newest
-  - elasticsearch-2
```
trentm: !

PR #2504, opened 2025-12-04
updated the __all__ section to include all  22 publicly exposed varia…

md-adil:

This makes it clear which symbols are part of the public API and improves IDE autocomplete and static analysis for Python integrations using this package.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,1 +1,1 @@
-updatecli v0.111.0
+updatecli v0.112.0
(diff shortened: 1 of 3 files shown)


xrmx:
@md-adil If you look at the commit history in https://github.com/elastic/apm-agent-python/pull/2504/commits you'll see that your commit is not linked to your account. I guess that's because you commit with an email that is different than the one you use on github. I think that's the reason CLA check is red.

md-adil:
> __all__

Some tools rely on `__all__` to determine which names are publicly exported. In **VS Code, this is handled by Pylance (Pyright)** for auto-imports, IntelliSense, and static analysis. If a symbol is exposed at runtime but missing from `__all__`, those tools treat it as non-public.

`async_capture_span` is just one example of a symbol that was available but not listed in `__all__`. This PR ensures that `__all__` accurately reflects all intended public exports, so editor tooling and static analysis behave consistently.

basepi (approved):

PR #2516, opened 2025-12-19
elasticapm: introduce ELASTIC_APM_SKIP_SERVER_INFO

xrmx:

Introduce this option in order to avoid fetch the apm server info to get its version, off by default.
Calling that endpoint with the aws lambda extension incurs in a measurable latency on AWS Lambda.
Enabling this issue requires the receiving APM server version to be greater than 8.7.0.

Option is still in technical preview.


basepi (approved):
Looks great! Definitely don't want that kind of startup latency on lambda. 👍

PR #2584, opened 2026-03-19
fix grpc host issue and add grpc test cases

ajaygupta2790:
The current grpc instrumentation expects a host value with a port that may not always be true. In our project, we use an internal host without a port value
```
target = kwargs.get("target") or args[0]
if ":" in target:
    host, port = target.split(":")
    try:
        port = int(port)
    except ValueError:
        port = None
else:
    host, port = None, None
```

This results in the span showing a None value, and we cannot track the external host value. All grpc hosts are clubbed together under https://none/ value

The proposal is to keep the target value as is in case there is no ':' in the host, and keep the port as None


ajaygupta2790:
@xrmx changes done

xrmx (approved):

PR #2675, opened 2026-06-15
tests/requirements: limits kafka-newest to < 3

xrmx:

Looks like kafka-python 3.0 requires a newer server so limit to <3


hegerchr (approved):
LGTM

